PR #1, opened 2019-07-16
Configure web application with npm & webpack

mihailradkov:
Changes summary:

- Configured GIT ignore
  - Added rules for IntelliJ specific IDE files and for Node JS development
- Removed Maven related XMLs
  - The workbench will be transitioned to a web application and Maven will not be needed anymore
- Moved workbench modules from root directory to src/ folder
- Flattened the workbench directory structure
  - Moved the framework folders content to src/css src/js src/img etc
  - Moved tests to test/ folder
- Moved resources/lib to src/js/lib and fixed affected paths
- Added NPM package.json and Webpack configs
  - Added webpack.config.js which is for preparation for transitioning the workbench code for bundling
  - Added webpack.config.dev.js which starts a development server with proxying
  - Moved app.js from src/js/angular/ to src/ to act as the main app entrypoint
- Added example Dockerfile and Docker compose
  - The Dockerfile acts as an example how to serve the workbench application with Nginx along with proxying to the GraphDB REST APIs
- Updated the readme file


framework/similarity/pages/viewQuery.html
```diff
@@ -1,13 +0,0 @@
-<div class="modal-header">
```
desislava-hristova-ontotext: Why is that file removed?

mihailradkov: It was duplicated

PR #9, opened 2019-07-23
Configure eslint and editorconfig

svilenvelikov:
* Introduced eslint and applied google's code style rules.
* Introduced editorconfig.
* Increased nesting space
* Decreased eslint version due to issues with intellij.
* Disabled some rules which are not mandatory for now.
* Introduced .eslintignore and excluded every js file but the main app.js and workbench code.


.eslintrc
```diff
@@ -0,0 +1,23 @@
...
+    "sourceType": "module",
+    "ecmaFeatures": {
+      "impliedStrict": true
+    }
+  },
+  "env": {
+    "es6": true,
+    "node": true,
+    "browser": true,
+    "jest": true
```
mihailradkov: Copy paste?

svilenvelikov: obviously

PR #15, opened 2019-07-25
Reorganize code

svilenvelikov:
* Move js library files in js/lib.
* Move application code in js/angular.
* Group some library files in packages along with their css files.
* Deleted many unused files.
* Grouped fonts in their own root fonts folders.


desislava-hristova-ontotext:
LGTM

PR #18, opened 2019-07-26
Define which directory should be published in npm registry

svilenvelikov:
Currently we deploy the whole application code along with the libraries. That's why we publish only the src folder. Next step is to clean up the code and create bundles from it using webpack. Then we would need to publish just the dist folder with the created bundles.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,54 +1,57 @@
 {
-  "name": "graphdb-workbench",
-  "version": "0.1.0",
-  "description": "The web application for GraphDB APIs",
-  "scripts": {
-    "build": "npm run less:preprocess",
-    "lint": "eslint ./src",
-    "start:dev": "webpack-dev-server --config=webpack.config.dev.js",
-    "test": "karma start",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,54 +1,57 @@
...
+        "build": "npm run less:preprocess",
+        "lint": "eslint ./src",
+        "start:dev": "webpack-dev-server --config=webpack.config.dev.js",
+        "test": "karma start",
+        "test:ci": "rimraf coverage/ && karma start --single-run --browsers PhantomJS",
+        "test:coveralls": "cat coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
+        "less:watch": "less-watch-compiler --config=./less-watch-compiler.config.json",
+        "less:preprocess": "less-watch-compiler --config=./less-compiler.config.json"
+    },
+    "files": [
```
svilenvelikov: I'm sorry, I've formatted the file. The only difference is this files section.

PR #24, opened 2019-07-30
Configured Sonar coverage and project version

mihailradkov:


package.json
```diff
@@ -10,7 +10,8 @@
         "test:ci": "rimraf coverage/ && karma start --single-run --browsers PhantomJS",
         "test:coveralls": "cat coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
         "less:watch": "less-watch-compiler --config=./less-watch-compiler.config.json",
-        "less:preprocess": "less-watch-compiler --config=./less-compiler.config.json"
+        "less:preprocess": "less-watch-compiler --config=./less-compiler.config.json",
+        "sonar": "sonar-scanner -Dsonar.projectVersion=$npm_package_version"
```
mihailradkov: I'm not very comfortable with this being invoked as a script due to binding to a tool which is outside of this application. But currently this is the easiest approach to provide the version. The other is to use a tool to extract it from the JSON.

PR #29, opened 2019-08-02
Bundle workbench with webpack

svilenvelikov:
Workbench application consists from many resources: script files, stylesheets, images, fonts. On an initial load more than 140 http requests for resources are being executed. It's a standard these days web applications to be built and optimized in a such way that as few http requests as possible to be made. This usually is made by utilizing bundlers like webpack, browserify and others. 
This MR introduces a webpack configuration and some code refactoring needed in order the workbench code to be properly built and optimized for production. 
After the bundling requests made during initial load are reduced to 34. 

**Related MR for GDB which has to be merged as well: https://gitlab.ontotext.com/graphdb-team/graphdb/merge_requests/1220**

Changes introduced are as follows:
* Introduced webpack dev and production configurations as well as a common config which is merged with the prod and dev once. This is done for avoiding duplications and sophisticated code in the webpack configuration regarding production and developer targetted builds.
* Removed the separated less stylesheets processing and replaced with less-loader rule in webpack.
* Introduced a template for the `index.html` file which is built by webpack.
* Introduced `vendor.js` file where the third party libraries are imported instead of loading them by a `<script>` tag from whithin the html pages/templates. The vendor file is processed by the webpack and all imported scripts and stylesheets are injected there.
* Introduced `main.js` where some workbench application files are imported and processed by webpack. We can't import those in the `app.js` because this would break the bundle and the tests.
* The libraries which are now bundled and globally exposed are no more requiested in the application code. Those are mainly the jquery and angularjs. As some libraries depend from the jquery library, it is now exposed globally through the webpack's `expose-loader` loader.
* `/pages` folder is now directly copied in `/dist`. This forlder contains mainly view templates.
* `/res` folder is directly copied in `/dist`. It contains the swagger resources and some other templates.
* Template directories for all components which have angular directives are also directly copied in `/dist`.
* Many resources: css, js, and images are directly referenced from templates. These are now directly copied in `/dist` to allow proper loading.
* A small change in the rzslider component was made due to the fact that for some reason it was not properly initialized, which led to an error because angular was not injected properly in it. Most likely the reason for this is incorrect initialization of a module which uses it. Now rzslider is preloaded in the `main.js` bundle and it seems to work correctly using the globally available angular.
* Deleted some commented out code.
* Some images were referenced from `src/less/core.less` and I didn't found them so I deleted those references.
* Deleted the minified versions of the less stylesheets as those are processed during the bundling and are no more requested directly but from the bundle itself.
* `bootstrap-switch.min.js` is now bundled in the `main.js` as it was loaded from within a template with a script tag.
* In production build bundles and the index.html are minified.
* Updated the `package.json#files` to point the `/dist` folder which means that now the when the workbench is published the `/dist` will be published in NPM instead of `/src`.

After the workbench bundling it wouldn't depend on requirejs in production code anymore.

**Possible future improvements**
* More granular splitting of the source code during the bundling. Currently a single bundle with application code is made plus one small bundle for the `clustermanager` module which is made automatically by webpack because of the dynamic import in the main app.js. In future, application views could be analized and eventually split as separate bundles. This would decrease the main bundles size and improve peformance even more.
* Convert all application stylesheets to Less and Import those which has impact on the overall view in the main bundle. Also find a way to import the stylesheets related to specific views and angular directives in their components only.
* The d3 library was loaded in the main index.html and now it is still preloaded during the initial load. It should be checked if it's mandatory to have it loaded this way or it can be moved to a bundle/component which depends on it and not to have it preloaded by default without need.
* The yasr.js currently contains the source of couple of other libraries like jquery and lodash and maybe others. This is the reason the file to be so big even minified. Check if it's possible to have the external libraries removed from the yasr bundle and just make it dependent on them. Those libraries are used in the workbench source anyway, so this redundancy could eventually be avoided. Eventual blocker for this task would be if the yasr somehow depends on specific versions of the jquery, lodash and the other bundled libraries. This would be blocking issue for separating them unless yasr could be updated to work with more recent library versions.

Jira: https://ontotext.atlassian.net/browse/GDB-3708


svilenvelikov:
Soo, @desislava-hristova-ontotext and @mihailradkov don't be afraid to review and comment on this PR :)

package.json
```diff
@@ -43,23 +43,36 @@
...
+        "style-loader": "^0.23.1",
+        "url-loader": "^2.1.0",
         "webpack": "^4.35.3",
         "webpack-cli": "^3.3.6",
-        "webpack-dev-server": "^3.7.2"
-    }
+        "webpack-dev-server": "^3.7.2",
+        "webpack-merge": "^4.2.1"
+    },
+    "proxy": "http://localhost:7200"
```
mihailradkov: I don't think this is needed/used

src/js/lib/common/svg-export.js
```diff
@@ -1,30 +1,33 @@
-var SVG = SVG || {};
+define([], function() {
```
mihailradkov: I'm starting to think these 3 files belong in js/angular, not in js/lib

src/template.html
```diff
@@ -6,28 +6,12 @@
     <title>
         GraphDB Workbench
     </title>
-    <base href="/" target="_blank"/>
-    <meta content="width=device-width" name="viewport">
-    <link rel="icon" href="img/icon.png">
-    <link rel="apple-touch-icon-precomposed" href="img/icon.png">
-
-    <link href="js/lib/bootstrap/bootstrap.min.css" rel="stylesheet">
```
mihailradkov: It satisfactory to see these gone

test-cypress/integration/explore/visual.graph.spec.js
```diff
@@ -17,8 +19,12 @@ describe('Visual graph screen validation', () => {
     });
 
     it('Test notification when autocomplete is disabled', () => {
```
mihailradkov: _Could_ be moved in context('When autocomplete is disabled') and the rest in similar context. The idea is to `enableAutocomplete` only where needed without having to visit multiple pages

package.json
svilenvelikov: xmm did we remove it?

src/js/lib/common/svg-export.js
svilenvelikov: yeah makes sense I can move them in a `js/angular/common/` library. @desislava-hristova-ontotext what do you think?

test-cypress/integration/explore/visual.graph.spec.js
svilenvelikov: It's only this specific test where autocomplete is disabled. And to be honest I didn't manage to find out why it started failing. It used to pass on master. For some reason autocomplete disabling through the cypress custom command yelds successful autocomplete disabling but the test fails like it is not disabled.

src/js/lib/common/svg-export.js
desislava-hristova-ontotext: It's not obvious, but these are third party libraries. See https://docs.google.com/spreadsheets/d/12cIovBIlN_xZph3P6DuFn4dvqNSR6F0ctqFy5k0n41g/edit#gid=0
I prefer keeping them in lib for now.

svilenvelikov: ah I see. We though that these are ours :)
I'll move them back in lib then

PR #33, opened 2019-08-12
Stabilizing import tests + code reorganization & simplification

mihailradkov:
Added import-steps file with static utility functions for working with the user and server import pages.
Added several classes in the import templates to simplify selectors in Cypress and ease the testing.

Additional changes:

- Changed retry count to 1 instead of 2.
- Tried to stabilize the selectRepo command
- Moved coveralls step in after_scrip section in Travis config. Sometimes the coveralls service is down which breaks any build in the meantime.

### DIFF
--- a/test-cypress/integration/import/import.server.files.spec.js
+++ b/test-cypress/integration/import/import.server.files.spec.js
@@ -1,137 +1,71 @@
-describe('Import screen validation', function () {
-    let existingRepo = 'w11' + Date.now();
-    let baseURI = 'http://purl.org/dc/elements/1.1/';
-    let context = 'http://example.org/context';
-    let messageSuccess = 'Imported successfully';
-    let fileToImport = 'italian_public_schools_links.nt.gz';
-    let blankNodeFile = 'bnodes.ttl';
+import ImportSteps from '../../steps/import-steps';
 
-    before(function () {
-        cy.visit('/repository');
-        // Create new one
-        cy.createNewRepo(existingRepo).wait(2000);
+describe('Import screen validation - server files', () => {
+
+    let repositoryId;
+
+    const BASE_URI = 'http://purl.org/dc/elements/1.1/';
+    const CONTEXT = 'http://example.org/context';
+    const SUCCESS_MESSAGE = 'Imported successfully';
+    const FILE_FOR_IMPORT = 'italian_public_schools_links.nt.gz';
+    const BLANK_NODE_FILE = 'bnodes.ttl';
+
+    beforeEach(() => {
+        repositoryId = 'server-import-' + Date.now();
+        cy.createRepository({id: repositoryId});
+        ImportSteps.visitServerImport(repositoryId);
     });
 
-    beforeEach(function () {
-        cy.visit('/import#user').wait(1000);
-        cy.selectRepo(existingRepo);
+    afterEach(() => {
+        cy.deleteRepository(repositoryId);
     });
 
-    afterEach(function () {
-        cy.checkAllServerFiles(true);
-        cy.resetStatusOfUploadedFiles().wait(1000);
-        cy.checkAllServerFiles(false);
+    it('Test import Server files successfully without changing settings', () => {
+        ImportSteps
+            .selectServerFile(FILE_FOR_IMPORT)
+            .importServerFiles()
+            .verifyImportStatus(FILE_FOR_IMPORT, SUCCESS_MESSAGE)
+            .verifyImportStatusDetails(FILE_FOR_IMPORT, '"preserveBNodeIds": false,');
     });
 
-    after(function () {
-        cy.navigateToPage('Setup', 'Repositories');
-        cy.deleteRepo(existingRepo).wait(1000);
+    it('Test import Server files successfully with changing settings', () => {
+        ImportSteps.selectServerFile(FILE_FOR_IMPORT)
+            .importServerFiles(true)
+            .fillBaseURI(BASE_URI)
+            .selectNamedGraph()
+            .fillNamedGraph(CONTEXT)
+            .expandAdvancedSettings()
+            .enablePreserveBNodes()
+            .importFromSettingsDialog()
+            .verifyImportStatus(FILE_FOR_IMPORT, SUCCESS_MESSAGE)
+            .verifyImportStatusDetails(FILE_FOR_IMPORT, [CONTEXT, BASE_URI, '"preserveBNodeIds": true,']);
     });
 
-    it('Test import Server files successfully without changing settings', function () {
-        cy.chooseServerFilesTab();
-        cy.selectDistinctFileToImport(fileToImport);
-        cy.importChangingSettings().wait(2000);
-        cy.verifyMessageOccurs(messageSuccess);
+    it('Test import all Server files successfully without changing settings', () => {
+        ImportSteps
+            .selectAllServerFiles()
+            .importServerFiles()
+            .verifyImportStatus(FILE_FOR_IMPORT, SUCCESS_MESSAGE)
+            .verifyImportStatus(BLANK_NODE_FILE, SUCCESS_MESSAGE);
     });
 
-    it('Test import Server files successfully with changing base uri', function () {
-        cy.chooseServerFilesTab();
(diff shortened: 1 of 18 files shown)


test-cypress/integration/import/import.server.files.spec.js
```diff
@@ -1,137 +1,72 @@
...
 
-    it('Test import all Server files successfully without changing settings', function () {
-        cy.chooseServerFilesTab();
-        cy.checkAllServerFiles(true);
-        cy.importChangingSettings().wait(2000);
-        cy.verifyMessageOccurs(messageSuccess);
+    it('Test import with resetting status of all imported files', () => {
+        ImportSteps
+            .selectAllServerFiles()
+            .importServerFiles()
```
mihailradkov: There is a lot of room for improvement. For example these two can be combined under single function +  `verifyNoImportStatus` can receive array of file names.

test-cypress/integration/import/import.server.files.spec.js
```diff
@@ -1,137 +1,72 @@
...
-        cy.confirmTextPresentsInDistinctFileInfo(context, fileToImport);
-
-        // First navigate to create repository
-        cy.navigateToPage('Setup', 'Repositories');
-        cy.deleteRepo(repoId).wait(1000);
-        cy.selectRepo(existingRepo).wait(1000);
-        // Should return to import page
-        cy.navigateToPage('Import', 'RDF');
-        cy.chooseServerFilesTab();
+    it('Test import Server files successfully with changing settings', () => {
```
mihailradkov: I've combined several tests in this one cause there is no need to test each setting separately when we are not testing how the data is actually imported (it's tested what is shown in the popup).

test-cypress/support/commands.js
```diff
@@ -79,13 +87,16 @@ Cypress.Commands.add('selectRDFLinkButton', () => {
...
 
 Cypress.Commands.add('selectRepo', (repoId) => {
-    cy.get('#repositorySelectDropdown')
-        .click({force: true})
-        .then(() => {
-            cy.get('.dropdown-item', {timeout: 1000})
-                .contains(repoId)
-                .click({force: true});
-        });
+    cy.get('#repositorySelectDropdown', {timeout: 30000}).should('be.visible');
```
svilenvelikov: Is the timeout needed? Can we enforce it globally or it will slow down the suite?

test-cypress/support/repository-commands.js
```diff
@@ -0,0 +1,22 @@
+import repoTemplate from '../fixtures/repo-template.json';
+import _ from 'lodash';
```
svilenvelikov: Uhmm actually Cypress wraps the lodash library already. Try `Cypress._.defaultsDeep`

test-cypress/steps/import-steps.js
```diff
@@ -0,0 +1,225 @@
+/**
+ * Reusable functions for interacting with the import page.
+ */
+class ImportSteps {
```
svilenvelikov: I don't like this class very much. It encompasses and mixes logic which is related to the precondition, action and verification steps. Although the steps are very granular and simple, they break the BDD approach which I think we should follow. Also those steps doesn't say much about what's tested and what's not actaully because the assertions are mixed within the actions. Not to mention that with this approach many verifications are now executed with every test in the spec which might not be necessary.

test-cypress/support/commands.js
mihailradkov: Some times the repo dropdown is not rendered at all and I have no idea why, that's why I've added this particular wait.

On the global timeout note: default timeout is 4 seconds, which should be enough for most UI operations (in the workbench) and I don't want to change it globally, only in specific cases.

test-cypress/support/repository-commands.js
mihailradkov: Huh, ok. Will use it.

test-cypress/steps/import-steps.js
mihailradkov: Yes, most of them were commands, I gave them scope with this class.

I find it difficult to separate logic, assertion & wait in Cypress. Especially because assertion & waiting are under the same hood. Cypress bundles default assertions in most of its actions/commands!
I don't see how an action could wait for some condition without an assertion.

So, what do you suggest? Split actions from assertions in separate methods or split and move elsewhere?

test-cypress/integration/import/import.server.files.spec.js
desislava-hristova-ontotext: Yes, in general I agree that we should test all settings in one test.

test-cypress/support/commands.js
desislava-hristova-ontotext: Raised a JIRA (GDB-3768) to check if the problem is in the workbench code.

test-cypress/support/repository-commands.js
```diff
@@ -0,0 +1,22 @@
...
+            'Content-Type': 'application/json'
+        }
+    }).should((response) => expect(response.status).to.equal(201)); // 201 Created
+});
+
+Cypress.Commands.add('deleteRepository', (id) => {
+    const url = REPOSITORIES_URL + id;
+    cy.request('DELETE', url).should((response) => expect(response.status).to.equal(200));
+    // TODO: needed?
+    cy.request({method: 'HEAD', url, failOnStatusCode: false}).should((response) => expect(response.status).to.equal(404));
```
mihailradkov: Create specification that test these two and move the HEAD check there. No need to check every time.

test-cypress/steps/import-steps.js
svilenvelikov: An assertion usually do wait of course and I don't mean we have to separate them. 
> Cypress bundles default assertions in most of its actions/commands! I don't see how an action could wait for some condition without an assertion.

And I got confused by this comment. I don't know what you mean, but cypress doesn't bundle assertions with actions. It performs checks by following some rules before executing any action in order to verify the state of the elements so to be able to report proper errors if need. See this for example: https://docs.cypress.io/guides/core-concepts/interacting-with-elements.html#Visibility
Given this, its not necessary to do assertion for visibility before clicking a button for example because cypress will blow with appropriate error message in either case.
I could comment case by case in the file below what I mean if you want. 
Also I'm OK if you disregard this comment for now because we could improve these on later stages.

test-cypress/steps/import-steps.js
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        cy.get('#import-' + type).should('be.visible');
+
+        return ImportSteps;
+    }
+
+    static openImportURLDialog(importURL) {
+        cy.get('#import-user .import-from-url-btn').click();
+        cy.get('.url-import-form').should('be.visible');
```
svilenvelikov: This check is redundant as cypress ensures element is visible before executing the `type` action

test-cypress/steps/import-steps.js
svilenvelikov: and some example tests from cypress repos :)
 https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/blogs__application-actions/cypress/integration/spec.js

test-cypress/integration/import/import.server.files.spec.js
```diff
@@ -1,137 +1,72 @@
...
-        cy.selectRepo(existingRepo);
-    });
+    let repositoryId;
 
-    afterEach(function () {
-        cy.checkAllServerFiles(true);
-        cy.resetStatusOfUploadedFiles().wait(1000);
-        cy.checkAllServerFiles(false);
-    });
+    const baseURI = 'http://purl.org/dc/elements/1.1/';
```
mihailradkov: To upper case

test-cypress/steps/import-steps.js
mihailradkov: I will disregard it until you give examples how the code should be reorganised, improved etc

test-cypress/steps/import-steps.js
mihailradkov: Removed it

test-cypress/steps/import-steps.js
svilenvelikov: OK. Will discuss it when I come back.

PR #36, opened 2019-08-20
Refactoring of class hierarchy tests

svilenvelikov:


test-cypress/integration/explore/class.hierarchy.spec.js
```diff
@@ -1,180 +1,179 @@
...
+        // value.
+        return cy.get('.rz-pointer[role="slider"]')
+            .then(($element) => $element.prop('aria-valuenow'));
+    }
+
+    function searchForClass(name) {
+        cy.get('.toolbar-holder .icon-search')
+            .click()
+            .then(() => {
+                cy.get('#search_input_value').type(name).type('{enter}').wait(500);
```
mihailradkov: Document the need of the `wait` or try to properly wait?

test-cypress/integration/explore/class.hierarchy.spec.js
```diff
@@ -1,180 +1,179 @@
...
+
+        // Verify that prefixes are removed from diagram
+        verifyPrefixes(($element) => expect($element.text()).to.not.contain(':'));
     });
 
-    it('Test reload diagram', function () {
+    it('Test focus on diagram', () => {
         let className = ':CabernetSauvignon';
+        findClassByName(className);
+        cy.get('@classInHierarchy').then(verifyClassIsNotExpanded);
```
mihailradkov: This confused me, can't `findClassByName` return the element?

test-cypress/integration/explore/class.hierarchy.spec.js
```diff
@@ -1,180 +1,179 @@
...
+    });
 
-        // Click on "Reload diagram"
-        cy.get('#toolbar > [tooltip="Reload diagram"] > .icon-reload').click({force: true}).wait(200);
+    it('Test reload diagram', () => {
+        // Initial class count is 50
+        // Change the class count to a custom value
+        cy.get('.class-cnt-slider > .ng-isolate-scope')
+            .trigger('click', 'center', {force: true});
```
mihailradkov: I like to document why force is needed.

test-cypress/integration/explore/class.hierarchy.spec.js
```diff
@@ -1,180 +1,179 @@
...
+        return expect(fontSize['font-size']).be.eq('10px');
+    }
+
+    function verifyClassIsExpanded($element) {
+        let fontSize = $element.prop('style');
+        return expect(fontSize['font-size']).not.be.eq('10px');
+    }
+
+    function confirmReload() {
+        cy.get('.modal-footer .confirm-btn').click();
```
mihailradkov: I wonder if ng-animate would case issues in this spec.

test-cypress/integration/explore/class.hierarchy.spec.js
svilenvelikov: I actually have forgotten about this. I'll just try to refactor the `findClassByName` function and remove the wait. Currently the way the class is searched in the diagram is ugly and would cause the tests to fail if the wait is missing because cypress doesn't wait in the `each` loop as I can remember.

test-cypress/integration/explore/class.hierarchy.spec.js
svilenvelikov: It didn't see it failing.

test-cypress/integration/explore/class.hierarchy.spec.js
svilenvelikov: It's because of the way `findClassByName` is implemented. As I said in a previous comment, I'll try to refactor it and then it'll make a return as you suggested.

test-cypress/integration/explore/class.hierarchy.spec.js
svilenvelikov: Honestly I don't know. I decided to leave this action as is because it operates on a third party library realizing the slider and I didn't want to waste time to see why its needed initially.

test-cypress/integration/explore/class.hierarchy.spec.js
svilenvelikov: Actually I can't refactor it because of the way diagram works. Visible classes visualized with circles and their labels are not in any relation with each other. Labels (the text elements) are not descendants of the circles in the DOM tree. The labels of some classes on the other side might be clipped in order to not overlap with other labels if their circle is too small. So the only way to get a class label is to get it through the data property of the element which makes it impossible to craft a simple css selector to grab the desired text element.

test-cypress/integration/explore/class.hierarchy.spec.js
svilenvelikov: As I ended up not refactoring the `findClassByName` this will stay as is. The reason is that it's not possible to return anything else from an `each` loop in cypress. It always yelds the subject on which it was called. It's also not possible to make a local variable store in the function because as we know every cypress command is async and the method will always return before the each loop is completed.

test-cypress/integration/explore/class.hierarchy.spec.js
svilenvelikov: I removed it and it didn't fail. Will see.

PR #39, opened 2019-08-22
Introduced monitor queries tests

svilenvelikov:
Implement Initial state scenario tests

### DIFF
--- a/test-cypress/integration/monitor/monitor.queries.spec.js
+++ b/test-cypress/integration/monitor/monitor.queries.spec.js
@@ -1,0 +1,26 @@
+describe('Monitor Queries', () => {
+
+    let repositoryId;
+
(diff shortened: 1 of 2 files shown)


test-cypress/integration/monitor/monitor.queries.spec.js
```diff
@@ -0,0 +1,24 @@
...
+
+        cy.visit('/monitor/queries');
+    });
+
+    after(() => {
+        cy.deleteRepository(repositoryId);
+    });
+
+    it('Initial state ', () => {
+        cy.get('.no-running-queries-alert').contains('No running queries or updates.');
```
mihailradkov: Maybe also `.should('be.visible')` ? Cause `contains` will check for existence only, I think.

PR #40, opened 2019-08-22
Introduced monitor resources tests

svilenvelikov:
Implement Initial state scenario tests

### DIFF
--- a/test-cypress/integration/monitor/monitor.resources.spec.js
+++ b/test-cypress/integration/monitor/monitor.resources.spec.js
@@ -1,0 +1,55 @@
+describe('Monitor Resources', () => {
+
+    let repositoryId;
+
+    before(() => {
+        repositoryId = 'monitoring-repo' + Date.now();
+        cy.createRepository({id: repositoryId});
+    });
+
+    beforeEach(() => {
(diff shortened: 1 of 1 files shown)


test-cypress/integration/monitor/monitor.resources.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    function getMemoryUsageChart() {
+        return getTabContent().find('.nv-stackedAreaChart');
+    }
+
+    it('Initial state ', () => {
+        // Graphics container should be present
+        getTabsPanel().should('be.visible');
+        // All tabs should be visible
+        let tabs = ['Memory', 'Threads', 'CPU', 'Classes'];
+        getTabsPanel().find('.nav-item').should('have.length', 4).each(($tab, index) => {
```
mihailradkov: An alternative
```
tabs.forEach((tabName, index) => {
  getTabsPanel().find('.nav-item').eq(index).should('be.visible').and('contain', tabName);
})
```
although it needs separate check for the length

svilenvelikov: yeah, that and also this will make 4 selections in the DOM instead of 1

PR #42, opened 2019-08-23
Added Cypress tests for repositories CRUD UI operations

mihailradkov:
Added several classes for easier test interaction.

### DIFF
--- a/src/pages/repositories.html
+++ b/src/pages/repositories.html
@@ -46,13 +46,13 @@
 			</div>
 
 			<div ng-show="hasActiveLocation()" class="mb-3">
-				<div class="alert alert-warning" ng-hide="getRepositories().length || getActiveLocation().errorMsg">
+				<div class="alert alert-warning no-repositories-alert" ng-hide="getRepositories().length || getActiveLocation().errorMsg">
 					There are no repositories in the active location
 				</div>
 				<div ng-show="getRepositories().length" class="mb-1">
 					<table id="wb-repositories-repositoryInGetRepositories" class="table table-hover table-striped">
 						<tbody>
-							<tr
+							<tr class="repository"
 								ng-repeat="repository in getRepositories() | orderBy: ['type === \'system\'', 'id']"
 								ng-class="{'table-info text-secondary': getActiveRepository() == repository.id}">
 								<td width="25">
@@ -73,7 +73,7 @@
 											<strong class="repository-id">{{repository.id}}</strong></span><span class="small" ng-if="repository.title"><small> &middot; {{repository.title}}</small></span>
 									</div>
 								</td>
-								<td width="150" class="text-nowrap">
+								<td width="150" class="text-nowrap repository-actions">
 									<button class="btn btn-link p-0"
 											ng-click="copyToClipboard(repository.externalUrl)"
 											tooltip="Copy repository URL to clipboard">
@@ -85,21 +85,21 @@
 											tooltip-placement="top" tooltip-trigger="mouseenter"></i>
 									</span>
 									<span ng-hide="repository.type === 'system' || getDegradedReason()">
-										<a class="btn btn-link p-0"
+										<a class="btn btn-link p-0 edit-repository-btn"
 											href="repository/edit/{{repository.id}}"
 											tooltip="Edit repository"
 											tooltip-placement="top"
 											tooltip-trigger="mouseenter">
 											<span class="icon-edit"></span>
 										</a>
-										<a class="btn btn-link p-0 download-turtle-config" ng-href="{{getRepositoryDownloadLink(repository)}}"
+										<a class="btn btn-link p-0 download-turtle-config download-repository-config-btn" ng-href="{{getRepositoryDownloadLink(repository)}}"
 												target="_blank"
 												tooltip="Download repository configuration as Turtle"
 												tooltip-placement="top"
 												tooltip-trigger="mouseenter">
 												<i class="icon-download"></i>
 										</a>
-										<button class="btn btn-link p-0 secondary" type="button"
+										<button class="btn btn-link p-0 secondary delete-repository-btn" type="button"
 											tooltip="Delete repository"
 											tooltip-placement="top"
 											tooltip-trigger="mouseenter"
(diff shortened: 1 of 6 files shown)


src/pages/repositories.html
```diff
@@ -85,21 +85,21 @@ <h1>
 											tooltip-placement="top" tooltip-trigger="mouseenter"></i>
 									</span>
 									<span ng-hide="repository.type === 'system' || getDegradedReason()">
-										<a class="btn btn-link p-0"
+										<a class="btn btn-link p-0 edit-repository"
```
svilenvelikov: If we follow the pattern, this should be `edit-repository-btn`

test-cypress/integration/repository/repositories.spec.js
```diff
@@ -0,0 +1,404 @@
...
+            .find('.dropdown-menu .dropdown-item')
+            .should('not.contain', repositoryId);
+    });
+
+    it('should disallow creation of repositories without mandatory settings', () => {
+        createRepository();
+        saveRepository();
+
+        getRepositoryCreateForm().should('be.visible');
+        getRepositoryIdField().should('have.class', 'ng-invalid');
```
svilenvelikov: Maybe its better to check for the placeholder `This field is required` instead relying of the angular class.

test-cypress/integration/repository/repositories.spec.js
```diff
@@ -0,0 +1,404 @@
...
+        editRepository(repositoryId);
+
+        cy.url().should('include', '/repository/edit/' + repositoryId);
+
+        getRepositoryCreateForm().should('be.visible');
+        getRepositoryIdField().should('have.value', repositoryId);
+        getRepositoryTitleField().should('have.value', repoTitle);
+        // OWL-Horst (Optimized) has become 9
+        getRepositoryRulesetMenu().should('have.value', '9');
+        getRepositoryDisableSameAsCheckbox().should('have.value', 'false');
```
svilenvelikov: This can be just `.should('not.be.checked');`

test-cypress/integration/repository/repositories.spec.js
```diff
@@ -0,0 +1,404 @@
...
+        saveRepository();
+
+        createRepository();
+        typeRepositoryId(secondRepoId);
+        saveRepository();
+
+        // Connect to the first repo via the connection icon
+        // Note: Not using within() because the whole row will be re-rendered & detached
+        getRepositoryFromList(repositoryId)
+            .find('.icon-connection-off')
```
svilenvelikov: This could be extracted in a function eventually or at least make the selector a constant. There are some other similar repetitions for other elements in the tests above also.
```javascript
getRepositoryFromList(repositoryId).find('.icon-connection-off')
```

test-cypress/integration/repository/repositories.spec.js
```diff
@@ -0,0 +1,404 @@
...
+    it('should allow to delete existing repository', () => {
+        createRepository();
+        typeRepositoryId(repositoryId);
+        saveRepository();
+
+        selectRepoFromDropdown(repositoryId);
+
+        getRepositoryFromList(repositoryId)
+            .find('.repository-actions .delete-repository')
+            // Forcefully clicking it due to https://github.com/cypress-io/cypress/issues/695
```
svilenvelikov: Have you tried this suggestion
https://github.com/cypress-io/cypress/issues/695#issuecomment-379817133

test-cypress/integration/repository/repositories.spec.js
```diff
@@ -0,0 +1,404 @@
...
+
+        // Check the repo has been deselected and is not present in the repo dropdown menu
+        getRepositoriesDropdown().click().within(() => {
+            cy.get('.no-selected-repository').should('be.visible');
+            cy.get('.dropdown-menu .dropdown-item').should('not.contain', repositoryId);
+        });
+    });
+
+    const REPO_LIST_ID = '#wb-repositories-repositoryInGetRepositories';
+
```
svilenvelikov: You tend to put these functions at the end of the spec. Maybe I'll start placing mine at the end as well.

test-cypress/integration/repository/repositories.spec.js
```diff
@@ -0,0 +1,404 @@
...
+            .last();
+    }
+
+    function editRepository(repositoryId) {
+        getRepositoryFromList(repositoryId)
+            .should('be.visible')
+            .find('.repository-actions .edit-repository')
+            // Forcefully clicking it due to https://github.com/cypress-io/cypress/issues/695
+            .should('be.visible')
+            .and('not.be.disabled')
```
svilenvelikov: Maybe the disabled check is redundant here as the click would blow in case it's not enabled

test-cypress/integration/repository/repositories.spec.js
mihailradkov: The problem is that according to Cypress the DOM is "visible" and "actionable" when in fact the snapshots shows it's not yet rendered or properly rendered (due to the fact animations are not finished I think). So that's why I want to wait until the animation is finished.

test-cypress/integration/repository/repositories.spec.js
mihailradkov: Yes, I still haven't committed these changes

test-cypress/integration/repository/repositories.spec.js
mihailradkov: Well.. I think it's more important to see the spec and what is tested/verified instead of reading some functions with selections, actions etc.

src/pages/repositories.html
mihailradkov: Renamed

test-cypress/integration/repository/repositories.spec.js
mihailradkov: Updated to check for placeholder's text

test-cypress/integration/repository/repositories.spec.js
mihailradkov: Check it now :)

test-cypress/integration/repository/repositories.spec.js
mihailradkov: Added two reusable functions

test-cypress/integration/repository/repositories.spec.js
mihailradkov: Just did and it doesn't work

test-cypress/integration/repository/repositories.spec.js
mihailradkov: The click is forced so I doubt it would blow.

PR #45, opened 2019-08-27
Added Cypress tests for Setup/Namespaces page

mihailradkov:
Added several classes in the DOM for easier selection in the specs.

### DIFF
--- a/test-cypress/integration/setup/namespaces.spec.js
+++ b/test-cypress/integration/setup/namespaces.spec.js
@@ -1,0 +1,352 @@
+describe('Namespaces', () => {
+
+    let repositoryId;
+
+    beforeEach(() => {
+        repositoryId = 'namespaces-' + Date.now();
+        cy.createRepository({id: repositoryId});
+        cy.presetRepositoryCookie(repositoryId);
+        cy.warmRepositoryNamespaces(repositoryId);
+
+        cy.visit('/namespaces');
+
+        waitUntilPageIsLoaded();
+    });
+
+    afterEach(() => {
+        cy.deleteRepository(repositoryId);
+    });
+
+    function waitUntilPageIsLoaded() {
+        // Workbench loading screen should not be visible
+        cy.get('.ot-splash').should('not.be.visible');
+
+        // No active loader
+        cy.get('.ot-loader').should('not.be.visible');
+
+        getAddNamespaceForm().should('be.visible');
+    }
+
+    const DEFAULT_NAMESPACES = {
+        'gn': 'http://www.geonames.org/ontology#',
+        'owl': 'http://www.w3.org/2002/07/owl#',
+        'rdf': 'http://www.w3.org/1999/02/22-rdf-syntax-ns#',
+        'rdfs': 'http://www.w3.org/2000/01/rdf-schema#',
+        'wgs': 'http://www.w3.org/2003/01/geo/wgs84_pos#',
+        'xsd': 'http://www.w3.org/2001/XMLSchema#'
+    };
+
+    it('verify initial state', () => {
+        getNoNamespacesAlert().should('not.be.visible');
+
+        // Should be able to insert new prefix
+        getNamespacePrefixField().should('be.visible').and('not.be.disabled');
+        getNamespaceValueField().should('be.visible').and('not.be.disabled');
+        getAddNamespaceButton().should('be.visible').and('not.be.disabled');
+
+        // Should render a table with some default namespaces
+        getNamespacesTable().should('be.visible');
+        getNamespaces().should('have.length', Object.keys(DEFAULT_NAMESPACES).length);
+
+        // Should provide pagination options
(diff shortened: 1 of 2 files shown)


test-cypress/integration/setup/namespaces.spec.js
```diff
@@ -0,0 +1,359 @@
...
+            .should('have.length', 3);
+        getNamespacesPagination()
+            .should('be.visible')
+            .find('li')
+            // Single page + First & Last buttons
+            .should('have.length', 3);
+
+        // Verify default namespaces are present and interactable
+        const defaultPrefixes = Object.keys(DEFAULT_NAMESPACES);
+        getNamespaces().each(($row, $index) => {
```
mihailradkov: Not sure if it is such a great idea to test that exactly these namespaces are available after repo creation.

test-cypress/integration/setup/namespaces.spec.js
```diff
@@ -0,0 +1,352 @@
+describe('Namespaces', () => {
```
mihailradkov: I can easily test the pagination by importing _https://www.w3.org/TR/owl-guide/wine.rdf_ in one of the tests. WDYT ?

test-cypress/integration/setup/namespaces.spec.js
```diff
@@ -0,0 +1,352 @@
...
+            .and('contain', DEFAULT_NAMESPACES['owl']);
+        getNamespacesHeaderPaginationInfo()
+            .should('be.visible')
+            .and('contain', 'Showing 1 - 1 of 1 results');
+
+        getNamespacesFilterField()
+            .clear()
+            .type('missing_prefix');
+        // The table should still be visible but without any results
+        getNamespacesTable().should('be.visible');
```
mihailradkov: This test will fail until the GDB-3817 issue is fixed

test-cypress/integration/setup/namespaces.spec.js
```diff
@@ -0,0 +1,352 @@
...
+            // Single page + First & Last buttons
+            .should('have.length', 3);
+
+        // Verify default namespaces are present and interactable
+        const defaultPrefixes = Object.keys(DEFAULT_NAMESPACES);
+        getNamespaces().each(($row, $index) => {
+            const expectedPrefix = defaultPrefixes[$index];
+            const expectedNamespace = DEFAULT_NAMESPACES[expectedPrefix];
+
+            cy.wrap($row)
```
svilenvelikov: umm what about using the `root()` function?
```javascript
            cy.wrap($row)
                .find('.namespace-prefix')
                .should('contain', expectedPrefix)
                .root()
                .find('.namespaceURI')
                .should('contain', expectedNamespace)
                .root()
                .find('.select-namespace')
                .should('be.visible')
                .and('not.be.disabled')
                .root()
                .find('.edit-namespace-btn')
                .should('be.visible')
                .and('not.be.disabled')
                .root()
                .find('.delete-namespace-btn')
                .should('be.visible')
                .and('not.be.disabled');
```

mihailradkov: Didn't know about it but I'm not sure it will work with `each()` because the root element is not `$row`.

svilenvelikov: It's the `cy.wrap($row)`'s outcome what's considered as root, so it should be working.

test-cypress/integration/setup/namespaces.spec.js
desislava-hristova-ontotext: Yes, do it.

test-cypress/integration/setup/namespaces.spec.js
desislava-hristova-ontotext: If you do not change the ruleset when you create repository, the namespaces will be the same, so I think it is okay to test exactly these.

test-cypress/integration/setup/namespaces.spec.js
desislava-hristova-ontotext: But I don't think wine.rdf has enough namespaces for multiple pages.

test-cypress/integration/setup/namespaces.spec.js
desislava-hristova-ontotext: Fixed. See: https://github.com/Ontotext-AD/graphdb-workbench/pull/51

test-cypress/integration/setup/namespaces.spec.js
mihailradkov: It has for 2 pages

mihailradkov: Actually.. no :frowning_face: only 9 results after importing. Imma leave it for another time.

test-cypress/integration/setup/namespaces.spec.js
mihailradkov: This will work that way only if it is invoked inside `within()`

PR #46, opened 2019-08-27
Added Cypress autocomplete tests

mihailradkov:
Tested enabling it, suggesting resources in the SPARQL editor and added
tests that verify no auto suggests when autocomplete is not enabled.

Added several selectors in the autocomplete page for easier selections.


mihailradkov:
Pushed RDF ranks tests in this branch by mistake. Will skip 'em

PR #48, opened 2019-08-28
Added Cypress tests for the Swagger documentation page

mihailradkov:
Tested the presence of Swagger's iframe, GraphDB and RDF4J sections and the operations for each controller.

Added custom command for encapsulating iframes content due to Cypress inability to work with them out of the box. The command will yield the frame's body and allow further chaining of commands and assertions.

### DIFF
--- a/test-cypress/integration/help/rest-api.spec.js
+++ b/test-cypress/integration/help/rest-api.spec.js
@@ -1,0 +1,120 @@
+describe('Help / REST API', () => {
+
+    before(() => {
+        // Disables the security (only for this spec) to be able to interact with Swagger's iframe
+        // See https://github.com/cypress-io/cypress/issues/136
+        Cypress.config({chromeWebSecurity: false});
+    });
+
+    beforeEach(() => {
+        cy.visit('/webapi');
+        getSwaggerFrame()
+            .find('#message-bar')
+            .should('be.visible');
+    });
+
+    // TODO: Check explicitly for all controllers (name + summary) ?
+
+    it('should provide documentation and examples for GraphDB APIs', () => {
+        // The GraphDB API section should list the available controllers & each controller should have a heading and possible options
+        verifyControllerList(getGraphDBAPI());
+
+        // All controllers should be collapsed
+        verifyControllerListIsCollapsed(getGraphDBAPI());
+
+        // Examine the import controller
+        verifyControllerOptions('import-controller', 'Data import');
+    });
+
+    it('should provide documentation and examples for RDF4J APIs', () => {
+        // The RDF4J API section should list the available controllers & each controller should have a heading and possible options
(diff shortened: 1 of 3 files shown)


test-cypress/integration/help/rest-api.spec.js
```diff
@@ -0,0 +1,110 @@
+describe('Help / REST API', () => {
+
+    before(() => {
+        // Disables the security (only for this spec) to be able to interact with Swagger's iframe
```
mihailradkov: I observed that this will not always work but I'm not sure if it wasn't something from my end. 
Let's see how it behaves in Travis. If unstable, I'll add it to cypress.json

test-cypress/integration/help/rest-api.spec.js
```diff
@@ -0,0 +1,110 @@
...
+    });
+
+    beforeEach(() => {
+        cy.visit('/webapi');
+        getSwaggerFrame()
+            .find('#message-bar')
+            .should('be.visible');
+    });
+
+    // TODO: Check explicitly for all controllers (name + summary) ?
```
mihailradkov: Currently I test only single controller in each section. It could easily be extended to test each one of them.

test-cypress/integration/help/rest-api.spec.js
```diff
@@ -0,0 +1,110 @@
...
+    function verifyControllerList(controllerSection) {
+        controllerSection.within(($section) => {
+            cy.wrap($section)
+                .should('be.visible')
+                .find('.resource')
+                .should('be.visible')
+                .its('length')
+                .should('be.gt', 0);
+
+            cy.get('.resource .heading')
```
svilenvelikov: Can't this be chained with the above?

mihailradkov: Maybe with `root().find()` ? Will try it

mihailradkov: Added .root() although I don't really see the benefit here because it will make a selection nevertheless, `cy.root().find() === cy.get()`

mihailradkov: I also updated a bit the chains in `verifyControllerOptions`

svilenvelikov: > Added .root() although I don't really see the benefit here because it will make a selection nevertheless, `cy.root().find() === cy.get()`

But why? Isn't the result yeld by the above part, before the root actually `.result` list? Why do you need to return to the section and execute the selection for `.result > .heading` again? 
IMO you could just make `.find('.heading')...` instead of `root()`

mihailradkov: Cause there is `.its()` before that and `.find()` would not work in the chain.

PR #50, opened 2019-08-28
Fixing flaky Cypress tests

mihailradkov:
- Replaced `parentsUntil().last()` with `closest()` or simply `parent()`
- Updated the check in repositories default tests to simply check that the repo is not selected instead of not having a selected repo. For some reason, sometimes in Travis there is a selected repo or the repo dropdown is not yet fully loaded and the old check for no selected repo was failing.
- Added increased timeout in `getModal()` functions to allow the dialog to fully appear without any  angular animations in progress. It should not be needed but it seems in Travis animations are slow.
- Updated saved query tests to forcefully click the action buttons because Cypress sometimes determines them as not visible (having 0x0 dimensions) although they are visible and clickable and the checks for these pass before clicking them.
- Updated the `selectRepo` method in `HomeSteps` to select the repo row before clicking it to avoid the "having 0x0 dimensions" issue.
- Updated the `visitImport` method in `ImportSteps` to preset the repository cookie instead of selecting the  repo using the dropdown menu. This speeds up both user and server import tests and avoid using the old flaky command.
- Added more html classes for easier selection by Cypress
- Reorganized the test for invalid saved queries to avoid closing the dialog before attempting to save a query with duplicated name.
- Forced the selection of repositories from the dropdown menu due to Cypress's '0x0 dimension' issue in Travis.
- Modified the `ImportSteps` to wait modal's visibility and animation.
- Removed visibility check from SPARQL download menu items due to not always passing in Travis...
- Rearranged the SPARQL spec `beforeEach` wait until page is loaded
- Updated the query for the URL test to be more unique.
- Added waiting for splash screen do disappear in the monitoring specs.
- Removed `removeUploadedFiles()` step in the `afterEach()` of importing user data because it was failing the whole suite if a test failed. Added another test to verify this behaviour.
- Added increased timeout in the sparql test `beforeEach` to allow the active to appear. Sometimes in Travis this is very slow..
- Renamed `warmRepositoryNamespaces` to `initializeRepository` and updated the logic to invoke the /size endpoint.
- Fixed the namespaces filtering test to check if proper alert is shown instead of expecting an empty table. Modified the DOM to render the message as an alert. Added proper selector.
- Added visibility check as waiting for the saved query popover every time it is opened.

### DIFF
--- a/test-cypress/integration/home/workbench.home.spec.js
+++ b/test-cypress/integration/home/workbench.home.spec.js
@@ -3,10 +3,10 @@ import HomeSteps from '../../steps/home-steps';
 describe('Home screen validation', () => {
 
     const FOAT_SNIPPET = '@base <http://example.org/> .\n' +
-                       '@prefix foaf: <http://xmlns.com/foaf/0.1/> .\n' +
-                       '<#green-goblin>\n' +
-                           'a foaf:Person ;    # in the context of the Marvel universe\n' +
-                           'foaf:name "Green Goblin" .\n';
+        '@prefix foaf: <http://xmlns.com/foaf/0.1/> .\n' +
+        '<#green-goblin>\n' +
+        'a foaf:Person ;    # in the context of the Marvel universe\n' +
+        'foaf:name "Green Goblin" .\n';
     const GOBLIN_URI = 'http://example.org/#green-goblin';
 
     beforeEach(() => cy.viewport(1280, 1000));
@@ -41,6 +41,9 @@ describe('Home screen validation', () => {
             HomeSteps.verifyCreateRepositoryLink();
 
             let repositoryId = HomeSteps.createRepo();
+            // Initializing the repository to speed up retrieving repository info
+            cy.initializeRepository(repositoryId);
+
             HomeSteps.selectRepo(repositoryId);
             HomeSteps.verifyRepositoryIsSelected(repositoryId);
             HomeSteps.hasRepositoryInfo(repositoryId);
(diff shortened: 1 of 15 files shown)


test-cypress/integration/home/workbench.home.spec.js
```diff
@@ -42,6 +42,9 @@ describe('Home screen validation', function () {
         HomeSteps.verifyCreateRepositoryLink();
 
         let repositoryId = HomeSteps.createRepo();
+        // Warming the namespaces to speed up retrieving repository info
+        cy.warmRepositoryNamespaces(repositoryId);
```
desislava-hristova-ontotext: The delay is because the repository is not initialized. It gets initialized the first time a request to the repo is sent (get namespaces, execute query, get repo size). I suggest to move this to createRepo and send a GET /repositories/{repositoryID}/size to initialize it after creation since get size is the simplest. You can also leave it with the call to get namespaces, but please, rename the function to initialize repository.

test-cypress/integration/sparql/sparql.menu.spec.js
```diff
@@ -703,7 +717,7 @@ describe('SPARQL screen validation', () => {
         });
 
         it('Test URL to current query', () => {
-            const query = 'SELECT ?s ?p ?o WHERE {?s ?p ?o .} LIMIT 100';
+            const query = 'SELECT ?sub ?pred ?obj WHERE {?sub ?pred ?obj .} LIMIT 100';
```
svilenvelikov: but why this was changed? :)

mihailradkov: Because this is the default query but on single line. I wanted it to be more unique just to be sure.

mihailradkov:
1) Setup / RDF Rank should allow to enable the RDF rank with graphs and predicates filters:
     CypressError: Timed out retrying: expected '<span.tag.tag-success.ng-hide>' to be 'visible'

I think the 4secs timeout is not enough here. Let's see when we increase the default to 30secs

PR #56, opened 2019-09-03
Implement change user settings scenario

svilenvelikov:


test-cypress/support/sparql-commands.js
```diff
@@ -0,0 +1,65 @@
+Cypress.Commands.add('pasteQuery', (query) => {
```
mihailradkov: What about reusing them in the sparql spec? Or at least create an issue so we won't forget.

svilenvelikov: This is was WIP branch AFAIR. I'll update where necessary.

test-cypress/integration/setup/my-settings.spec.js
```diff
@@ -78,6 +76,58 @@ describe('My Settings', () => {
...
+        waitUntilSparqlPageIsLoaded();
+        verifySameAsIsActive(false);
+        verifyInferenceIsActive(false);
+
+        cy.pasteQuery('select * where { ?s ?p ?o . } limit 1002');
+        cy.get('#wb-sparql-runQuery').click();
+        cy.executeQuery();
+
+        cy.verifyResultsPageLength(1000);
+        cy.verifyResultsMessage('1,000 of 1,000');
```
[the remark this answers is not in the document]
desislava-hristova-ontotext: Not sure if this is expected to work actually. But we can use the same scenario as testing inference/sameAs in sparql spec. Using OWL-Horst (Optimized) rule set when creating the repository and same results counts expectations since the idea here is not to test inference itself.

hudson-onto:
Build finished.

hudson-onto:

PR #61, opened 2019-09-09
Refactored the acceptance tests script to use GraphDB ZIP distribution

mihailradkov:
Changed to the ZIP distribution of GraphDB because it is much more frequently published.

- Removed compose & custom GraphDB image from test-cypress/
- Removed wait-on because Cypress has internal wait of 60s for page load.

Added workaround for publishing the tests module

- Additionally reorganized graphdb-workbench-cypress executable to change the working directory instead of overriding the spec folders. Updated scripts to use npm test instead of directly invoking Cypress.

Bumped module versions to 1.1.0 due to minor API changes in GraphDB REST


test-cypress/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "graphdb-workbench-tests",
-    "version": "0.1.3",
+    "version": "0.1.4",
```
mihailradkov: This version should match the workbench's

PR #66, opened 2019-09-10
added tests for /connectors

antonsgeorgiev:
Added basic tests for /connectors - lucene

### DIFF
--- a/test-cypress/integration/setup/connectors-lucene.spec.js
+++ b/test-cypress/integration/setup/connectors-lucene.spec.js
@@ -1,0 +1,153 @@
+describe('Setup / Connectors - Lucene', () => {
+
+    let repositoryId;
+    const luceneConnectorName = 'lucene-test-index';
+    const fieldName = 'grape';
+    const connectorPropertyChain = 'http://www.ontotext.com/example/wine#madeFromGrape';
+    const uriType = 'http://www.ontotext.com/example/wine#Wine';
+    const connectorCreateToastMessage = 'Created connector ' + luceneConnectorName;
+    const connectorDeleteToastMessage = 'Deleted connector ' + luceneConnectorName;
+
+    before(() => {
+        repositoryId = 'repo' + Date.now();
+        cy.createRepository({id: repositoryId});
+        cy.presetRepositoryCookie(repositoryId);
+        cy.visit('/connectors');
+    });
+
+    after(() => {
+        cy.deleteRepository(repositoryId);
+    });
+
+    it('Test that no indexes are available', () => {
+        getConnectorsPage()
+            .should('be.visible')
+            .and('contain', 'No connector instances.');
+    });
+
+    it('Create a lucene connector', () => {
+        getNewLuceneConnectorButton()
+            .click();
+        getCreateLuceneConnectorPage()
+            .should('be.visible')
+            .and('contain', 'Create new Lucene Connector');
+        getConnectorNameField()
+            .find('input')
+            .type(luceneConnectorName);
+        getFieldNameField()
+            .find('input')
+            .type(fieldName, {force:true});
+        getPropertyChainField()
+            .find('input')
+            .type(connectorPropertyChain, {force:true});
+        getUriTypes()
+            .find('input')
+            .type(uriType);
+        getOkButton()
+            .click();
+        getConnectorCreationDialog()
+            .should('be.visible')
+            .and('contain', luceneConnectorName);
+        getConnectorStatusToastMessage()
+            .should('be.visible')
+            .and('contain', connectorCreateToastMessage);
+    });
+
+    it('Test connector copy functionality', () => {
+        getConnectorsPage()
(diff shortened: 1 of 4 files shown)


test-cypress/integration/setup/connectors-lucene.spec.js
```diff
@@ -0,0 +1,153 @@
...
+    const luceneConnectorName = 'lucene-test-index';
+    const fieldName = 'grape';
+    const connectorPropertyChain = 'http://www.ontotext.com/example/wine#madeFromGrape';
+    const uriType = 'http://www.ontotext.com/example/wine#Wine';
+    const connectorCreateToastMessage = 'Created connector ' + luceneConnectorName;
+    const connectorDeleteToastMessage = 'Deleted connector ' + luceneConnectorName;
+
+    before(() => {
+        repositoryId = 'repo' + Date.now();
+        cy.createRepository({id: repositoryId});
```
svilenvelikov: I you don't need to have a single repository for the whole spec, then better change the hooks to `beforeEach` and `afterEach` in order to spawn a new repository for each test and avoid possible issues with dependent tests.

test-cypress/integration/setup/connectors-lucene.spec.js
```diff
@@ -0,0 +1,153 @@
...
+            .should('be.visible')
+            .and('contain', luceneConnectorName);
+        getConnectorStatusToastMessage()
+            .should('be.visible')
+            .and('contain', connectorCreateToastMessage);
+    });
+
+    it('Test connector copy functionality', () => {
+        getConnectorsPage()
+            .should('be.visible');
```
svilenvelikov: This is already tested in the `Test that no indexes are available`, so no need to duplicate verificatons.

test-cypress/integration/setup/connectors-lucene.spec.js
svilenvelikov: hmm as I see it, you actually have dependencies between the tests. A connector is created in `Create a lucene connector` and then it's used in the following tests. My suggestion is to join those in a single test which to become a whole scenario: create connector -> copy it -> delete it. Having dependencies in the tests order is a bad practice because it makes them hard for maintainance and reason against. One should really read all the tests in order to understand how they depend on each other and how to not break them if wants to extend the spec somehow.

test-cypress/integration/setup/connectors-lucene.spec.js
```diff
@@ -0,0 +1,153 @@
...
+    function getNewLuceneConnectorButton() {
+        return getConnectorsPage().find('.new-connector-btn');
+    }
+
+    function getCreateLuceneConnectorPage() {
+        return cy.get('.modal-content');
+    }
+
+    function getConnectorNameField() {
+        return getCreateLuceneConnectorPage().find('.connector-name-field');
```
svilenvelikov: Better move the `input` selector in here to encapsulate it fully e.g.
```javascript
 function getConnectorNameField() {
        return getCreateLuceneConnectorPage().find('.connector-name-field input');
}
```
and use it like
```javascript
        getConnectorNameField()
            .type(luceneConnectorName);
```
Same applies to the other form fields: name and property chain

test-cypress/integration/setup/connectors-lucene.spec.js
antonsgeorgiev: Fixed with commit fe274051f6f1c19f89ac97e134e6086d2d98512c

test-cypress/integration/setup/connectors-lucene.spec.js
antonsgeorgiev: Fixed with commit fe274051f6f1c19f89ac97e134e6086d2d98512c

test-cypress/integration/setup/connectors-lucene.spec.js
antonsgeorgiev: Fixed with commit fe274051f6f1c19f89ac97e134e6086d2d98512c

test-cypress/integration/setup/connectors-lucene.spec.js
svilenvelikov: great

test-cypress/integration/setup/connectors-lucene.spec.js
svilenvelikov: This is also applicable to the other fields.

test-cypress/integration/setup/connectors-lucene.spec.js
antonsgeorgiev: Fixed in commit 331d9c9122e4c702a054325ddc791ec429a005dd

test-cypress/integration/setup/connectors-lucene.spec.js
antonsgeorgiev: Fixed in commit 331d9c9122e4c702a054325ddc791ec429a005dd

PR #70, opened 2019-09-11
fixing issues with location tests

svilenvelikov:


svilenvelikov:
We found the cause for the issue with failing tests for the locations controller so this is not needed anymore.

PR #71, opened 2019-09-11
Enabled unit tests and fixed issue

svilenvelikov:
It appeared that fantomjs which is used when tests are run in CI doesn't support latest javascript features. Fixed by not using `let` but `var` in controller.


mihailradkov:
image

PR #79, opened 2019-09-13
Fix test for not finding import button in modal dialog

svilenvelikov:
Animation library applies set of status classes around the base animated class. In this case the base animated class is `fade` and the animation lib applies classes like `in-add`, `in-add-active`, `in-remove`, 
 `in-remove-active`. When animation is complete it seems that just `fade in` classes remains.

### DIFF
--- a/test-cypress/steps/import-steps.js
+++ b/test-cypress/steps/import-steps.js
@@ -241,6 +241,7 @@ class ImportSteps {
         return cy.get('.modal')
             .should('be.visible')
             .and('not.have.class', 'ng-animate')
+            .and('have.class', 'in');
     }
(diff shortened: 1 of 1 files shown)


test-cypress/steps/import-steps.js
```diff
@@ -241,6 +241,9 @@ class ImportSteps {
         return cy.get('.modal')
             .should('be.visible')
             .and('not.have.class', 'ng-animate')
+            .and('have.class', 'in')
+            .find('.modal-footer').should('be.visible')
+            .root();
```
mihailradkov: That's gonna return `html`

PR #103, opened 2019-10-03
GDB-3936 Flaky connector test

desislava-hristova-ontotext:
Seems that there are two elements with class creating-connector-dialog and sometimes the wrong is checked for visibility. From jenkins:

_This element '<div.modal-header.creating-connector-dialog>' is not visible because its parent '<div.ng-hide>' has CSS property: 'display: none'_

Tune the selector with the parent class. We are looking for the element in the modal.

### DIFF
--- a/test-cypress/integration/setup/connectors-lucene.spec.js
+++ b/test-cypress/integration/setup/connectors-lucene.spec.js
@@ -117,7 +117,7 @@ describe('Setup / Connectors - Lucene', () => {
     }
 
     function getConnectorCreationDialog() {
-        return cy.get('.creating-connector-dialog');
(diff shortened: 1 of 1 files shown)


test-cypress/integration/setup/connectors-lucene.spec.js
```diff
@@ -117,7 +117,7 @@ describe('Setup / Connectors - Lucene', () => {
     }
 
     function getConnectorCreationDialog() {
-        return cy.get('.creating-connector-dialog');
+        return cy.get('.modal-content > .modal-header.creating-connector-dialog');
```
svilenvelikov: great catch

PR #113, opened 2019-10-08
Gdb 3934 extend users and security cypress tests

antonsgeorgiev:
Added tests for /users


test-cypress/integration/setup/user-and-access.spec.js
```diff
@@ -59,4 +47,241 @@ describe('User and Access', () => {
...
+        getCreateNewUserButton().click();
+        getUsernameField().type(user);
+        getPasswordField().type(password);
+        getConfirmPasswordField().type(password);
+        getUserRoleRadioButton("roleUser").click();
+        getRepositoryRights(repositoryId, "read").click();
+        getCreateButton()
+            .should('be.visible')
+            .click();
+        findUserInTable(user);
```
svilenvelikov: I guess we can extract these commands in  function like `createUser(role)` and reuse it in the other tests below.

svilenvelikov: I think this is not used `findUserInTable(user);`

test-cypress/integration/setup/user-and-access.spec.js
```diff
@@ -59,4 +47,241 @@ describe('User and Access', () => {
...
+            .and('contain', 'Some functionality is not available because you have no write permission to repository ' + repositoryId);
+        getNoPermissionMessage("rdfrank")
+            .should('be.visible')
+            .and('contain', 'Some functionality is not available because you have no write permission to repository ' + repositoryId);
+        cy.visit('/users');
+        getUserLogoutButton().click();
+        performLogin("admin", adminPassword);
+        getDeleteUserButton(user).click();
+        getDeleteUserConfirmButton().click();
+        getSecuritySwitch().click({force:true});
```
svilenvelikov: This could be extracted in a function for deleting user and reused as well.

test-cypress/integration/setup/user-and-access.spec.js
svilenvelikov: My bad, it's used for checking if user exists in the table

test-cypress/integration/setup/user-and-access.spec.js
antonsgeorgiev: Fixed with commit d5a40ea358aa33047d813e1f1b223285da550018

test-cypress/integration/setup/user-and-access.spec.js
antonsgeorgiev: Fixed with commit d5a40ea358aa33047d813e1f1b223285da550018

hudson-onto:
Build finished.

hudson-onto:

PR #116, opened 2019-10-10
GDB-3810 Show the correct text on the search box when default config is loaded

desislava-hristova-ontotext:


src/pages/graphs-visualizations.html
```diff
@@ -88,7 +88,7 @@ <h1>
                 <div class="card">
                     <div class="card-block">
                         <h3>{{configLoaded.name}}</h3>
-                        <p>Search for an IRI to view it. This is a user defined advanced configuration.</p>
+                        <p>Search for an IRI to view it. <span ng-show="configLoaded.name != defaultGraphConfig.name">This is a user defined advanced configuration.</span></p>
```
svilenvelikov: Isn't it better `ng-if` to be used instead?

PR #119, opened 2019-10-17
Replace static Swagger 2 with Swagger 3 from NPM

mihailradkov:


hudson-onto:
Build finished.

hudson-onto:

PR #120, opened 2019-10-18
Gdb 3973 create cypress tests for graph replacement

antonsgeorgiev:
Added additional tests for graph data replacement under different conditions.


svilenvelikov:
What happened to this MR? Can we merge it?

PR #121, opened 2019-10-18
Define modules as plugins and load them on demand

svilenvelikov:
* Part 1: Lazy loading implementation + defining plugins for all modules except the cluster management which is dynamically imported anyway.
  * Implement lazy loading for the modules based on the route state.
  * Define plugins for the route provider for all modules.
  * Define plugins for the menu for all modules.
  * Add /docs folder and developers-guide.md
* Part 2: Refactor all modules to untie them from each other in order to allow indipendent deployment of any module. The only dependencies usually should be third party libraries, angularjs stuff, workbench common components - like the rest clients, generic directives, utilities and so on.
* Other changes:
  * Extracted RdfRanRestService.
  * Refactored paginations directive to decouple it from the namespaces and export modules. Also moved it as a generic directive in a separate package.
  * Removed some unused route states and page templates.
  * Removed all empty directive definitions in modules.
  * Merged modules with their dependencies where they were set in separated files.
  * Removed many unused dependencies injected in components.

Jira: https://ontotext.atlassian.net/browse/GDB-3947


src/js/angular/core/directives.js
```diff
@@ -258,9 +258,9 @@ function searchResourceInput($location, toastr, ClassInstanceDetailsService, Aut
             empty: '='
         },
         templateUrl: 'js/angular/core/templates/search-resource-input.html',
-        link: function ($scope, scope, element) {
-            scope.autoCompleteStatus = undefined;
-            scope.autoCompleteWarning = false;
+        link: function ($scope, element, attrs) {
```
svilenvelikov: Renamed attributes to match what they really are and not being confusing.

src/js/angular/core/services.js
```diff
@@ -243,38 +243,3 @@ function ClassInstanceDetailsService($http) {
...
-} catch (e) {
-    // Fallback to simple validation, works in all browsers
-    iriRegExp = new RegExp("^[a-z](?:[-a-z0-9\\+\\.])*:", "i");
-}
-
-UtilService.$inject = ['toastr'];
-
-function UtilService(toastr) {
```
svilenvelikov: Moved these in the new UriUtil and Notifications services respectively

src/js/angular/directives/paginations.js
```diff
@@ -0,0 +1,7 @@
+const exportDirectives = angular.module('graphdb.framework.directives.paginations', []);
```
svilenvelikov: This directive was previously in angular/export/directives.js but was bound to the export and namespaces modules at the same time, so it was cleaned up and moved here as it's now a generic one.

src/js/angular/externalsync/app.js
```diff
@@ -11,22 +11,9 @@ const modules = [
     'graphdb.framework.core.directives'
 ];
 
-const connectors = angular.module('graphdb.framework.externalsync', modules);
+const externalsync = angular.module('graphdb.framework.externalsync', modules);
```
svilenvelikov: I maybe shouldn't have ranamed this as it was actually the connectors module and it was the package name which was incorrect I think. @desislava-hristova-ontotext

src/js/angular/security/controllers.js
```diff
@@ -2,6 +2,20 @@ import 'angular/core/services';
 import 'angular/security/services';
 import 'angular/rest/security.rest.service';
```
svilenvelikov: I'll probably extract most of these constants in enums to reuse them, because currently such literals are used in many different places.

src/js/angular/stats/app.js
```diff
@@ -2,57 +2,6 @@ import 'angular/core/services';
...
-            + 'running the GraphDB Workbench as well generate a detailed server report file you can use to hunt down issues'
-        }).when('/webapi', {
-            templateUrl: 'pages/webapi.html',
-            title: 'REST API documentation',
-            helpInfo: 'The REST API view documents the available public RESTful endpoints and '
-            + 'provides an interactive interface to execute the requests.'
-        });
-
-        const getDocLink = function () {
-            const pi = $menuItemsProvider.getProductInfo();
```
svilenvelikov: I removed the relation between the menuItemsProvider and these functions and passed the productInfo directly to the menu items during the rendering.

src/template.html
```diff
@@ -108,25 +108,25 @@
...
-    <li ng-repeat="item in menu" ng-if="hasRole(item.role)"
+    <li ng-repeat="item in menu track by $index" ng-if="hasRole(item.role)"
         ng-class="{open: $index === selected, clicked: clicked && $index === selected}"
         class="menu-element" ng-click="checkSubMenuPosition($index)">
         <div class="menu-element-root" ng-if="item.children.length" ng-click="select($index, $event, clicked);">
             <span ng-class=item.icon class="menu-item-icon"></span>
             <span class="menu-item">{{item.label}}</span>
         </div>
-        <a class="menu-element-root" ng-href="{{item.hrefFun ? item.hrefFun() : item.href}}" ng-if="!item.children.length"
+        <a class="menu-element-root" ng-href="{{::item.hrefFun ? item.hrefFun(productInfo) : item.href}}" ng-if="!item.children.length"
```
svilenvelikov: Added one time binding here an to the submenus below to prevent perpetual calculation of the hrefs on every digest cycle.

src/js/angular/externalsync/app.js
desislava-hristova-ontotext: But the module is externalsync now? I prefer externalsync actually, but this is not so important, so let's choose one and make it consistent.

svilenvelikov: I see. OK, I'll keep it `externalsync` then.

src/js/angular/settings/controllers.js
```diff
@@ -115,8 +108,8 @@ function RegisterLicenseCtrl($scope, LicenseRestService, $location, $modal, toas
                 .success(function (licenseCode) {
                     sendLicenseToValidateAndActivate(licenseCode);
                 }).error(function () {
-                    toastr.error('Could not upload file');
-                });
+                toastr.error('Could not upload file');
```
mihailradkov: Formatting seems wrong here

src/js/angular/directives/queryeditor/query-editor.controller.js
```diff
@@ -1,5 +1,5 @@
 angular
-    .module('graphdb.framework.sparql.controllers.queryeditor', [])
+    .module('graphdb.framework.directives.queryeditor.controllers', [])
```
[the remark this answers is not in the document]
svilenvelikov: Well, it seems a bit ugly, but as long as those controllers are related with the queryeditor directive, I decided to make it this way.

src/js/angular/settings/controllers.js
svilenvelikov: I will format it in my next MR

PR #124, opened 2019-10-25
Gdb 3888 extract services and utilities

svilenvelikov:
Extract new services and utilities:
* Renamed workbench-core to error-utils and moved in utils package.
* Moved angularCancelOnNavigate module in core package.
* Moved repositories service to core to prevent module dependencies.
* Extracted and moved jwtAuth service and unauthorized interceptor in the core package.
* Moved query editor related directives in the core package.
* Fix paginations module name.
* Extract constants.
* Moved the packaging info in the guide.
* Extracted dependencies service endpoint in a rest client.
* Extract rest service for license info.
* Hide the angular's localStorageService behind LocalStorageAdapter implementation and extracted constants for all used keys.
* Fixing sonar issues.

Jira: https://ontotext.atlassian.net/browse/GDB-3888


test-cypress/integration/repository/repositories.spec.js
```diff
@@ -22,7 +22,7 @@ describe('Repositories', () => {
         getCreateRepositoryButton().should('be.visible').and('not.be.disabled');
     }
 
-    it('should allow creation of repositories with default settings', () => {
+    it.only('should allow creation of repositories with default settings', () => {
```
mihailradkov: Revert?

test-cypress/integration/explore/similarity.spec.js
```diff
@@ -433,14 +433,7 @@ describe('Similarity screen validation', () => {
     }
 
     function setNewQuery(newQuery) {
```
mihailradkov: There is no point in having the function now.

src/js/angular/utils/local-storage-adapter.js
```diff
@@ -0,0 +1,59 @@
+angular
+    .module('graphdb.framework.utils.localstorageadapter', [])
+    .factory('LocalStorageAdapter', LocalStorageAdapter)
+    .value('LSKeys', {
+        'TABS_STATE': 'tabs-state',
+        'TABS_STATE_CURRENT_ID': 'tabs-state-currentid',
+        'TABS_STATE_MAXID': 'tabs-state-maxid',
+        'VIEW_MODE': 'viewMode',
+        'CLASS_HIERARCHY_HIDE_PREFIXES': 'classHierarchy-hidePrefixes',
```
mihailradkov: Ideally, these should be in their corresponding modules I guess.

test-cypress/integration/repository/repositories.spec.js
svilenvelikov: pff thx I'll

src/js/angular/utils/local-storage-adapter.js
svilenvelikov: Yeah, although many of these keys are used in multiple modules, so I either should have them duplicated in every module or I should introduce a way to provide/plug them which I guess might lead to some timing/synchronization issues.

PR #129, opened 2019-11-04
Describe the new plugin system in the developers guide

svilenvelikov:


docs/developers-guide.md
```diff
@@ -1,8 +1,190 @@
...
+    $routeProvider.when(route.url, {
+        controller: route.controller,
+        templateUrl: route.templateUrl,
+        title: route.title,
+        helpInfo: route.helpInfo,
+        reloadOnSearch: route.reloadOnSearch !== undefined ? route.reloadOnSearch : true
+    });
+});
+```
+Above is an example of extension point in the workbench which allows registering modules that can be
```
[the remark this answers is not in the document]
svilenvelikov: A plugin definition is identified as a minimum by the extension point to which it's registered. The order and priority are optional attributes. I guess you refer the 'name' attribute from the examples below. It's just an example of a plugin definition with some attributes that allows to explain the concept clearly.

docs/developers-guide.md
```diff
@@ -1,8 +1,190 @@
 # Developers Guide
 
+
+
 ## Extending the GraphDB Workbench
 
+## Plugin system
+
+Since v9.1, GraphDB workbench features a plugin system which allows components to be plugged in
```
mihailradkov: Kinda misleading I think. That's not the workbench version.

docs/developers-guide.md
```diff
@@ -1,8 +1,190 @@
...
+PluginRegistry {
+    add(extensionPoint:String, pluginDefinition:[Object|Array]),
+    get(extensionPoint:String):Array,
+    clear(extensionPoint:String),
+    listModules()
+}
+```
+The second component is the `plugin.js` file which the developer defines for each extension point
+where he needs new behavior or customization.  
+The last one is the `plugins.js` file which is autogenerated and is composed from all `plugin.js`
```
mihailradkov: Maybe mention it's generated by Webpack ?

docs/developers-guide.md
```diff
@@ -1,8 +1,190 @@
...
+// plugins[1] -> `plugin-2`
+```
+If order is not provided, then plugins are loaded in order of their registration which is the order 
+in which the plugin.js files are processed during the application bundling.
+
+* The `priority` attribute is optional and can be used when an ordered plugin should be overridden 
+by another plugin. If not provided, priority for every plugin is set by default to `0`. Having two 
+plugins with same order but different priority results in overriding the lower priority plugin by 
+the other. If there are two plugins with equal order and priority, then an error is thrown to warn 
+the developer for the issue.
```
mihailradkov: What if one of them is disabled? Will an error occur then?

docs/developers-guide.md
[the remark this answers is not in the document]
desislava-hristova-ontotext: Yes, the name is what I mean. That also my understanding that the name is just a sample attribute, but I found it a bit misleading.

docs/developers-guide.md
desislava-hristova-ontotext: I agree, we can use the workbench version here 1.2.

svilenvelikov: :D my bad

docs/developers-guide.md
svilenvelikov: It's described in the following sections

docs/developers-guide.md
svilenvelikov: No, it won't.

docs/developers-guide.md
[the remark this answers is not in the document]
svilenvelikov: I can change it to `id` I guess.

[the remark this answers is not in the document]
desislava-hristova-ontotext: Or something that does not look like identifier, maybe label?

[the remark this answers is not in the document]
svilenvelikov: OK, I'll change them in a minute

PR #142, opened 2019-11-20
Stub external resource to prevent tests to fail

svilenvelikov:


hudson-onto:
Build finished.

hudson-onto:

PR #150, opened 2019-11-29
GDB-4123 Register unauthorized interceptor in core

desislava-hristova-ontotext:
Currently the interceptor is attached when you access one of the user endpoints (login, user*). This means that the interceptor is attached when you hit the login endpoint  after securityInit is triggered (in security initialization). This is error prone since if you do not hit the login page for some reason (error during security init), the interceptor will not be attached.

### DIFF
--- a/src/js/angular/security/app.js
+++ b/src/js/angular/security/app.js
@@ -1,7 +1,6 @@
 import 'angular/core/services';
 import 'angular/core/directives';
 import 'angular/security/controllers';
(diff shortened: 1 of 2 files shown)


src/js/angular/security/app.js
```diff
@@ -14,12 +13,3 @@ const modules = [
 ];
 
 const securityApp = angular.module('graphdb.framework.security', modules);
```
svilenvelikov: We don't need the `securityApp` variable already.

PR #154, opened 2019-12-04
Gdb 4053 respect negotiate authentication

desislava-hristova-ontotext:
When security is on and we do not have valid principal and auth token (current user session), try to authenticate with kerberos. Send a request during security initialization to GDB to get currently authenticated user if present and if it is authenticated with negotiate header leave security to negotiate mechanism. This means leave the browser set the 'Authorization' header obtained from kerberos.
This mimics in the workbench automatic user login. In these cases logout for the user is not visible. 
Redirect to login when authentication fails.
A cookie 'auth.negotiate' is kept because we need to know that when we have principal but no auth token, this is because authentication is left to negotiate mechanism.

### DIFF
--- a/webpack.config.dev.js
+++ b/webpack.config.dev.js
@@ -41,7 +41,9 @@ module.exports = merge(commonConfig, {
             target: 'http://localhost:7200',
             onProxyRes: proxyRes => {
                 var key = 'www-authenticate';
-                proxyRes.headers[key] = proxyRes.headers[key] && proxyRes.headers[key].split(',');
+                if (proxyRes.headers[key]) {
+                    proxyRes.headers[key] = proxyRes.headers[key].split(',');
+                }
             }
         }]
     }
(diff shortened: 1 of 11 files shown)


webpack.config.dev.js
```diff
@@ -31,13 +31,18 @@ module.exports = merge(commonConfig, {
...
         compress: true,
         port: 9000,
         historyApiFallback: true,
         proxy: [{
             context: ['/rest', '/repositories', '/orefine', '/protocol', '/rdf-bridge'],
-            target: 'http://localhost:7200'
+            target: 'http://localhost:7200',
+            onProxyRes: proxyRes => {
+                var key = 'www-authenticate';
+                proxyRes.headers[key] = proxyRes.headers[key] && proxyRes.headers[key].split(',');
```
svilenvelikov: I think it's better to not set the header at all if there is no value in the response for it.

desislava-hristova-ontotext: Changed it to be more readable.

PR #183, opened 2020-03-23
GDB-4289 Autocomplete triple with label predicates

desislava-hristova-ontotext:
Link to the resource view with triple parameter for autocomplete results of type triple

### DIFF
--- a/src/js/angular/core/directives.js
+++ b/src/js/angular/core/directives.js
@@ -354,8 +354,7 @@ function searchResourceInput($location, toastr, ClassInstanceDetailsService, Aut
                     let textResource;
                     if (angular.isUndefined(resource)) {
                         textResource = $scope.searchInput;
-                    }
-                    if (typeof resource === 'object') {
+                    } else if (typeof resource === 'object') {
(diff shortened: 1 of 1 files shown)


src/js/angular/core/directives.js
```diff
@@ -343,37 +346,36 @@ function searchResourceInput($location, toastr, ClassInstanceDetailsService, Aut
...
-                    let textUri;
-                    if (angular.isUndefined(uri)) {
-                        textUri = $scope.searchInput;
+                    let textResource;
+                    if (angular.isUndefined(resource)) {
+                        textResource = $scope.searchInput;
                     }
-                    if (typeof uri === 'object') {
-                        textUri = uri.value;
+                    if (typeof resource === 'object') {
```
teodossidossev: else if maybe?

teodossidossev: otherwise textResource will become undefined when resource is undefined

desislava-hristova-ontotext: Fixed

PR #186, opened 2020-03-25
GDB-4386 Fix minimist vulnerability

desislava-hristova-ontotext:


package.json
```diff
@@ -62,6 +62,7 @@
         "less": "^3.10.3",
         "less-loader": "^5.0.0",
         "mini-css-extract-plugin": "^0.8.0",
+        "minimist": "^1.2.5",
```
svilenvelikov: You should update the coveralls library which requires this library. Revert these changes and execute `npm update coveralls`

PR #206, opened 2020-04-03
GDB-2088 Added possibility for user to explore class hierarchy and dependencies for particular graph

sava-savov-ontotext:


src/css/dependencies.css
```diff
@@ -219,3 +219,11 @@ svg {
     font-size: 14px;
 }
 
+#btnGraphsGroup[aria-expanded="false"] {
+    background-color: white;
+    border: none;
+}
+
+#btnGraphsGroup[aria-expanded="true"] {
+    none: inherit;
```
svilenvelikov, line 228: I haven't heard of such css property `none`.

src/js/angular/repositories/controllers.js
```diff
@@ -270,6 +271,16 @@ function LocationsAndRepositoriesCtrl($scope, $modal, toastr, $repositories, Mod
     };
 
     $repositories.init();
+
+    function removeCachedGraphsOnDelete(repoId) {
+        angular.forEach(LocalStorageAdapter.keys(), function (key) {
+            // remove everything but the hide prefixes setting, it should always persist
+            if (key.startsWith("classHierarchy-selectedGraph-" + repoId) || key.startsWith("dependencies-selectedGraph-" + repoId)) {
```
svilenvelikov, line 278: These `"classHierarchy-selectedGraph-" + repoId` and `"dependencies-selectedGraph-" + repoId` can be computed once outside of the loop.

src/js/angular/rest/graph-data.rest.service.js
```diff
@@ -136,4 +153,32 @@ function GraphDataRestService($http) {
             }
         });
     }
+
+    function resolveGraphs(scope, selectedGraph) {
```
svilenvelikov, line 157: I have a couple of considerations here. 
* Services should not deal with UI stuf in either way! Never :)... The problem with this is that services are singletons in angular and we might easily reach to some nasty memmory leaks and degraded performance. Some issues which @teodossidossev is currently dealing with. So, my advice is to remove this `scope` parameter and just pass the resolved values from the method clients e.g. pass `isActiveRepository` and `activeRepository` paramaters instead.
* I don't get the idea of using the `RDF4JRepositoriesRestService` in this `GraphDataRestService`. Why don't you move this method in RDF4JRepositoriesRestService if you use the same endpoint method?
*

src/js/angular/rest/graph-data.rest.service.js
```diff
@@ -136,4 +153,32 @@ function GraphDataRestService($http) {
...
+                        if (binding.contextID.type === "bnode") {
+                            binding.contextID.value = '_:' + binding.contextID.value;
+                        } else if (binding.contextID.type === "default") {
+                            binding.contextID.uri = "http://www.openrdf.org/schema/sesame#nil";
+                        } else {
+                            binding.contextID.uri = binding.contextID.value;
+                        }
+                    });
+                    graphs.results.bindings.unshift(selectedGraph);
+                    scope.graphsInRepo = graphs.results.bindings;
```
svilenvelikov, line 180: And this should not happen as well `scope.graphsInRepo = graphs.results.bindings;`. This can be the return value and the scope.graphsInRepo assignment to happen in the controller.

src/js/angular/rest/graph-data.rest.service.js
```diff
@@ -136,4 +153,32 @@ function GraphDataRestService($http) {
             }
         });
     }
+
+    function resolveGraphs(scope, selectedGraph) {
+        if (scope.getActiveRepository() !== "") {
```
svilenvelikov, line 158: As long as this service method is called in a promise, this check here might cause the promise to be resolved with `undefined`.
Which might be OK actually as you don't actually expect a return value at all due to the side effect on the scope.

src/pages/rdfClassHierarchyInfo.html
```diff
@@ -10,8 +10,27 @@
 
 <div class="toolbar-holder pull-right mt-1" id="toolbar"
 	ng-class="{ 'pushed-toolbar': showClassInfoPanel }"
-	ng-show="hasClassHierarchy() && !loader">
-
+	ng-show="hasClassHierarchy() && !loader || !isAllGraphsSelected()">
+    <div id="selectGraphDropdown" class="btn-group" role="group" ng-show="graphsInRepo.length > 2">
```
svilenvelikov, line 14: I'd have this dropdown implementation as a directive and reuse it here and in the dependencies view above as it's actually the same, but this is out of scope I guess :)

src/js/angular/rest/rdf4j.repositories.rest.service.js
```diff
@@ -84,4 +85,38 @@ function RDF4JRepositoriesRestService($http, $repositories) {
...
+                        }
+                    });
+                    graphs.results.bindings.unshift({
+                        contextID: {
+                            type: "all",
+                            value: "All graphs",
+                            uri: ""
+                        }
+                    });
+                    return  graphs.results.bindings;
```
svilenvelikov, line 118: Are you sure this is correct? You now return just bindings whereas you expect graphs as i can see.

sava-savov-ontotext:
Here I'm sure

PR #210, opened 2020-04-27
Fix test promise issue

teodossidossev:
Jira: https://ontotext.atlassian.net/browse/GDB-4403


test-cypress/steps/import-steps.js
```diff
@@ -191,21 +191,20 @@ class ImportSteps {
         ImportSteps.getServerFileElement(fileToSelect).find('.import-status .import-status-info').then(infoIconEl => {
             cy.wrap(infoIconEl).should('be.visible');
             cy.wrap(infoIconEl).trigger('mouseover');
+            cy.wait(1000)
```
svilenvelikov, line 194: The test suite before we opensource the workbench was full of unconditional waits like these and it was brittle like hell. We managed to stabilize it by rewriting most of it and removing almost all such waits that are a bad practice as you know. Now, you either document a good reason for these new waits to remain or find another way to assure that the popover is rendered.

teodossidossev: 1 removed, 1 set the wait time to 0 millis - to play a role of a timeout

PR #211, opened 2020-04-30
GDB-4497 Fix jquery vulnerabilities

teodossidossev:


teodossidossev:
The new version of jQuery 3.5.0 breaks projects that depend on `hasOwnProperty()` function availability such Bootstrap. 
There is an issue for https://github.com/jquery/jquery/issues/4665
Put this MP OnHold until v3.5.1 is released

PR #214, opened 2020-05-05
GDB-4400 Provides font auto resize to match parent directive

teodossidossev:
The directive manipulates the font size of DOM elements that they fill the width of a parent element. Desired max and min font size can be passed as argument, or get from source. Works with window resize, dynamic recalculates font size on parent changes. Line heights and display property are preserved.

 Related to GDB-4400 Resource view: Reduce the font of nested triples in the resource view

### DIFF
--- a/src/js/angular/explore/fit-text.directive.js
+++ b/src/js/angular/explore/fit-text.directive.js
@@ -1,0 +1,98 @@
+angular
+    .module('graphdb.framework.explore.directives.fittext', [])
+    .directive('fitText', fitTextDirective);
+
+fitTextDirective.$inject = ['$timeout'];
+
+function fitTextDirective($timeout) {
+    return {
+        restrict: 'A',
+        scope: {
+            min: '=',
+            max: '='
+        },
+        link: function (scope, element, attrs) {
+
+            const config = {
+                'debounce': _.debounce,
+                'delay': 100,
+                'loadDelay': 10,
+                'compressor': 1,
+                'min': attrs.fitTextMin || 'inherit',
+                'max': attrs.fitTextMax || 'inherit',
+                'calcSize': 10,
+                'lines': 1
+            };
+
+            const parent = element.parent();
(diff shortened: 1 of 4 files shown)


src/js/angular/explore/fit-text.directive.js
```diff
@@ -0,0 +1,98 @@
+angular
+    .module('graphdb.framework.explore.directives.fittext', [])
```
svilenvelikov, line 2: Is this directive related with the `explore` or it's a generic one? If it's generic isn't it better to move it core/directives?

src/js/angular/explore/fit-text.directive.js
```diff
@@ -0,0 +1,98 @@
...
+        scope: {
+            min: '=',
+            max: '='
+        },
+        link: function (scope, element, attrs) {
+
+            const config = {
+                'debounce': _.debounce,
+                'delay': 100,
+                'loadDelay': 10,
```
svilenvelikov, line 19: What's the purpose of this `compressor` config. I guess this directive is borrwed from somewhere, if so, can you please add link to the source so that we can see some documentation eventually?

src/js/angular/explore/fit-text.directive.js
teodossidossev: I agree - will be better to be generic one.

src/js/angular/explore/fit-text.directive.js
teodossidossev: Is useful when dealing with untypical custom fonts - to correct the ratio calc.
Is reworked from https://github.com/patrickmarabeas/ng-FitText.js

svilenvelikov: Cool. Can you please add the link in the source code here?

teodossidossev: sure

PR #222, opened 2020-07-17
GDB 4705 User query monitoring

tokovach:
Updated plugins to reflect added monitoring rights for regular clients, but no rights for viewing resources.


avataar:
Nice to have but we can do it as separate task:
* Show the username in the query monitoring table

PR #244, opened 2020-09-14
GDB-4965 Handle response status after successful redirect to login page

teodossidossev:


src/js/angular/core/interceptors/unauthorized.interceptor.js
```diff
@@ -11,8 +11,10 @@ angular.module('graphdb.framework.core.interceptors.unauthorized', [
                         // Redirect to login page only if this isn't the endpoint that checks
                         // the externally logged user and when no external authentication is available.
                         // This check is essential for making free access and external auth working.
-                        $rootScope.redirectToLogin();
-                        return $q.reject(response);
```
svilenvelikov, line 15: I don't get this. Can you please describe what do you think the problem is? Given the fix, seems like you suppose that `$rootScope.redirectToLogin()` returns something, which clearly is not the case, so I don't understand how wrapping it in a promise would solve something. I don't see anything asynchronous in the `redirectToLoging` call, but even if there is, Promise.resolve wouldn't wait for it. IMO the call would be resolved immediately with `undefined` and if there are synchronization issues those would still be present I think.
I can be wrong though, but still, can you please shed some more light what's happening here?

teodossidossev: I extended the MR and added comment to describe the changes.

svilenvelikov (approved):

PR #277, opened 2020-10-14
GDB-5092 Stopped InstanceNotFoundException in Query Monitoring when connected to a virtual repository.

sava-savov-ontotext:
Introduced new core-errors attrs "ontop". If such is found and such repo, proper message about the restrictions related with repository of type Ontop will be shown to the user. Also added and JDBC page to the those with warning about Virtual Repository, because latter doesn't have data directory.

### DIFF
--- a/src/js/angular/controllers.js
+++ b/src/js/angular/controllers.js
@@ -321,8 +321,8 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, $cookies, toastr, $locati
      *  If on the other hand attribute "ontop" is found and such repo, proper message about the
      * restrictions related with repository of type Ontop will be shown to the user
      */
-    $scope.setShowWarning = function () {
-        $scope.showWarning = $scope.attrs.hasOwnProperty('write') ||
+    $scope.setRestricted = function () {
+        $scope.isRestricted = $scope.attrs.hasOwnProperty('write') ||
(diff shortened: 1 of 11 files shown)


src/js/angular/controllers.js
```diff
@@ -307,6 +307,25 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, $cookies, toastr, $locati
...
+    };
+
+    /**
+     *  If the attribute "write" is provided and repository other than Ontop one,
+     * directive will require a repository with write access.
+     *  If on the other hand attribute "ontop" is found and such repo, proper message about the
+     * restrictions related with repository of type Ontop will be shown to the user
+     */
+    $scope.setShowWarning = function () {
+        $scope.showWarning = $scope.attrs.hasOwnProperty('write') ||
```
teodossidossev, line 325: It's a misleading name. Why not `isRestricted` or similar

src/js/angular/jdbc/controllers.js
```diff
@@ -49,6 +54,11 @@ function JdbcListCtrl($scope, $repositories, JdbcRestService, toastr, ModalServi
 
             });
     }
+
+    // Check if warning message should be shown or removed on repository change
+    $scope.$on('repositoryIsSet', function () {
```
teodossidossev, line 59: unsubscribe on exit

PR #289, opened 2020-10-26
GDB-5095 Integrate OpenID authentication in GDB Workbench

desislava-hristova-ontotext:
Introduce an openid-auth-service to do the OpenID authentication which obtains as a result auth tokens.
Tha auth token is then set as auth in GDB workbench and attach as Authorization header in each request.
The OpenID service is initialized with the security config from GDB REST API.
Add login with OpenID button only when openID enabled.
Do not execute queries in the main controller before security is initialized, as they redirect to login and break security init itself


avataar (approved):
I've changed how the implicit hack is handled so it should be more universal now.

avataar (dismissed):

avataar (approved):

PR #313, opened 2020-11-12
Various bugs related to security

avataar:
GDB-5225: OpenID login button sometimes not shown

GDB-5228: Misleading message on users screen when not using local database

GDB-5230: Unauthorized interceptor causes error() block to be called on non-errors


avataar:
Reverted GDB-5230: Unauthorized interceptor causes error() block to be called on non-errors as the problem was something else (not using Angular to remove the openid redirect params from URL), fixed now. This also fixes an annoying browser refresh when logging with OpenID. As a side effect also fixed an issue with tutorial slides showing all at once.

Fixed issue with clicking on home while not logged in - it now redirects to login if needed.

Handle null auth and repository and replace them with undefined to prevent sending "Authorization: null" and "X-GraphDB-Repository: null".

src/js/angular/controllers.js
```diff
@@ -503,10 +509,10 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repos
         ];
         $scope.activePage = 0;
         $(".pages-wrapper .page-slide").css("opacity", 100);
-        const withOfParentElm = $(".pages-wrapper")[0].offsetWidth + 200;
+        const widthOfParentElm = $(".main-container")[0].offsetWidth + 200;
         $timeout(function () {
```
teodossidossev, line 513: This timeout may cause a memory leak. There is a task to fix them, so I suggest to mark the method with a TODO.

src/pages/login.html
```diff
@@ -3,7 +3,7 @@
 		<div class="card">
 			<div class="card-block">
 				<h1>Login</h1>
-				<form class="form-horizontal loginForm"  ng-submit="login()" ng-show="isGDBLoginEnabled">
+				<form class="form-horizontal loginForm"  ng-submit="login()" ng-show="isGDBLoginEnabled()">
```
teodossidossev, line 6: use `ng-if` instead `ng-show`

src/pages/login.html
```diff
@@ -25,7 +25,7 @@ <h1>Login</h1>
 						<button id="wb-login-submitLogin" type="submit" class="btn btn-primary btn-lg">Sign in</button>
 					</div>
 				</form>
-                <button ng-show="isOpenIDEnabled" ng-click="loginWithOpenID()" class="btn btn-primary btn-lg">Sign in with OpenID</button>
+                <button ng-show="isOpenIDEnabled()" ng-click="loginWithOpenID()" class="btn btn-primary btn-lg">Sign in with OpenID</button>
```
teodossidossev, line 28: use `ng-if` instead `ng-show`

teodossidossev (changes requested):

desislava-hristova-ontotext (approved):

src/pages/login.html
avataar: done

src/pages/login.html
avataar: done

src/js/angular/controllers.js
avataar: Since this has nothing to do with security or the fixes I left it as it is. I'm sure it won't be difficult to find all $timeout usage when we get to the task you mentioned.

teodossidossev (approved):

PR #319, opened 2020-11-17
GDB-5234: Wrong credentials upon login with valid user (fixed to refe…

avataar:
…r to parent scope regarding username and password model inside ng-if)

Updated version to 1.5.0-TR16.


teodossidossev (approved):
lgtm

PR #349, opened 2020-12-13
GDB-5286 introduce angular translate

svilenvelikov:
* Introduce angular translate service
* Integrated the translate in a couple of files
* Introduced a language selector component which is hidden for now until desided about the look&feel
* Added 2 bundles for english and german


hudson-onto:

src/i18n/locale-de.json
```diff
@@ -0,0 +1,6 @@
+{
```
teodossidossev, line 1: I would prefer the labels to be in capital letters for better detection in the code. They are also constants.

yordanalexandrov: And I would prefer them to be hierarchically structured. But as this is just starting out, it can be changed down the road.

yordanalexandrov (approved):
LGTM

src/i18n/locale-de.json
svilenvelikov: I agree. We already discussed this and I'll change them with the next MR

teodossidossev (dismissed):

src/js/angular/core/services/openid-auth.service.js
```diff
@@ -69,7 +69,7 @@ angular.module('graphdb.framework.core.services.openIDService', modules)
                     $window.location.href = that.getLoginUrl(state, '', returnToUrl, openIDConfig);
                 } else {
                     console.log('oidc: unknown auth flow: ' + authFlow);
-                    toastr.error('Unknown auth flow: ' + authFlow)
+                    toastr.error($translate.instant('openid-auth.unknown.flow', {authFlow: authFlow}));
```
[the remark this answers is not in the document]
sava-savov-ontotext: This will be handled in separate issue

sava-savov-ontotext:
Jenkins retry build

PR #350, opened 2020-12-14
GDB-5242 Added shacl shape graph to graph overview

tokovach:
Replaced YASR bundle to support shacl context/ Added methods to explore/export controller.


hudson-onto:

avataar (approved):

PR #352, opened 2020-12-16
GDB-2068 Expose resource search box so it is always visible

yordanalexandrov:


teodossidossev (approved):

src/js/angular/core/templates/rdfResourceSearchTemplate.html
```diff
@@ -0,0 +1,25 @@
+<link href="css/rdf-resource-search.css?v=[AIV]{version}[/AIV]" rel="stylesheet"/>
+
+<div ng-if="getActiveRepository() && !isLoadingLocation() && hasActiveLocation()" ng-cloak>
+  <div class="card search_inputRDF" id="search_inputRDF">
```
svilenvelikov, line 4: Do you need this ID? Also you better change the underscore with a dash in `search_inputRDF` which you put as a class.

src/js/angular/core/directives/rdf-resource-search.directive.js
```diff
@@ -0,0 +1,66 @@
+angular
```
svilenvelikov, line 1: You can put this directive and its template in its own package under `core/directives`.

src/js/angular/core/templates/rdfResourceSearchTemplate.html
```diff
@@ -0,0 +1,25 @@
+<link href="css/rdf-resource-search.css?v=[AIV]{version}[/AIV]" rel="stylesheet"/>
```
svilenvelikov, line 1: I don't like how in general the versionning is implemented right now. It forces us to do such things like the above instead of just importing the stylesheet in the directive like this
```javascript
import './rdf-resource-search.css';
```

src/js/angular/core/templates/rdfResourceSearchTemplate.html
```diff
@@ -0,0 +1,25 @@
...
+      <search-resource-input class="search-b"
+                             namespacespromise="getNamespacesPromise"
+                             placeholder="Search RDF resources"
+                             autocompletepromisestatus="getAutocompletePromise"
+                             empty="empty">
+      </search-resource-input>
+      <span class="btn btn-link close-rdf-icon"
+            tooltip="Close search" tooltip-placement="bottom"
+            tooltip-trigger="mouseenter">
+      <i class="icon-close icon-2x" aria-hidden="true" ng-click="hideInput()"></i>
```
svilenvelikov, line 15: not properly formatted

src/js/angular/core/templates/rdfResourceSearchTemplate.html
```diff
@@ -0,0 +1,25 @@
...
+<div ng-if="getActiveRepository() && !isLoadingLocation() && hasActiveLocation()" ng-cloak>
+  <div class="card search_inputRDF" id="search_inputRDF">
+    <div class="card-block">
+      <search-resource-input class="search-b"
+                             namespacespromise="getNamespacesPromise"
+                             placeholder="Search RDF resources"
+                             autocompletepromisestatus="getAutocompletePromise"
+                             empty="empty">
+      </search-resource-input>
+      <span class="btn btn-link close-rdf-icon"
```
svilenvelikov, line 12: Can you not use spans as buttons! Its semantically wrong.

src/js/angular/core/directives/rdf-resource-search.directive.js
svilenvelikov, line 1: Not a single test for this directive? We should at least check if it's present on each page and its behavior: open, close, search, etc.

src/js/angular/core/templates/rdfResourceSearchTemplate.html
```diff
@@ -0,0 +1,25 @@
+<link href="css/rdf-resource-search.css?v=[AIV]{version}[/AIV]" rel="stylesheet"/>
+
+<div ng-if="getActiveRepository() && !isLoadingLocation() && hasActiveLocation()" ng-cloak>
+  <div class="card search_inputRDF" id="search_inputRDF">
+    <div class="card-block">
+      <search-resource-input class="search-b"
```
svilenvelikov, line 6: @desislava-hristova-ontotext can we control how this `search-resource-input` directive handles search button click? Can we open results in separate tab? Maybe we should discuss if it's not better that way.

src/js/angular/core/templates/rdfResourceSearchTemplate.html
```diff
@@ -0,0 +1,25 @@
...
+                             empty="empty">
+      </search-resource-input>
+      <span class="btn btn-link close-rdf-icon"
+            tooltip="Close search" tooltip-placement="bottom"
+            tooltip-trigger="mouseenter">
+      <i class="icon-close icon-2x" aria-hidden="true" ng-click="hideInput()"></i>
+    </span>
+    </div>
+  </div>
+  <span class="btn btn-link search-rdf-icon"
```
svilenvelikov, line 19: `search-rdf-icon`? This is not the icon, but the button. Right?

src/js/angular/core/directives/rdf-resource-search.directive.js
```diff
@@ -0,0 +1,66 @@
...
+        element.find('.search_inputRDF').css('top', '-150px');
+        element.find('search-resource-input .view-res-input')[0].value = '';
+        element.find('search-resource-input .view-res-input').change();
+        element.find('.close-rdf-icon')
+            .removeClass('show-search-btn')
+            .addClass('hide-search-btn');
+        $timeout(function () {
+          element.find('.search-rdf-icon')
+              .removeClass('hide-search-btn')
+              .addClass('show-search-btn')
```
svilenvelikov, line 60: It's kind of confusing that you use `show-search-btn` and `hide-search-btn` for toggling both the search and the close buttons. Can you name the class with a more generic name or just use the twitter's utility classes?

svilenvelikov:
I'm ready with the review

src/js/angular/core/templates/rdfResourceSearchTemplate.html
yordanalexandrov: No, I used it at first. I will remove the ID and change the name of the class.

src/js/angular/core/templates/rdfResourceSearchTemplate.html
yordanalexandrov: It now opens in separate tab. I have added an input for the `search-resource-input` directive, to chose whether to open in new tab.

src/js/angular/core/directives.js
```diff
@@ -303,16 +304,30 @@ function searchResourceInput($location, toastr, ClassInstanceDetailsService, Aut
...
+                const param = params.type || 'uri';
+                if ($scope.openInNewTab === 'true') {
+                    openInNewWindowTab(true, params);
+                } else {
+                    $location.path('graphs-visualizations').search(param, params.uri);
                 }
             };
 
+            const openInNewWindowTab = function (visual, params) {
+                const asd = visual ? 'graphs-visualizations' : 'resource';
```
svilenvelikov, line 325: `asd` is very cryptic, don't you think? :)

src/js/angular/core/directives.js
```diff
@@ -303,16 +304,30 @@ function searchResourceInput($location, toastr, ClassInstanceDetailsService, Aut
...
+                const baseUrl = $location.absUrl().substr(0, $location.$$absUrl.indexOf($location.url()));
```
svilenvelikov, line 326: This `$location.$$absUrl` is an internal property and it's not OK to use it directly I think. Better try `$location.absUrl()`

svilenvelikov (approved):

avataar (approved):

PR #357, opened 2020-12-22
Gdb 5276 select rdbms provider in ontop repo creation

sava-savov-ontotext:


desislava-hristova-ontotext:
Jenkins retry build

sava-savov-ontotext:
Jenkins retry build

desislava-hristova-ontotext:
Jenkins retry build

sava-savov-ontotext:
Jenkins retry build

src/js/angular/repositories/ontop-repo-constants.js
```diff
@@ -0,0 +1,240 @@
...
+            userName: "",
+            password: "",
+            driverClass: "",
+            url: ""
+        },
+        urlStart: "",
+        downloadDriverUrl: ""
+}
+
+export const getSupportedDriversData = function ($scope, RepositoriesRestService, toastr) {
```
desislava-hristova-ontotext, line 27: I think the functions should better be extracted as a service. I wouldn't expect functions in a constants file.

desislava-hristova-ontotext (changes requested):

src/js/angular/repositories/controllers.js
```diff
@@ -715,6 +692,14 @@ function EditRepositoryCtrl($scope, $routeParams, toastr, $repositories, $locati
                             $scope.ontopRepoFileNames[key] = $scope.repositoryInfo.params[key].value;
                         }
                     });
+
+                    if ($scope.repositoryInfo.type === ONTOP_TYPE) {
+                        $scope.selectedDriver = Object.assign({},DEFAULT_SELECTED_DRIVER);
+                        Promise.resolve(getSupportedDriversData($scope, RepositoriesRestService, toastr))
```
svilenvelikov, line 698: As long as `getSupportedDriversData` returns a promise, why do you wrap it in a promise again?

src/js/angular/repositories/ontop-repo-constants.js
```diff
@@ -0,0 +1,240 @@
...
+            return 'password';
+        case 'port':
+            return 'number';
+        default:
+            return 'text';
+    }
+}
+
+export const checkForRequiredOntopFiles = function ($scope, RepositoriesRestService, toastr) {
+    if ($scope.repositoryInfo.type === ONTOP_TYPE) {
```
svilenvelikov, line 146: For convenience it's better if a method returns a promise in some cases, to make it always return a promise. Otherwise the method client would need to wrap the result in promise again or to check if the result is promise or not.

src/js/angular/repositories/ontop-repo-constants.js
svilenvelikov: In general I strongly agree with Desi here. But other than that, I really don't see what is the point in moving these methods in another file. If these should be extracted, the best IMO is to be created another component with a template and these methods to be in its own controller. As it is now, you generally bind the view controller with a service, which is very bad.

src/js/angular/repositories/ontop-repo.directive.js
```diff
@@ -0,0 +1,300 @@
+import {getFileName} from "./controllers";
```
svilenvelikov, line 1: You can move this function here. It's not used anywhere else.

src/js/angular/repositories/ontop-repo.directive.js
```diff
@@ -0,0 +1,300 @@
...
+                url: ""
+            },
+            urlTemplate: "",
+            downloadDriverUrl: ""
+        };
+        $scope.ontopRepoFileNames = {};
+        $scope.supportedDriversData = [];
+        $scope.ontopRepoFileLabels =
+            {propertiesFile: 'JDBC properties', obdaFile: 'OBDA or R2RML', owlFile: 'ontology file', constraintFile: 'constraint file'};
+        $scope.ontopRepoFiles = _.keys($scope.ontopRepoFileLabels);
```
svilenvelikov, line 39: No need to use lodash for this, just replace with `Object.keys($scope.ontopRepoFileLabels);`

src/js/angular/repositories/ontop-repo.directive.js
```diff
@@ -0,0 +1,300 @@
...
+        $scope.ontopRepoFiles.forEach(function(key) {
+            if ($scope.repositoryInfo.params[key]) {
+                $scope.ontopRepoFileNames[key] = $scope.repositoryInfo.params[key].value;
+            }
+        });
+        $scope.supportedDriverLabels = {hostName: 'Hostname', port: "Port", databaseName: 'Database name',
+                                        userName: 'Username', password: 'Password', driverClass: 'Driver class', url: 'URL'};
+        $scope.propertiesFileParams = _.keys($scope.supportedDriverLabels);
```
svilenvelikov, line 47: Use `Object.keys` instead of `_.keys`

src/js/angular/rest/repositories.rest.service.js
```diff
@@ -72,6 +75,19 @@ function RepositoriesRestService($http) {
     }
 
     function validateOntopPropertiesConnection(ontopProperties) {
-        return $http.post(`${REPOSITORIES_ENDPOINT}/test-connection`, ontopProperties);
+        return $http.post(`${REPOSITORIES_ENDPOINT}/ontop/test-connection`, ontopProperties);
+    }
+
+    function getSupportedDriversData() {
+        return $http.get(`${REPOSITORIES_ENDPOINT}/ontop/drivers`);
```
svilenvelikov, line 82: Are you sure this service call is OK? As I can see, the rest call fails with status 500 and `{"message":"Request method 'GET' not supported"}`. Better check this!

src/js/angular/repositories/ontop-repo.directive.js
```diff
@@ -0,0 +1,300 @@
...
+            return updateProperties()
+                .then(function () {
+                    const missingRequired = REQUIRED_ONTOP_REPO_PARAMS.filter(function (requiredFile) {
+                        return !$scope.repositoryInfo.params[requiredFile].value;
+                    });
+                    if (missingRequired.length > 0) {
+                        toastr.error('Missing required ontop repo file');
+                        return Promise.reject('Missing required ontop repo file');
+                    }
+                }).catch(function (err) {
```
svilenvelikov, line 172: This catch is useless here. The error will be propagated eiter way if you don't catch it here.

src/js/angular/repositories/ontop-repo.directive.js
sava-savov-ontotext: It is used in the controllers.

sava-savov-ontotext: In let name = **getFileName**(data.params.ruleset.value); in EditRepositoryCtrl

svilenvelikov (approved):

desislava-hristova-ontotext (approved):

PR #360, opened 2021-01-04
GDB-5325 UI issue with uploaded file's info tooltip

yordanalexandrov:


desislava-hristova-ontotext (approved):

src/css/import.css
```diff
@@ -19,3 +19,12 @@
...
     white-space: pre-wrap;
 }
+
+.impex-popover {
+    max-width: 470px;
+}
+
+.impex-popover pre {
+    white-space: pre-wrap;
+    word-wrap: break-word;
```
svilenvelikov, line 29: I don't like this approach because this will actually force the browser to break lines exactly on the spot where the containing box ends which is undesireable. Better use `workd-wrap: normal` in order to achieve normal hyphenation wherever its possible and just then a force break.

yordanalexandrov: I removed the css style for here and also removed the `break-word` class from the `<td>` tag. This will set `word-wrap: normal` as it is the set by the `popover` class.

svilenvelikov: great

src/js/angular/import/templates/filesTable.html
```diff
@@ -63,10 +63,10 @@ <h4>Last import settings</h4>
         <table class="table table-sm m-0" ng-init="settings = filterSettings(file.name)">
             <tr ng-repeat="setting in settings">
                 <td><strong>{{setting[0]}}</strong></td>
-                <td class="break-word" ng-show="setting[0] == 'parserSettings'">
+                <td ng-show="setting[0] == 'parserSettings' || setting[0] == 'requestIdHeadersToForward'">
```
svilenvelikov, line 66: I wonder where this additional check came from `|| setting[0] == 'requestIdHeadersToForward'` and why it was needed?

yordanalexandrov: Left a comment in the issue.
The `requestIdHeadersToForward`'s value is a JSON object but was not displayed in a `pre` tag.

yordanalexandrov:
Jenkins retry build

svilenvelikov (approved):

svilenvelikov:
LGTM in case the failing test is flaky

yordanalexandrov:
Jenkins retry build

PR #373, opened 2021-01-15
GDB-5320: Added support for signature authentication with remote loca…

avataar:
…tion (besides basic). Includes improvement in how we report remote location errors and a small fix that disables repo state refresh when the current location has an error.

Updated links to the documentation to use https:// (there is a link in the new add/edit location modal now)

Restored scripts/run-cypress-tests.sh at it's a convenient way to run the cypress tests via npm run test:acceptance.

### DIFF
--- a/src/js/angular/templates/modal/edit-location.html
+++ b/src/js/angular/templates/modal/edit-location.html
@@ -6,25 +6,58 @@
 </div>
 <form id="addLocationForm" autocomplete="off" name="form">
 	<div class="modal-body">
-		<fieldset class="mb-1">
-			<legend>Optional</legend>
-			<div class="form-group row">
-				<label for="username" class="col-md-4 col-form-label">Username</label>
-				<div class="col-md-8">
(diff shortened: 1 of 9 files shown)


src/js/angular/templates/modal/edit-location.html
```diff
@@ -6,26 +6,59 @@ <h3 class="modal-title">Edit location</h3>
...
-		</fieldset>
-	</div>
+        <div class="form-group row">
+            <label class="col-md-4 col-form-label">Authentication type</label>
+            <div class="col-md-8">
+                <label class="col-form-label" tooltip="No authentication used with remote location">
+                    <input type="radio" name="authType" ng-model="editedLocation.authType" value="none"/>
+                    None
+                </label>
+                &nbsp;
```
desislava-hristova-ontotext, line 16: Isn't there an alternative with bootstrap classes?

desislava-hristova-ontotext (approved):

PR #375, opened 2021-01-18
GDB-5367 Refresh autocomplete status in rdf search box

yordanalexandrov:
Added a `autocompleteStatus` service which sets the autocomplete enabled status to local storage.
The service also monitors local storage for change (supporting different tabs) and broadcasts the status on change.
Added watchers for all search resource inputs.

### DIFF
--- a/src/js/angular/core/directives/rdfresourcesearch/rdf-resource-search.directive.js
+++ b/src/js/angular/core/directives/rdfresourcesearch/rdf-resource-search.directive.js
@@ -13,13 +13,23 @@ function rdfResourceSearchDirective($rootScope, $timeout,
     templateUrl: 'js/angular/core/directives/rdfresourcesearch/templates/rdfResourceSearchTemplate.html',
     restrict: 'AE',
     link: function ($scope, element) {
+      let autocompleteStatus = false;
 
       function refreshRepositoryInfo() {
         if ($scope.getActiveRepository()) {
           $scope.getNamespacesPromise = RDF4JRepositoriesRestService.getNamespaces(
               $scope.getActiveRepository())
               .success(function () {
-                $scope.getAutocompletePromise = AutocompleteRestService.checkAutocompleteStatus();
+                checkAutocompleteStatus();
+              });
+        }
+      }
+
+      function checkAutocompleteStatus() {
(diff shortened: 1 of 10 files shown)


src/js/angular/core/directives/rdfresourcesearch/rdf-resource-search.directive.js
```diff
@@ -13,13 +13,23 @@ function rdfResourceSearchDirective($rootScope, $timeout,
...
               $scope.getActiveRepository())
               .success(function () {
-                $scope.getAutocompletePromise = AutocompleteRestService.checkAutocompleteStatus();
+                checkAutocompleteStatus();
+              });
+        }
+      }
+
+      function checkAutocompleteStatus() {
+        if (!autocompleteStatus) {
```
svilenvelikov, line 29: If `autocompleteStatus=false` then you are going to check it against the GDB and eventually if it's enabled you rise the flag. So far so good. But what happens if `autocompleteStatus=true` and the user tries to disable the autocomplete? You won't notice it here, will you?

svilenvelikov: Maybe better approach is to find if it's possible to expose the autocomplete status globally from the `AutocompleteCtrl` somehow, e.g. through a statefull service or through the global scope.

yordanalexandrov: Refactored code into a service. Added local storage key and watcher. When status is changed it is broadcasted in rootScope.

src/js/angular/core/services/autocomplete-status.service.js
```diff
@@ -0,0 +1,26 @@
...
+
+  function refreshAutocompleteStatus() {
+    AutocompleteRestService.checkAutocompleteStatus()
+        .success(function (autocompleteStatusResponse) {
+          that.setAutocompleteStatus(autocompleteStatusResponse);
+        });
+  }
+
+  $rootScope.$on('repositoryIsSet', refreshAutocompleteStatus);
+}
```
svilenvelikov, line 26: Great. I like it.

svilenvelikov (approved):

PR #382, opened 2021-01-21
GDB-3946 Introduced wb-tags-input class and applied GDB-WB styles for it's elements.

sava-savov-ontotext:


src/css/rdf-details-side-panel.css
```diff
@@ -127,3 +127,35 @@
 	background-color: #eee;
 }
```
desislava-hristova-ontotext, line 128: rdf-details-side-panel is a css with the styles for the side panels, whereas tags inputs can be present on many places, not only in side panels, as is the case with the rdf rank page. This styling should be part of the graphdb bootstrap theme, or come from bootstrap itself. Sava, did you check why there is no styling coming from the bootstrap theme? I would expect such, see https://bootstrap-tagsinput.github.io/bootstrap-tagsinput/examples/. I can see that there is a regression, because in the documentation the screenshots are correct. Please, investigate further with older version of GDB.

desislava-hristova-ontotext (changes requested):

src/css/rdf-details-side-panel.css
desislava-hristova-ontotext: The tags were correct when this feature was introduced, https://graphdb.ontotext.com/documentation/free/exploring-data.html so please find out when and why the regress happened.

desislava-hristova-ontotext (changes requested):

desislava-hristova-ontotext (approved):

PR #388, opened 2021-01-29
GDB-5414 Do not show two search boxes on home page

teodossidossev:
Removed popup search on home page
Added close icon to home page rdf search
Added {esc} key press listener to home page rfd search to clean input

### DIFF
--- a/src/js/angular/core/directives.js
+++ b/src/js/angular/core/directives.js
@@ -262,6 +262,7 @@ function searchResourceInput($location, toastr, ClassInstanceDetailsService, Aut
             uriValidation: '@',
             preserveInput: '@',
             empty: '=',
+            clear: '=?clear',
             openInNewTab: '@',
             preserveSearch: '@',
             radioButtons: '@',
@@ -272,6 +273,7 @@ function searchResourceInput($location, toastr, ClassInstanceDetailsService, Aut
             element.autoCompleteStatus = undefined;
             element.autoCompleteWarning = false;
             $scope.empty = false;
+            $scope.clear = false;
             $scope.searchDisplayType = SEARCH_DISPLAY_TYPE;
             const MIN_CHAR_LEN = 0;
             const IS_SEARCH_PRESERVED = $scope.preserveSearch === 'true';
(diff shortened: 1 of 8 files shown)


src/js/angular/core/directives.js
```diff
@@ -517,6 +517,8 @@ function searchResourceInput($location, toastr, ClassInstanceDetailsService, Aut
                     } else {
                         $scope.searchRdfResourceByEvent(event.target.value, event);
                     }
+                } else if (event.keyCode === 27 && $location.url() === '/') {
```
yordanalexandrov, line 520: I don't like this here. Wouldn't it be better to be in the home controller? Listen for it and prevent propagation.

src/js/angular/core/directives/rdfresourcesearch/rdf-resource-search.directive.js
```diff
@@ -69,7 +73,7 @@ function rdfResourceSearchDirective($rootScope, $timeout,
       }
 
       $scope.onKeyDown = function(event) {
-        if (event.keyCode === 27) {
+        if ($location.url() !== '/' && event.keyCode === 27) {
```
yordanalexandrov, line 76: If you remove it from the DOM on the home page (as I think you should), this will become unnecessary.

src/pages/home.html
```diff
@@ -31,12 +31,16 @@ <h1 class="card-title">{{page.title}}</h1>
         <div class="col-lg-12 home-rdf-resource-search">
             <h3><span  tooltip="View RDF data as a text table or as a visual graph" tooltip-placement="right">View resource</span></h3>
             <div class="card">
-                <div class="card-block">
+                <div class="card-block" ng-keydown="onKeyDown($event)">
```
yordanalexandrov, line 34: You must define the function to be called in the controller! (homeCtrl or something)

yordanalexandrov: If the default behavior is different

yordanalexandrov (changes requested):

src/js/angular/core/directives.js
teodossidossev: agree

src/js/angular/core/directives/rdfresourcesearch/rdf-resource-search.directive.js
teodossidossev: yup

src/pages/home.html
teodossidossev: agree

yordanalexandrov (approved):

PR #399, opened 2021-02-16
GDB-5461 Added resize event listener in rdf-class-hierarchy.directive , which emits classCount event, triggering initClassCountSlider.

sava-savov-ontotext:


teodossidossev (approved):

desislava-hristova-ontotext:
Jenkins retry build

PR #409, opened 2021-03-04
GDB-5416 Added logic, which saves previously selected option from search box into local storage and selects it on reopening the latter. Also added scroll to that element.

sava-savov-ontotext:


desislava-hristova-ontotext (approved):

src/js/angular/core/templates/search-resource-input.html
```diff
@@ -28,8 +28,9 @@
 <span class="autocomplete-hint">Hint: "ab c" matches "abC*", "ab c*" and "ab-c*"</span>
 <div id="auto-complete-results-wrapper">
     <div ng-repeat="autoCompleteUriResult in autoCompleteUriResults track by $index"
+         id="{{'auto_' + $index}}"
          ng-click="searchRdfResourceByEvent(autoCompleteUriResult, $event)" ng-class="{active: activeSearchElm === $index}"
```
yordanalexandrov, line 32: I don't like this approach. You are using the hover styling to show that you have selected an element.
Those are two different indications. When you move the mouse, you don't know anymore what was selected.
You should just use another property to store the selected element's index and add another conditional class to the `auto-complete-results-wrapper` elements.
```
ng-class="{active: activeSearchElm === $index, selected: selectedElementIndex === $index}"
```

yordanalexandrov (changes requested):

yordanalexandrov (approved):

yordanalexandrov:
LGTM

PR #444, opened 2021-06-22
GDB-5879 Handle unsupported repository in repositories view

sava-savov-ontotext:
Added check for unsupported repository type in repositories view, which if true removes all buttons except the delete one and shows warning icon


desislava-hristova-ontotext (approved):

sava-savov-ontotext:
Request is closed due to the newly implementation for GDB 10

PR #446, opened 2021-07-06
GDB-5855-Fixed-not-initialized-variable-in-JDBC

IvanovaD:
Add current query initialization to prevent problems with creating new sql configuration when exception is thrown from backend


avataar (approved):

avataar:
Jenkins retry build

PR #466, opened 2021-07-27
GDB-6007 Handle Adding of identical repos as members

vasilnv:
The relevant changes are in the $scope.ok() function. 
Added a toastr error if the repo that the user is trying to add already exists as a member.

### DIFF
--- a/src/js/angular/repositories/fedx-repo.directive.js
+++ b/src/js/angular/repositories/fedx-repo.directive.js
@@ -41,7 +41,7 @@ function fedxRepoDirective($modal, RepositoriesRestService, toastr, $timeout) {
             }
         }
 
-        $scope.setWritableRepo = function(member) {
+        $scope.setWritableRepo = function (member) {
             let currentWritable = getWritableRepo();
             if (currentWritable && currentWritable.repositoryName !== member.repositoryName) {
                 currentWritable.writable = 'false';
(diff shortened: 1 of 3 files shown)


src/js/angular/repositories/fedx-repo.directive.js
```diff
@@ -218,25 +218,37 @@ function fedxRepoDirective($modal, RepositoriesRestService, toastr, $timeout) {
...
             } else if ($scope.model.repositoryName && $scope.model.store === REMOTE_REPO_STORE) {
                 member = {
-                    store : REMOTE_REPO_STORE,
-                    repositoryName : $scope.model.repositoryName,
-                    repositoryServer : $scope.model.repositoryServer,
-                    username : $scope.model.username,
-                    password : $scope.model.password,
+                    store: REMOTE_REPO_STORE,
+                    repositoryName: $scope.model.repositoryName,
+                    repositoryServer: $scope.model.repositoryServer.charAt($scope.model.repositoryServer.length - 1) === '/' ? $scope.model.repositoryServer.substring(0, $scope.model.repositoryServer.length - 1) : $scope.model.repositoryServer,
```
desislava-hristova-ontotext, line 223: You can use $scope.model.repositoryServer(-1) to get the last character and .slice(0, -1) to get a string until the last character. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice

desislava-hristova-ontotext (changes requested):

desislava-hristova-ontotext (approved):

PR #492, opened 2021-08-26
Resolve GDB-6080 Browser REST API page errors 

vasilnv:


webpack.config.common.js
```diff
@@ -214,12 +214,12 @@ module.exports = {
...
             {
-                test: /\.woff(\?v=\d+\.\d+\.\d+)?$/,
+                test: /\.woff(2)?(\?v=\d+\.\d+\.\d+)?$/,
                 use: {
-                    loader: "url-loader?mimetype=application/font-woff",
+                    loader: "file-loader",
                     options: {
-                        name: '[name].[hash].[ext]',
-                        outputPath: 'font'
+                        name: '[name].[ext]',
```
svilenvelikov, line 221: If you remove the `[hash] ` then the resource wouldn't be a versionable anymore, although the fonts are not changed often.

svilenvelikov, line 221: My another concern is that according to the error described in the ticket and is also visible in the browser it seems that the woff fonts are somewhat "corrupted". Those files should be served as a binary as I understand it and it might be our DGB server that serves them wrong.

svilenvelikov (approved):

PR #500, opened 2021-09-29
GDB-6018 Set log query plan to be always true in backend and hide fro…

desislava-hristova-ontotext:
…m workbench, since it is needed for explain


sava-savov-ontotext:
Jenkins retry build

PR #508, opened 2021-11-01
GDB-6254 IRIs with hashtag in them aren't properly resolved by $locat…

sava-savov-ontotext:
…ion.search() method, because hashtags could be found by $location.hash() method. Added check for hashtag on redirection and if such is found is added into IRI.


sava-savov-ontotext:
Jenkins retry build

desislava-hristova-ontotext (approved):

PR #509, opened 2021-11-02
GDB-6272 Removed starting slashes in URLs of similarity, jdbc and SPARQL rest services and redirect hrefs.

sava-savov-ontotext:


sava-savov-ontotext:
Jenkins retry build

PR #535, opened 2021-12-22
GDB-6446-Clear-Repository-with-enabled-History-Plugin-should-show-app…

IvanovaD:
…ropriate-error-message-at-the-workbench

Added appropriate error message to show when an error occurs while trying to clear repository when history plugin is enabled


IvanovaD:
Jenkins retry build

PR #536, opened 2021-12-22
Gdb 6446 clear repository with enabled history plugin should show appropriate error message at the workbench

IvanovaD:
Added appropriate error message to show when an error occurs while trying to clear repository when history plugin is enabled


IvanovaD:
Jenkins retry build

PR #539, opened 2021-12-30
GDB-6423 Make explain plan syntax highlighted

DenitsaStoianova:
Apply syntax highlighting to the explain plan shown in GDB Workbench by using the same colour as the original SPARQL query

### DIFF
--- a/src/js/lib/yasr.bundled.js
+++ b/src/js/lib/yasr.bundled.js
@@ -55640,9 +55640,6 @@ var root = module.exports = function(parent, options, queryResults) {
 		}
 	}
 
-    yasr.isExplainPlan = function() {
-        return window.editor.getValue().includes('onto:explain') || window.editor.getValue().includes('http://www.ontotext.com/explain');
-    }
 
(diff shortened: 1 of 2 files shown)


src/js/lib/yasr.bundled.js
```diff
@@ -55640,7 +55640,9 @@ var root = module.exports = function(parent, options, queryResults) {
 		}
 	}
 
-
+    yasr.isExplainPlan = function() {
+        return window.editor.getValue().includes('onto:explain') || window.editor.getValue().includes('http://www.ontotext.com/explain');
```
desislava-hristova-ontotext, line 55644: We can move this check directly in highlightExplainPlan() in query-editor-directive.js. so that yasr do not need to know about window.editor.getValue and explain plan details.

desislava-hristova-ontotext (changes requested):

desislava-hristova-ontotext (approved):

DenitsaStoianova:
Jenkins retry build

PR #546, opened 2022-01-07
GDB-6323 Added check if SPARQL Template plugin is active. Moved check…

sava-savov-ontotext:
… for active plugin into new inactive plugin directive and added this directive into all plugin views like Similarity, Autocomplete and RDF rank for consistency


sava-savov-ontotext:
Jenkins retry build

PR #548, opened 2022-01-12
GDB-6423 Make explain plan syntax highlighted

DenitsaStoianova:
Change the color of the explain plan metrics and optimization group


desislava-hristova-ontotext:
Jenkins retry build

PR #562, opened 2022-01-28
GDB-6448 Fix Warn users when they unset the password 1.11

IvanovaD:
added after all removing unnecessary users and switching off security if needed. Also separated create user warning test


IvanovaD:
Jenkins retry build

PR #569, opened 2022-02-10
GDB-6402 The visual graph can display non-existent data

IvanovaD:
Visual graph visualizes all valid IRIs by design, no matter if they are actually existing in the repository.  If they are not existing, then a pop-up message shows. The message is now changed to:
This node either does not exist or has no visible connections. Check your Graph Settings if you expect such. Thus it is more descriptive.


IvanovaD:
Jenkins retry build

IvanovaD:
Jenkins retry build

IvanovaD:
Jenkins retry build

PR #573, opened 2022-02-14
GDB-6423 Change explain plan font and detection

DenitsaStoianova:
Change explain plan highlighting strategy
Change explain plan font to be the same monospaced font as in the editor


DenitsaStoianova:
Jenkins retry build

avataar (approved):

PR #579, opened 2022-02-17
GDB-6280 Changed the regex pattern, that validates all the input fiel…

sava-savov-ontotext:
…ds of type number to mark double values as invalid.


desislava-hristova-ontotext (approved):

desislava-hristova-ontotext:
Jenkins retry build

PR #582, opened 2022-02-22
GDB-6592 Fixed race condition in login controller, which was causing …

sava-savov-ontotext:
…sending non authenticated requests to endpoints in GDB that require access="IS_AUTHENTICATED_FULLY" in GDB security-config.xml. As a result of this is returned unauthorized error and the unauthorized.interceptor rejects the promise.


sava-savov-ontotext:
Jenkins retry build

sava-savov-ontotext:
Jenkins retry build

PR #585, opened 2022-02-23
GDB-6592 Changed deleteUser logic in user-and-access.spec. Added fix …

sava-savov-ontotext:
…for flaky sparql-commands function


sava-savov-ontotext:
Jenkins retry build

desislava-hristova-ontotext (approved):

PR #596, opened 2022-03-14
GDB-5286 - Creates keys for strings in graphexplore

DesislavaBorisova:
The strings are extracted to the locale-en.json file and replaced with their respective keys.


hudson-onto:

PR #602, opened 2022-03-16
GDB-5286 Fixed ontorefine, queries, rdfrank, repositories modules and…

sava-savov-ontotext:
… repository.html and repositories.html


DesislavaBorisova (approved):
Looks good to me!

PR #603, opened 2022-03-17
Gdb 5286 introduce angular translate settings similarity sparql sparql template stats

sava-savov-ontotext:


DesislavaBorisova (approved):
Looks good to me!

PR #620, opened 2022-04-13
GDB-6750-Simultaneous-start/stop-of-plugins-from-2-users-causes-infin…

IvanovaD:
…ite-loading

Also, if 2 tabs are simultenously opened, if a plugin is started and stopped in both tabs, it could lead to infinite loading. Both cases fixed


IvanovaD:
Jenkins retry build

src/js/angular/plugins/controllers.js
```diff
@@ -12,22 +12,58 @@ PluginsCtrl.$inject = ['$scope', '$interval', 'toastr', '$repositories', '$licen
...
+    function cancelTimer() {
+        if (timer) {
+            $interval.cancel(timer);
+        }
+    }
+
     $scope.setPluginIsActive = function (isPluginActive) {
         $scope.pluginIsActive = isPluginActive;
     };
+    const objectsEqual = (o1, o2) =>
```
desislava-hristova-ontotext, line 26: Use underscore's isEqual function instead _.isEqual(o1, o2)

src/js/angular/plugins/controllers.js
```diff
@@ -12,22 +12,58 @@ PluginsCtrl.$inject = ['$scope', '$interval', 'toastr', '$repositories', '$licen
...
+                    if (!objectsEqual($scope.plugins.map((plugin) => ({
+                        name: plugin.name,
+                        enabled: plugin.enabled
+                    })), $scope.pluginsArray)) {
+                        $scope.plugins = $scope.pluginsArray;
+                        if (angular.isDefined($scope.plugins)) {
+                            $scope.displayedPlugins = $scope.plugins;
+                        }
+                        $scope.matchedElements = $scope.plugins;
+                        toastr.info("A plugin status or multiple statuses were changed");
```
desislava-hristova-ontotext, line 47: Do we really need the toast if the status of the plugin is updated and displayed on the page?

src/js/angular/plugins/controllers.js
```diff
@@ -12,22 +12,58 @@ PluginsCtrl.$inject = ['$scope', '$interval', 'toastr', '$repositories', '$licen
...
-                }
-                $scope.matchedElements = $scope.plugins;
-            }).error(function (data) {
-            toastr.error(getError(data));
-        }).finally(function () {
-            $scope.setLoader(false);
-        });
+        timer = $interval(function () {
+
+            PluginsRestService.getPlugins($scope.getActiveRepository())
```
desislava-hristova-ontotext, line 35: This code block is almost identical with the code in initPlugins? Do you really need them both? Is it possible to somehow reuse the logic?

desislava-hristova-ontotext (changes requested):

PR #631, opened 2022-04-28
GDB-6813 Fix issues with cluster management

yordanalexandrov:
- redesigned cluster create page to hide advanced options
 - added validation to rpc address
 - added validations to advanced option numeric fields
 - fixed problem with force delete checkbox not working correctly
 - added French translations for cluster

### DIFF
--- a/src/css/clustermanagement.css
+++ b/src/css/clustermanagement.css
@@ -247,6 +247,23 @@ text.icon-any {
     font-size: 56px;
 }
 
+.text-btn {
+  background: none;
+  border: none;
+  margin: 0;
+  padding-top: .1rem;
+  cursor: pointer;
+}
+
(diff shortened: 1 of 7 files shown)


src/css/clustermanagement.css
```diff
@@ -247,6 +247,23 @@ text.icon-any {
     font-size: 56px;
 }
 
+.text-btn {
```
svilenvelikov, line 250: What's the point of this? Isn't it better to just use a link?

svilenvelikov (approved):

src/css/clustermanagement.css
yordanalexandrov: This is how it is in the repository create view. I left it like that for consistency of look.

PR #640, opened 2022-05-11
GDB-6401 Create cluster from locations

yordanalexandrov:
- refactored create cluster dialog to be able to select nodes from remote locations


src/css/clustermanagement.css
```diff
@@ -415,3 +415,27 @@ div.nodetooltip {
...
 }
+
+.locations-list {
+  border: 1px solid;
+  min-height:8rem;
+}
+
+.location-item {
+  padding: 2px;
+  -webkit-user-select: none;
```
svilenvelikov, line 426: Maybe use just `user-select` and the less will apply needed vendor prefixes.

svilenvelikov: pff ignore this. I now saw that the file is plain css.

svilenvelikov (approved):

PR #642, opened 2022-05-11
GDB-6401 Add tooltip to cluster nodes

yordanalexandrov:
- added a tooltip which is visible when clicked on node to display node status

### DIFF
--- a/src/pages/cluster-management/clusterInfo.html
+++ b/src/pages/cluster-management/clusterInfo.html
@@ -34,31 +34,27 @@
 </div>
 
 <div class="nodetooltip" ng-hide="selectedNode == null">
-    <strong>{{selectedNode.endpoint}}</strong>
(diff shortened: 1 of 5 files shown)


src/pages/cluster-management/clusterInfo.html
```diff
@@ -33,6 +33,35 @@ <h1>
             tooltip-placement="left"><span class="icon-settings settings-icon small"></span></button>
 </div>
 
+<div class="nodetooltip" ng-hide="selectedNode == null">
+    <strong>{{selectedNode.endpoint}}</strong>
+    <hr style="margin-top: 10px; margin-bottom: 10px"/>
+    <div>
+        <div>
```
svilenvelikov, line 40: too much div's

svilenvelikov (approved):

PR #645, opened 2022-05-14
Gdb 6765 implement add remove nodes from cluster

yordanalexandrov:
* implemented action buttons, that are only visible if node is leader, for adding and removing nodes actions
* implemented modal dialogs with own controllers for both functions
* refactored to be able to select nodes from remote locations
* changed visualization to match create cluster dialog


src/js/angular/clustermanagement/controllers.js
```diff
@@ -548,3 +642,73 @@ function AddLocationCtrl($scope, $modalInstance, toastr, productInfo, $translate
...
+AddNodesDialogCtrl.$inject = ['$scope', '$modalInstance', 'data', '$modal', 'RemoteLocationsService'];
+
+function AddNodesDialogCtrl($scope, $modalInstance, data, $modal, RemoteLocationsService) {
+    const clusterConfiguration = angular.copy(data.clusterConfiguration);
+    const clusterModel = angular.copy(data.clusterModel);
+    $scope.nodes = [];
+
+    $scope.clusterNodes = clusterModel.locations.filter((location) => clusterConfiguration.nodes.includes(location.rpcAddress));
+    $scope.locations = clusterModel.locations.filter((location) => !$scope.clusterNodes.includes(location));
+    $scope.locations.forEach((location) => location.isNew = true);
```
svilenvelikov, line 655: Can't you chain this with the filter above?

src/js/angular/clustermanagement/templates/modal/add-nodes-dialog.html
```diff
@@ -0,0 +1,67 @@
+<!-- Start add nodes to cluster modal -->
```
svilenvelikov, line 1: what are these comments?!

src/js/angular/clustermanagement/templates/modal/remove-nodes-dialog.html
```diff
@@ -0,0 +1,44 @@
...
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </div>
+    </div>
+    <div class="modal-footer">
+        <button type="button" class="btn btn-secondary" ng-click="cancel()">{{'common.cancel.btn' | translate}}</button>
+        <button id="wb-remove-nodes-from-cluster-submit" class="btn btn-primary"
+                ng-disabled="!nodesToRemoveCount || (clusterNodes.length - nodesToRemoveCount < 2)"
```
svilenvelikov, line 38: Can you expose a variable in the controller for this `clusterNodes.length - nodesToRemoveCount < 2` and bind to it, then reuse it few lines above.

src/js/angular/clustermanagement/controllers.js
yordanalexandrov: No, as the consumer does not return the array.

src/js/angular/clustermanagement/templates/modal/add-nodes-dialog.html
yordanalexandrov: I saw them somewhere else. They are for consistency :)

yordanalexandrov: removed them

PR #647, opened 2022-05-14
GDB-6868 Cluster UI improvements

yordanalexandrov:
- aligned delete and settings buttons
 - fixed node icons center position on FF
 - fixed different cursor in FF. set cursor to pointer for all browsers as the nodes are clickable
 - added clarification on cluster create dialog
 - set Signature option as default when adding remote locations
 - added break-word to endpoints in node tooltip as they might be big
 - fixed problem with remote locations not visible if create window is reopened or adding a location errors out

### DIFF
--- a/src/i18n/locale-en.json
+++ b/src/i18n/locale-en.json
@@ -59,6 +59,7 @@
             "current_node": "Current node",
             "advanced_options": "Advanced options",
             "advanced_options_tooltip": "Advanced options for cluster configuration",
+            "select_from_locations_hint": "Select nodes from remote locations",
             "errors": {
                 "only_positive_integers": "Enter only positive integers",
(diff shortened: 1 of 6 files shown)


src/i18n/locale-en.json
```diff
@@ -1181,7 +1182,7 @@
     "my.user.placeholder": "my-user",
     "password.required": "Password*",
     "my.password.placeholder": "my-password",
-    "signature.auth.info": "Signature authentication uses the token secret, which must be the same on both GraphDB instances. For more information on configuring the token secret, please refer to ",
```
svilenvelikov, line 1184: BTW, you know that the cypress tests also stub the i18n bundle. Right?

svilenvelikov (approved):

PR #658, opened 2022-05-20
GDB-5293 - add Cypress tests for language switching menu

DesislavaBorisova:
## What?
Two new Cypress test classes added along with a new steps file. Updated existing Cypress test class.

## Why?
There are no existing tests for the language switching functionality in the Workbench, YASR or YASQE. Functions were extracted to avoid duplicate code.

## How?
I created two new language change test classes for SPARQL view and Home view. They check a few labels in each view and change the language before checking the labels have been translated. I also created a steps file where I moved common functions from the new specs and an existing spec.


test-cypress/integration/home/language.change.spec.js
```diff
@@ -0,0 +1,44 @@
+import HomeSteps from '../../steps/home-steps';
+import SparqlSteps from '../../steps/sparql-steps';
+
+describe('Home screen language validation', () => {
+    let repositoryId;
+
+    beforeEach(() => {
+        cy.viewport(1280, 1000);
```
svilenvelikov, line 8: Is there a reason for the view port resolution to be changed?

test-cypress/integration/home/language.change.spec.js
```diff
@@ -0,0 +1,44 @@
...
+            cy.contains('License');
+            SparqlSteps.changeLanguage('French');
+
+            // The text in the labels should change
+            cy.contains('Voir la ressource');
+            cy.contains('Répertoire actif');
+            cy.contains('Requêtes SPARQL sauvegardées');
+            cy.contains('Licence');
+
+            cy.deleteRepository(repositoryId);
```
svilenvelikov, line 41: Better remove this and leave repository deletion in the afterEach callback.

test-cypress/integration/home/language.change.spec.js
```diff
@@ -0,0 +1,44 @@
...
+    context('Language change label checks', () => {
+        it('Default language should be active and language change should affect labels', () => {
+            cy.createRepository({id: repositoryId});
+            cy.initializeRepository(repositoryId);
+            cy.enableAutocomplete(repositoryId);
+            cy.presetRepository(repositoryId);
+
+            // When I visit home page with selected repository
+            HomeSteps.visitAndWaitLoader();
+            // Check some labels are in default language
```
svilenvelikov, line 28: You need to select a concrete element and check for its content. The reason is that `cy.contains('some text')` actually finds the first element in the DOM which contains the text and it might not be the one you expected. So basically you shuld have something like
```
HomeSteps.getViewResourceAsTableButton().should('contain', 'Table');
```
And in the HomeSteps:
```
static getViewResourceAsTableButton() {
  return cy.get('.display-type-table-btn');
}
```

Same applies for the rest of the labels you need to verify. If concrete elements in the application don't have proper css classes or ids, the you can add some in order to be able to select them in the tests.

test-cypress/steps/sparql-steps.js
```diff
@@ -0,0 +1,119 @@
+class SparqlSteps {
+
+    static changeLanguage(language) {
+        // Change the language
+        cy.get('#languageGroupDrop').should('be.visible').click().then(() => {
```
svilenvelikov, line 5: It's OK this way, but generally, you don't need to chain the next command in the promise callback. It should also work like:
```
cy.get('#languageGroupDrop').should('be.visible').click();
cy.get('.dropdown-menu .dropdown-item').contains(language).click();
```

svilenvelikov (changes requested):

test-cypress/integration/home/language.change.spec.js
DesislavaBorisova: This appeared in a few other tests. Wasn't sure if it had a significance that I wasn't getting.

test-cypress/integration/home/language.change.spec.js
DesislavaBorisova: OK, I will correct this.

test-cypress/integration/home/language.change.spec.js
DesislavaBorisova: I understand. I'll see what I can do here. I think I'll need to add some ids.

test-cypress/steps/sparql-steps.js
DesislavaBorisova: OK, I'll keep this in mind.

svilenvelikov (approved):

svilenvelikov:
Jenkins retry build

svilenvelikov (approved):

PR #676, opened 2022-06-08
GDB-6968 Fetch current node not recognized when using ext proxy

yordanalexandrov:
- the call to find which the current node is, was made only once which causes problems when using ext proxy
 - changed to re-fetch current node if there is a leader change

### DIFF
--- a/src/js/angular/clustermanagement/controllers.js
+++ b/src/js/angular/clustermanagement/controllers.js
@@ -49,6 +49,8 @@ function ClusterManagementCtrl($scope, $http, $q, toastr, $repositories, $modal,
     $scope.currentNode = null;
     $scope.clusterModel = {};
     $scope.NodeState = NodeState;
+    $scope.leaderChanged = false;
+    $scope.currentLeader = null;
 
     // Holds child context
     $scope.childContext = {};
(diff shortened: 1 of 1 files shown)


src/js/angular/clustermanagement/controllers.js
```diff
@@ -130,19 +131,31 @@ function ClusterManagementCtrl($scope, $http, $q, toastr, $repositories, $modal,
...
                 }
             })
             .catch(() => {
                 $scope.clusterConfiguration = null;
             });
     };
 
+    function isLeaderChanged(currentLeader, newLeader) {
+        if (!newLeader) {
+            return true;
```
svilenvelikov, line 144: shouldn't this be `false`?

svilenvelikov (approved):

src/js/angular/clustermanagement/controllers.js
yordanalexandrov: If there was an elected leader and he died, I consider it a leader change although there is no new leader at that time.

svilenvelikov: Well. Maybe add this comment in the code, because it's somewhat confusing

yordanalexandrov: added comment

PR #689, opened 2022-06-14
GDB-7021 Fix add remove locations needing page refresh

yordanalexandrov:
- when adding/removing locations they are not refetched/reloaded
 - added locations reload in timer function
 - added locations reload to add/remove actions
 - the locations are reloaded in the service only when there is change
 - amended tests to expect another request for locations

### DIFF
--- a/src/js/angular/repositories/controllers.js
+++ b/src/js/angular/repositories/controllers.js
@@ -154,7 +154,7 @@ function LocationsAndRepositoriesCtrl($scope, $modal, toastr, $repositories, Mod
 
     //Get locations
     function getLocations() {
-        $repositories.getLocations()
+        return $repositories.getLocations()
             .then((locations) => {
                 $scope.locations = locations;
             })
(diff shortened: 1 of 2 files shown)


src/js/angular/repositories/controllers.js
```diff
@@ -153,11 +153,13 @@ function LocationsAndRepositoriesCtrl($scope, $modal, toastr, $repositories, Mod
     };
 
     //Get locations
-    $repositories.getLocations()
-        .then((locations) => {
-            $scope.locations = locations;
-        })
-        .finally(() => $scope.loader = false);
+    function getLocations() {
+        $repositories.getLocations()
```
svilenvelikov, line 157: Eventually return the promise for consistency.

svilenvelikov (dismissed):

src/js/angular/repositories/controllers.js
yordanalexandrov: good point

svilenvelikov (approved):

PR #691, opened 2022-06-15
GDB-6781 Fix problem with explore view and async requests

yordanalexandrov:
- the view expects the security principal from jwtAuth Service, which is an async request
 - sometimes the request is delayed and a NPE is thrown
 - refactored, so if the request is in progress, the controller will wait before using it
 - refactored QueryEditorController and some tests to wait for the new promise

### DIFF
--- a/src/js/angular/core/directives/queryeditor/query-editor.controller.js
+++ b/src/js/angular/core/directives/queryeditor/query-editor.controller.js
@@ -24,12 +24,16 @@ function QueryEditorCtrl($scope, $timeout, toastr, $repositories, $modal, ModalS
     };
 
     let principal;
-    $scope.getPrincipal = function() {
-        return Promise.resolve($jwtAuth.getPrincipal()).then((response) => principal = response);
+    // Wrapped in function for ease of testing
+    $scope.getPrincipal = function () {
+        return $jwtAuth.getPrincipal()
+            .then((response) => {
+                principal = response;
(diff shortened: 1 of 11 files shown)


src/js/angular/core/directives/queryeditor/query-editor.controller.js
```diff
@@ -23,17 +23,14 @@ function QueryEditorCtrl($scope, $timeout, toastr, $repositories, $modal, ModalS
         sameAs: true
     };
 
-    let principal = $jwtAuth.getPrincipal();
+    let principal;
+    $scope.getPrincipal = function() {
+        return Promise.resolve($jwtAuth.getPrincipal()).then((response) => principal = response);
```
svilenvelikov, line 28: I don't like this very much. IMO it'd be better if the $jwtAuth.getPrincipal() method is refactored to return a Promise in either case in order to not force method clients to either check if the result is a promise or not and to wrap it in later scenario. It's sloppy implementation.

yordanalexandrov: Refactored it. Now `jwtAuth.getPrincipal()` always returnes a promise. If the principal is already fetched it will return it wrapped in a promise. If not, it will return a promise which will resolve once secutiry is initialized (there is an event thrown then).

svilenvelikov (approved):

PR #692, opened 2022-06-20
GDB-7042 Set whole line as button in remove node from cluster dialog

yordanalexandrov:
- removed action trigger on list element
 - left button as a decoration


src/js/angular/clustermanagement/templates/modal/add-nodes-dialog.html
```diff
@@ -37,16 +37,17 @@ <h4>{{'cluster_management.cluster_page.remote_locations' | translate}}</h4>
...
-                            <span class="icon-minus"></span>
-                        </button>
+                    <li ng-repeat="node in nodes" class="list-group-item list-group-item-action list-group-item-node">
+                        <a ng-click="removeNodeFromList($index, node)" class="d-flex justify-content-between align-items-center"
+                           tooltip="{{'cluster_management.cluster_page.remove_node_btn_tooltip' | translate}}" tooltip-placement="top">
+                            <div class="location-item">
+                                <div class="logo-image"></div>
+                                {{node.endpoint}}
+                            </div>
+                            <button type="button" class="btn btn-link remove-node-btn">
```
svilenvelikov, line 47: And now, you don't need the button anymore. Just leave the icon if possible.

svilenvelikov (dismissed):

src/js/angular/clustermanagement/templates/modal/add-nodes-dialog.html
yordanalexandrov: I removed them, as with the behaviour change they don't look so good

svilenvelikov (approved):

PR #693, opened 2022-06-20
GDB-7040 Undefine location header for openid requests

desislava-hristova-ontotext:


src/js/angular/core/services/repositories.service.js
```diff
@@ -222,7 +222,6 @@ repositories.service('$repositories', ['$http', 'toastr', '$rootScope', '$timeou
             }
         };
 
-        that.getLocations();
```
yordanalexandrov, line 225: This is probably failing the tests.
A wild guess is that views expect the locations to be fetched repository service when they load.

yordanalexandrov:
Jenkins retry build

yordanalexandrov (approved):
The change itself is ok, but the failing cypress tests must be fixed. I will open an issue for fixing them.

yordanalexandrov (approved):

PR #700, opened 2022-06-24
GDB-7070: Show both HTTP and RPC addresses (as well as the current st…

avataar:
…ate) in the cluster config side panel to avoid user confusion

Also: format the numbers in the same panel using thousands separators


yordanalexandrov (approved):
LGTM

PR #709, opened 2022-07-20
GDB-7206 Change check on error for existing queries when trying to save a query

vasilnv:
Currently if saving a query returns an error, a modal appears stating that the query already exists (even if the problem is something else). Fix the check for existing queries.  
Changed the GraphDB version to use 10.1.0-TR2 as we need it to return status 409 on error.


src/js/angular/core/directives/queryeditor/query-editor.controller.js
```diff
@@ -539,14 +539,14 @@ function QueryEditorCtrl($scope, $timeout, toastr, $repositories, $modal, ModalS
...
-                let msg = getError(data);
+                const msg = getError(data);
                 toastr.error(msg, $translate.instant('query.editor.create.saved.query.error'));
-                // TODO: This condition will always be true
-                if (msg = "Query '" + query.name + "' already exists!") {
-                    query.query = query.body;
-                    const queryExists = true;
-                    $scope.saveQuery(query, queryExists);
+                let queryExists = false;
+                if (msg === "Query '" + query.name + "' already exists!") {
```
svilenvelikov, line 545: This check will fail if the interface is in french for example. Can't we check only for the query name instead of the whole message?!

vasilnv: This message comes from the backend and it is always in English. The one used in the modal is different (query.sample.existing.query.warning in locale-en or locale-fr).

vasilnv:
Jenkins retry build

src/js/angular/core/directives/queryeditor/query-editor.controller.js
vasilnv: Currently there are no other errors containing the query name that can be returned from the controller, so a check only on the query name would be valid. 
But currently if the query exists, the controller returns status 500 with the error message, which I believe should be fixed?

svilenvelikov (approved):

src/js/angular/core/directives/queryeditor/query-editor.controller.js
desislava-hristova-ontotext: We don't have a practise until now to return the error message codes in public APIs and we cannot change this in minor versions. The fact that the status code is 500 is a bug, and can be fixed for 10.1, so let's make the status code 409 and from there the workbench can show whatever message we want to. So a simple check of the status code here would work.

svilenvelikov (dismissed):

vasilnv:
Jenkins retry build

desislava-hristova-ontotext (approved):

PR #713, opened 2022-07-27
GDB-7185 Use the same binding for the inference select, set inference…

desislava-hristova-ontotext:
Use the same binding for the inference select, set inference to explicit always when role is context. Fixes right inference value being sent when loading resources for a context.


yordanalexandrov (approved):
lgtm

PR #715, opened 2022-07-29
GDB-7245: Missing sub menu item

boyan-tonchev:
WHEN:

Add a new sub menu in Help menu. The sub menu is missing.
WHY:

All registered extension points are processed one by one. When an extension point is registered we add it as child (submenu) if it has set parent property. The problem is when we try to register an extension point for a parent which is not already registered. We skip such extension point.
HOW:
Added a collection where to save such extension points and when we register a new parent we will try to register all extension points with missed parents.

### DIFF
--- a/src/js/angular/core/services.js
+++ b/src/js/angular/core/services.js
@@ -79,19 +79,44 @@ function MenuItemsProvider() {
     let productInfo;
 
     const items = [];
+    let itemsWithMissedParent = [];
 
     this.addItem = function (item) {
         if (angular.isUndefined(item.parent)) {
             if (!exists(items, item.label)) {
                 items.push(cloneItem(item, []));
+                // A new parent is registered so we try to process items which parents were missed.
+                this.updateItemsWithMissedParent();
             }
         } else {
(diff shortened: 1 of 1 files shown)


src/js/angular/core/services.js
```diff
@@ -79,20 +79,45 @@ function MenuItemsProvider() {
...
+        if (!angular.isUndefined(parentItem)) {
+            if (!exists(parentItem.children, item.label)) {
+                parentItem.children.push(cloneItem(item, item.children ? item.children : []));
+                return true;
             }
         }
+        return false;
+    }
+
+    this.parentAdded = function () {
```
yordanalexandrov, line 110: Please change the name so it is more descriptive of what it does, not when it's called.

yordanalexandrov (changes requested):

src/js/angular/core/services.js
boyan-tonchev: done

yordanalexandrov (approved):

PR #716, opened 2022-08-01
GDB-7224-"Delete" and "edit" visual graph buttons are missing.

MartinaDimova:


src/pages/graphs-visualizations.html
```diff
@@ -149,7 +149,7 @@ <h5>
                                     <samp class="text-muted small text-overflow d-block" ng-if="config.startMode == 'query'">{{config.startGraphQuery}}</samp>
                                     <samp class="text-muted small text-overflow d-block" ng-if="config.startMode == 'node'">{{config.startIRI}}</samp>
                                 </td>
-                                <td class="text-xs-right" ng-if="isUser() && config.owner === principal().username">
+                                <td class="text-xs-right" ng-if="isUser() && data.config.owner === data.principal().username">
```
desislava-hristova-ontotext, line 152: Actually here data is undefined, so both data.config.owner === data.principal().username === undefined, so this is always true and meaningless, evaluating {{config.owner}} you can see that it has a value (i.e the admin if no user). The actual cause is that principal() is undefined here and needs to be replaced with principal only. See https://github.com/Ontotext-AD/graphdb-workbench/commit/527e8d62eea383b99aa4c7258b9eb57291f1865d. 
Please, also search for all places where principal() is called in html and replace with principal. So in short here it should be: && config.owner === principal.username

desislava-hristova-ontotext (changes requested):

desislava-hristova-ontotext (approved):

PR #717, opened 2022-08-02
Visual Graph link limit

nikolatulechki:
I changed the default visual graph link limit to 200. The current one  (20) is way too low and prevents us from being able to demonstrate the visualization without having to awkwardly explain to the other party how to increase the limit


hudson-onto:

PR #723, opened 2022-09-04
GDB-7292: Easy guide building blocks - various improvements

avataar:
- Split visual-graph into 3 blocks + some changes in visual graph code to allow better integration with easy guide
- Added generic-message block
- Added guide-ended block
- Refactored select-repository into two independent blocks that select the repo in different ways
- Added validation on some blocks, e.g. when user has to type something or upload a file, and a general mechanism for validation via 'onNextValidate'
- Reorganized how translations are handled: title/content can now provide all the translations for custom messages
- Improved placement for some blocks
- Added a complete example of a guide (the_beatles) showing all blocks in logical sequence


src/guides/guide-beatles/beatles_guide.json
```diff
@@ -0,0 +1,69 @@
...
+            "options": {
+                "clickOnIRI": "http://stardog.com/tutorial/Imagine",
+                "shortIRI": "Imagine"
+            }
+        },
+        {
+            "guideBlockName": "generic-message",
+            "options": {
+                "title": {
+                    "en": "Visual graph: more",
```
boyan-tonchev, line 56: If we make translations in this way, we have to update all guides, when we add a new language. 
In my opinion better is if we use label key and a file per language, because when we add a new language we have to add only one file.

src/guides/guide-beatles/beatles_guide.json
```diff
@@ -0,0 +1,69 @@
...
+        },
+        {
+            "guideBlockName": "visual-graph-properties",
+            "options": {
+                "clickOnIRI": "http://stardog.com/tutorial/Imagine",
+                "shortIRI": "Imagine"
+            }
+        },
+        {
+            "guideBlockName": "generic-message",
```
boyan-tonchev, line 53: You can just use "info-message" instead create new complex step.

src/js/angular/guides/steps/core/plugin.js
```diff
@@ -79,5 +80,17 @@ PluginRegistry.add('guide.step', [
...
         }
+    },
+    {
+        'guideBlockName': 'guide-end',
+        'getStep': (options, GuideUtils) => {
+            const notOverridable = {
+                'type': 'readonly'
+            };
+            const step = angular.extend({}, BASIC_STEP, options, notOverridable);
+            step.title = 'End of guide';
```
boyan-tonchev, line 91: Have to be translated instead hard coded.

src/js/angular/repositories/controllers.js
```diff
@@ -362,6 +362,11 @@ function LocationsAndRepositoriesCtrl($scope, $modal, toastr, $repositories, Mod
 
     getLocations();
     const timer = $interval(function () {
+        if (ShepherdService.isActive() && !$rootScope.guidePaused) {
```
boyan-tonchev, line 365: Add isActive function to GuideService and use it here instead to use it directly from ShepherdService. Idea behaind ShepherdService is to wrapp shepard libraries. 
If we decide to change tour libraries we will implement a new wrapper service and change it only in GuideService.

src/js/angular/controllers.js
```diff
@@ -157,10 +157,12 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repos
 
     $rootScope.$on('guideReset', function () {
         $scope.guidePaused = false;
+        $rootScope.guidePaused = false;
```
boyan-tonchev, line 160: Add `$rootScope.guidePaused = false;` in function resumeGuide above.

src/guides/guide-beatles/beatles_guide.json
avataar: Removed "generic-message" and replaced usage with "info-message"

src/js/angular/repositories/controllers.js
avataar: Fixed.

src/js/angular/controllers.js
avataar: Fixed.

PR #727, opened 2022-09-09
GDB-7292: Easy guide building blocks - single JSON for all guides + use of /rest/guides

avataar:
- Uses a single JSON file that lists all guides
- Generates an ID for each guide on load
- Fetches guides from /rest/guides
- Downloadable resources are downloaded via /rest/guides/download - which should always provide a content-disposition header so I've removed the "download" attribute from the link. This helps with spotting errors (the error will be shown in browser instead of downloading the file)


src/js/angular/guides/controllers.js
```diff
@@ -25,16 +25,27 @@ function GuidesCtrl($scope, $rootScope, GuidesService, $filter, $translate) {
...
+                $scope.guides.forEach(guide => guide.translatedGuideName = GuideUtils.translateLocalMessage($translate, guide.guideName));
                 $scope.matchedElements = $scope.guides;
                 $scope.changePagination();
             });
-    }
 
-    $scope.startGuide = function (guideFileName) {
-        GuidesService.startGuide(guideFileName);
-    }
+        $rootScope.$on('guideResume', () => {
```
boyan-tonchev, line 34: The controller of “Guides” view will be initialized when the view is opened only then it will be start to listening for the event. This will not work if an user has guide on pause and refresh the page or close browser and open it again on other view and click on button “Guide paused”.

avataar: I've restored the old mechanism in the service, with the drawback that the guides will be refetched on resume.

src/js/angular/guides/guides.service.js
```diff
@@ -263,8 +222,11 @@ function GuidesService($http, $rootScope, $translate, ShepherdService, $reposito
      */
     this.getGuides = () => {
         return new Promise(resolve => {
-            $http.get(`${this.guideRepositoryUrl}/guides/guides.json`)
+            $http.get(GuideUtils.GUIDES_LIST_URL)
                 .success(function (data) {
+                    angular.forEach(data, (guide, index) => {
+                        guide.guideId = index;
```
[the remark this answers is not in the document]
avataar: I don't believe any user will put a guide on pause and then change the guides. However, I added a check for guideId - if present we won't generate one.

boyan-tonchev (approved):

PR #740, opened 2022-09-21
GDB-7400: Replace usages of "répertoire" with "dépôt" and "domaine-gamme" with "domaine-image" in the French translation

avataar:
- also corrected some capitalisation


nikolatulechki (approved):
👍

PR #743, opened 2022-09-23
GDB-7404: Cannot release workbench after new dependencies were added

avataar:
- Updated the nodejs version used by Jenkins for both build and release to latest 18.9.0
- Added .npmrc file with legacy-peer-deps=true - this seems to be needed with newer nodejs/npm
- Added @babel/runtime as a dev dependency - seems to be needed now
- Fixed issue with ng-options/ng-repeat on empty elements like \<tag/\> by converting to \<tag\>\</tag\> (unknown why this caused an issue but IntelliJ marked those as a warning anyway)
- Removed some redundant closing html tags
- Upgraded cypress-localstorage-commands to newer version
- Added NODE_OPTIONS = "--openssl-legacy-provider" to Jenkins pipelines - needed at least on the Linux env we use with Jenkins but not needed on Mac with the same nodejs version


svilenvelikov:
@avataar Did you by any chance forget to commit the package.json from the test-cypress package? I see the package-lock.json in the files list but not the package.json.

avataar:
> 

There were no changes so nothing to commit. Cypress's package-lock.json changed mostly because the newer nodejs uses a different format.

avataar:
> 

When running Cypress it was complaining that it needs versions of nodejs from 10 to 16. It was upgraded in package.json of the main directory (since it was defined there originally).

svilenvelikov (approved):

PR #755, opened 2022-10-06
GDB-7449_GDB-7451

boyan-tonchev:
GDB-7449: There was while cycle which never end. Fixed.
GDB-7451: Remove automatically enabling of autocomplete when next button is clicked. Added canBePaused false flag to step which shows status of the autocomplete.


boyan-tonchev:
Jenkins retry build

avataar (approved):

avataar (approved):

PR #761, opened 2022-10-10
fix(GDB-7265): Disable FTS checkbox tests

teodossidossev:
## What
Reported failing test for enabled FTS by default.

## Why
The FTS configuration is changed, FTS is not enabled by default. As the config state is unclear by the moment, we need to stop the FTS checkbox state.

## How
Commented FTS checkbox interactions and state checks. Added TODOs with description


svilenvelikov:
If I understand what @avataar said during the daily we had to disable all verifications for options related with the FTS and not only this particular one.

svilenvelikov (approved):

teodossidossev:
> If I understand what @avataar said during the daily we had to disable all verifications for options related with the FTS and not only this particular one.

IMO I've covered them all.

PR #764, opened 2022-10-11
GDB-7482: User Guide pausing during creation of repo removes the repo…

boyan-tonchev:
… name

Added show function to steps after filling of repository id which populate it with id configured in guide description.


boyan-tonchev:
Jenkins retry build

avataar (approved):

PR #766, opened 2022-10-12
GDB-7500: User Guide crashes if autocomplete is turned off after stat…

boyan-tonchev:
Disabled processing of mouse event on autocomplete checkbox. Added implementation to the autocomplete plugin which block guide if autocomplete is not enabled.


boyan-tonchev:
Jenkins retry build

avataar (approved):

PR #770, opened 2022-10-14
GDB-7467: Add more FTS configurations

Borislav-Bonev-Ontotext:
Updated repository page to support the new configurations and also renamed the old properties to match the new one.


src/pages/repository.html
```diff
@@ -296,21 +296,51 @@ <h5>{{'repo.indexing.section' | translate}}</h5>
...
+                            <label
+                                class="col-xs-12 col-md-3 col-lg-2 col-form-label">{{'menu.ftsIrisIndex.label' | translate}}</label>
+                            <div class="col-xs-3 col-md-2" tooltip="{{'repoTooltips.ftsIrisIndex' | translate}}">
+                                <input class="form-control" id="ftsIrisIndex" type="text"
+                                       name="ftsIrisIndex"
+                                       ng-model="repositoryInfo.params.ftsIrisIndex.value"
+                                       ng-disabled="repositoryInfo.params.enableFtsIndex.value === 'false'"/>
+                            </div>
+                        </div>
+                        <div class="form-group row indented-div" style="display: none">
```
Borislav-Bonev-Ontotext, line 336: This configuration will be in the next version, Should I remove it or I can leave it there until the next version

yordanalexandrov: I guess you can leave it, but I would suggest adding an `ng-if="false"` (so it does not show in the DOM at all) and a `TODO` inside, instead of the `display: none` style

yordanalexandrov:
It would be nice to extend the tests

yordanalexandrov (approved):

PR #775, opened 2022-10-20
GDB-6076 Changed the default limit of raw response from 100 to 1000

sava-savov-ontotext:


src/js/lib/yasr.bundled.js
```diff
@@ -56918,7 +56926,7 @@ var root = module.exports = function(yasr) {
 		var contentType = yasr.results.getOriginalContentType();
 		var type = yasr.results.getType();
 		return {
-			getContent: function() {return yasr.results.getOriginalResponseAsString();},
+			getContent: function() {return yasr.results.getOriginalResponse();},
```
svilenvelikov, line 56929: @sava-savov-ontotext Where this change came from? Any idea?

sava-savov-ontotext: Probably this is a kind of a fix that @MartinaDimova has done but which was not included into YASR project but fixed straight into WB project

svilenvelikov: Well, this is not a good practice at all

PR #780, opened 2022-10-21
GDB-7527-Missing label in Visual graph

MartinaDimova:
Fixed sidepanel.show.preferred.predicates to sidepanel.show.preferred.predicates.only in graphs-visualizations.html so it can be properly translated.


desislava-hristova-ontotext (approved):

desislava-hristova-ontotext:
Jenkins retry build

PR #783, opened 2022-10-21
GDB-7501: User Guide can't be rerun if more than one repo exists

boyan-tonchev:
Fixed. 
 - GDB-7538
Additional work:
 - There was a bug when a guide is on step to choose a repository and user click on any menu item of any repository, the dropdown wouldn't close. The problem was that we stop dropdown to be closed when a guide is active, this is because of other bug "GDB-7450: User Guide crashes on repo select step". Fixed, when the step is shown we add a class "autoCloseOff" to dropdown. This class will block closing of dropdown when user click outside dropdown. This class will be removed when user click on "Next" button or some menu item.

### DIFF
--- a/src/js/angular/guides/steps/complex/select-repository/plugin.js
+++ b/src/js/angular/guides/steps/complex/select-repository/plugin.js
@@ -1,11 +1,20 @@
+const getRepositoryName = (services, options) => {
+    return services.$repositories.getRepositories().find((repo) => repo.id === options.repositoryId) ? options.repositoryId : options.repositoryIdBase;
+};
+
+const getRepositoryElementSelector = (services, options) => {
+    return services.GuideUtils.getGuideElementSelector(`repository-${getRepositoryName(services, options)}-button`);
+};
+
 PluginRegistry.add('guide.step', [
     {
         guideBlockName: 'select-repository-dropdown',
         getSteps: (options, services) => {
             const GuideUtils = services.GuideUtils;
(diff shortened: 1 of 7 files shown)


src/js/angular/guides/steps/complex/select-repository/plugin.js
```diff
@@ -17,17 +26,40 @@ PluginRegistry.add('guide.step', [
             }, {
                 guideBlockName: 'clickable-element',
                 options: angular.extend({}, {
-                    content: 'guide.step_plugin.select-repository.content',
-                    elementSelector: repositoryButtonSelector,
+                    content: () => {
```
avataar, line 29: Why not make restore the general way content/extraContent is handled (both here and in shepherd.service.js) and change the translation template to e.g. "Click on the \<b\>{{**getRepositoryId()**}}\</b\> repository button."

Then you simply need to define the getRepositoryId() function within the options at some point.

boyan-tonchev: I will try.

boyan-tonchev: I tried, but it did not work as we expected. When content property of step description is string it is calculated only once when steps of Shepherdjs library are initialized. At those time, we do not know which name to use. If content is function it is called every time when step is shown.

avataar (approved):

PR #787, opened 2022-10-31
GDB-6908 Added method compactAuthorities in the security controller and tests verifying it.

sava-savov-ontotext:


src/js/angular/security/controllers.js
```diff
@@ -104,6 +104,37 @@ const parseAuthorities = function (authorities) {
     };
 };
 
+const compactAuthorities = function (authorities) {
+    const writeAllRepos = authorities.find(authority => authority === `${WRITE_REPO_PREFIX}*`);
```
svilenvelikov, line 108: `writeAllRepos` is not a good name, better name would be something like `writeAllRole` or `writeAllAuthority`.

src/js/angular/security/controllers.js
```diff
@@ -104,6 +104,37 @@ const parseAuthorities = function (authorities) {
...
+    if (writeAllRepos) {
+        return  _.remove(authorities, function (_authority) {
+            if (_authority.indexOf(READ_REPO_PREFIX) === -1 && _authority.indexOf(WRITE_REPO_PREFIX) === -1) {
```
svilenvelikov, line 111: Can't we have constants for these WRITE_REPO_PREFIX_* and READ_REPO_PREFIX_*? As far as I can see, these are just static strings and can be defined once and reused. 
E.g. WRITE_ALL_REPO_WILDCARD or something.

src/js/angular/security/controllers.js
```diff
@@ -104,6 +104,37 @@ const parseAuthorities = function (authorities) {
     };
 };
 
+const compactAuthorities = function (authorities) {
```
svilenvelikov, line 107: It's hard to understand what this function does without reading the code and why. Can you drop some doc comment here. Other than that, I wonder, why this is not placed inside some service?

src/js/angular/security/controllers.js
```diff
@@ -104,6 +104,37 @@ const parseAuthorities = function (authorities) {
...
+    const writeAllRepos = authorities.find(authority => authority === `${WRITE_REPO_PREFIX}*`);
+    if (writeAllRepos) {
+        return  _.remove(authorities, function (_authority) {
```
svilenvelikov, line 110: You know that the function `_.remove` from lodash actually mutates the array argument, aren't you? Is it intentional that you just use the return value of the function and not aware of the actual authorities array? Isn't it better to use Array.reduce or Array.filter for example instead?

test/security/controllers.spec.js
```diff
@@ -426,6 +426,28 @@ describe('==> Controllers tests', function () {
...
+                $httpBackend.flush();
+                $timeout.flush();
+            });
+            it('should compact the user granted authority \'READ_REPO_*\' properly', function () {
+                $httpBackend.flush();
+                $scope.user.grantedAuthorities = ['ROLE_USER', 'WRITE_REPO_myrepo', 'READ_REPO_myrepo', 'READ_REPO_*'];
+                $scope.createUserHttp();
+                $httpBackend.expectPOST("rest/security/users/testov", {"password":"testova",
+                    "appSettings":{"DEFAULT_SAMEAS":true,"DEFAULT_INFERENCE":true,"EXECUTE_COUNT":true, "IGNORE_SHARED_QUERIES":false,"DEFAULT_VIS_GRAPH_SCHEMA":true},
+                    "grantedAuthorities":["ROLE_USER","WRITE_REPO_myrepo","READ_REPO_myrepo","READ_REPO_*"]})
```
svilenvelikov, line 446: Maybe I don't get the requirement correctly, but shouldn't you expect 
`"grantedAuthorities":["ROLE_USER","WRITE_REPO_myrepo","READ_REPO_*"]`

svilenvelikov (changes requested):

test/security/controllers.spec.js
sava-savov-ontotext: If there is "WRITE_REPO_myrepo", "READ_REPO_myrepo" is expected as well.

src/js/angular/security/controllers.js
sava-savov-ontotext: The idea is to modify the array, leaving only "WRITE_REPO_*", "READ_REPO_*" and all other authorities, that aren't related with read or write ones.

PR #789, opened 2022-10-31
GDB-6595 Added the same behaviour from SPARQL and Visual graph views for inference and sameAs

sava-savov-ontotext:


boyan-tonchev (approved):

src/js/angular/security/controllers.js
```diff
@@ -455,6 +455,16 @@ securityCtrl.controller('CommonUserCtrl', ['$rootScope', '$scope', '$http', 'toa
                 $scope.confirmPasswordError = '';
             }
         };
+
+        $scope.shouldDisableSameAs = function () {
+            const sameAsCheckbox = $('#sameAsCheck');
+            if ($scope.user && !$scope.user.appSettings.DEFAULT_INFERENCE && sameAsCheckbox.prop('checked')) {
+                sameAsCheckbox.prop("checked", false);
+                $scope.user.appSettings.DEFAULT_SAMEAS = false;
```
desislava-hristova-ontotext, line 463: Why are you modifying the default user settings, this does not look okay to me?

sava-savov-ontotext: In this way we force user to properly configure sameAs and include inferred. 
The same behaviour could be found on Visual and SPARQL views and was discussed with Pavel that we want the WB to behave consistent among different views.

desislava-hristova-ontotext: Ah, okay, that is exactly what my thoughts were. I didn't see that you make the change in the security page controller, I thought that you are adding such logic to the sparql view. My mistake. This is exactly how I thought it should work.

desislava-hristova-ontotext (approved):

PR #794, opened 2022-11-04
GDB-6815: Resource view on mobile

boyan-tonchev:
What?
 There was wrong calculation of table size when page is loaded on mobile.

Why?
 "colResizable-1.4.js" library is used for column size calculation. It don't do its job properly on mobile.

How?

 - Added functionality to switch off column resizing when screen width is small. Affected pages are:
  - explore
  - create jdbc
  - similarity, create similarity
  - SPARQL query
  - SPARQL template create

### DIFF
--- a/src/js/angular/core/directives/queryeditor/query-editor.directive.js
+++ b/src/js/angular/core/directives/queryeditor/query-editor.directive.js
@@ -4,6 +4,7 @@ import 'angular/externalsync/controllers';
 import YASQE from 'lib/yasqe.bundled';
 import YASR from 'lib/yasr.bundled';
 import {decodeHTML} from "../../../../../app";
+import {YasrUtils} from "../../../utils/yasr-utils";
 
 angular
     .module('graphdb.framework.core.directives.queryeditor.queryeditor', [
@@ -48,6 +49,10 @@ function queryEditorDirective($timeout, $location, toastr, $repositories, Sparql
         scope.nostatus = attrs.hasOwnProperty('nostatus');
         // Doesn't show the run button
         scope.norun = attrs.hasOwnProperty('norun');
+
+        scope.enableColumnResizingOnWindowWidth = attrs.hasOwnProperty('enableColumnResizingOnWindowWidth');
+
+
         // Name of the Run button in the editor
         scope.runButtonName = 'query.editor.run.btn';
         if (attrs.runButtonName) {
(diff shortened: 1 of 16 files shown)


boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

src/js/angular/core/directives/queryeditor/query-editor.directive.js
```diff
@@ -48,6 +48,10 @@ function queryEditorDirective($timeout, $location, toastr, $repositories, Sparql
         scope.nostatus = attrs.hasOwnProperty('nostatus');
         // Doesn't show the run button
         scope.norun = attrs.hasOwnProperty('norun');
+
+        scope.columnResizeWidth = scope.$eval(attrs.columnResizeWidth);
```
svilenvelikov, line 52: What is the meaning of this `columnResizeWidth`? As I can see from your changes in the code in the other project, this is the minimum window width on which the collumns can be resizable in the table renderer in YASR. Right? IMO the name of this variable is quite misleading as well as the parameter passed to the YASR. Also see the comment in the other MR as well.

src/js/angular/explore/controllers.js
```diff
@@ -96,12 +96,22 @@ function ExploreCtrl($scope, $http, $location, toastr, $routeParams, $repositori
...
                         yasr = YASR(document.getElementById('yasr'), { // eslint-disable-line new-cap
                             //this way, the URLs in the results are prettified using the defined prefixes
                             getUsedPrefixes: $scope.usedPrefixes,
                             persistency: false,
                             hideHeader: true,
-                            locale: $languageService.getLanguage()
+                            locale: $languageService.getLanguage(),
+                            pluginsOptions: {
+                                table: {
+                                    columnResizeWidth: 1000
```
svilenvelikov, line 112: Can we have this as a constant somewhere?

src/js/angular/similarity/controllers/similarity-list.controller.js
```diff
@@ -107,11 +107,21 @@ function SimilarityCtrl($scope, $interval, toastr, $repositories, $licenseServic
...
+                            yasr.destroy();
+                        }
+                    });
                     yasr = YASR(document.getElementById('yasr'), { // eslint-disable-line new-cap
                         //this way, the URLs in the results are prettified using the defined prefixes
                         getUsedPrefixes: $scope.usedPrefixes,
                         persistency: false,
-                        hideHeader: true
+                        hideHeader: true,
+                        pluginsOptions: {
```
svilenvelikov, line 120: I'd encapsulated this in some factory function in an utility or a service because it's repeated a couple of times in different components which is bad.

src/pages/jdbc-create.html
```diff
@@ -129,7 +129,7 @@ <h1>
                 <br/>
                 <query-editor nostorage notabs notoolbar-saved notoolbar-copy nostatus nocount norun
                               notoolbar-inference
-                              notoolbar-same-as callback-on-change="updateDirty">
+                              notoolbar-same-as callback-on-change="updateDirty" column-resize-width="1000">
```
svilenvelikov, line 132: Is there a reason for having this parameter at all? Why don't you just set it inside the directive's controller by reading it from a constant or environment variable.

boyan-tonchev: Yes, there is. In some place we don't need to turn off column resizing, because everything look like good.
 For the value I thought that is better to be configurable because in some place the width can be different.
 But may be you are right that we do not need this. I will remove the value and add it in constant.

src/js/angular/core/directives/queryeditor/query-editor.directive.js
boyan-tonchev: I will rename it.

boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

PR #795, opened 2022-11-04
GDB-6815_Resource_view_on_mobile

boyan-tonchev:
GDB-6815: Resource view on mobile

What?
 There was wrong calculation of table size when page is loaded on mobile.

Why?
 "colResizable-1.4.js" library is used for column size calculation. It don't do its job properly on mobile.

How?

 - Added functionality to switch off column resizing when screen width is small. Affected pages are:
  - explore
  - create jdbc
  - similarity, create similarity
  - SPARQL query
  - SPARQL template create


boyan-tonchev:
Jenkins retry build

PR #797, opened 2022-11-08
GDB-7585: User Guide crashes when a double click happens on import segment

boyan-tonchev:
What?

 User guide crashes when a double-clicked on import button.

Why?

The problem was not because the user is double-clicked the button. The modal dialog was opened with options which close it if the user click outside the dialog. So first click on the button opens "Import" dialog and second one close it.

How?

When a guide is active the dialog is opened with options to not be closed when user click outside it.


desislava-hristova-ontotext (approved):

boyan-tonchev:
Jenkins retry build

PR #804, opened 2022-11-09
Removed leading / so it would pick the correct context path

Sukhumi:
When GraphDB is behind a reverse proxy and there is a context path the Interactive guides page will fail to load. 

Example: 
If GraphDB is lunched in kubernetes with our helm chart the address you access the interactive guide page is:
http://graphdb.local/graphdb/guides
but it will look for its content at http://graphdb.local/rest/guides


desislava-hristova-ontotext:
Jenkins retry build

boyan-tonchev (approved):

Sukhumi:
Jenkins retry build

PR #817, opened 2022-11-22
GDB-7434 Class relationship and hierarchy crashes the web browser when many contexts exist

MartinaDimova:
Inserted two ng-ifs to ensure there is no empty reloading of the page when there is no class hierarchy or class relations in the relevant html pages.


src/js/angular/graphexplore/controllers/rdf-class-hierarchy.controller.js
```diff
@@ -488,6 +488,14 @@ function RdfClassHierarchyCtlr($scope, $rootScope, $location, $repositories, $li
     };
 
     $scope.isAllGraphsSelected = function () {
-        return $scope.getSelGraphValue() === 'all.graphs.label'
-    }
+        return $scope.getSelGraphValue() === 'all.graphs.label';
+    };
+
+    //Method used to limit the graphs loaded in the drop-down menu, otherwise browsers crash.
+    $scope.downsizeGraphsInRepoIfMoreThan1000 = function () {
```
sava-savov-ontotext, line 495: You don't need this function in both rdf-class-hierarchy and dependencies controllers. You could add this check and slice when you're assigning the returned value from RDF4JRepositoriesRestService.resolveGraphs() in initView function of the controllers.

MartinaDimova:
Jenkins retry build

sava-savov-ontotext (approved):

PR #819, opened 2022-12-01
GDB-6647 "copy link" icon cannot be clicked

MartinaDimova:
* Changed copy link to be able to copy the uri directly without opening the modal.

* Made copy link to be available when mouse is in the current cell and visible if the column is narrower than the uri.

### DIFF
--- a/src/css/yasr.custom.css
+++ b/src/css/yasr.custom.css
@@ -131,6 +131,10 @@
     padding: 0px;
 }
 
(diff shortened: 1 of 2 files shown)


MartinaDimova:
The MR is draft as a decision has to be made whether to leave the copy function of each uri in the Data Table Resources, to be "direct" - meaning without opening a confirming modal.

sava-savov-ontotext (approved):

PR #824, opened 2022-12-15
GDB-7591 The Similarity indexes "Show SPARQL query" gives a broken SPARQL

MartinaDimova:
* Inserted check for use of a missing prefix in View mode /View Index Query/.

* Replacing custom prefix in the template for http://www.ontotext.com/graphdb/similarity/instance/ in View mode /View SPARQL Query/.

### DIFF
--- a/src/js/angular/similarity/controllers/similarity-list.controller.js
+++ b/src/js/angular/similarity/controllers/similarity-list.controller.js
@@ -256,7 +256,7 @@ function SimilarityCtrl($scope, $interval, toastr, $repositories, $licenseServic
         let tokens = [];
         let prefix = '';
         tokens = queryTemplate.match(/[a-zA-Z0-9-]+:<http:\/\/www.ontotext.com\/graphdb\/similarity\/instance\/>/);
-        prefix = tokens[0].substring(0, tokens[0].indexOf(':'));
+        prefix = tokens == null ? "similarity-index" : tokens[0].substring(0, tokens[0].indexOf(':'));
 
         const replacedQuery = queryTemplate
             .replace('?index', prefix + ':' + $scope.selected.name)
--- a/src/js/angular/similarity/controllers/similarity-list.controller.js
+++ b/src/js/angular/similarity/controllers/similarity-list.controller.js
@@ -256,7 +256,7 @@ function SimilarityCtrl($scope, $interval, toastr, $repositories, $licenseServic
         let tokens = [];
         let prefix = '';
         tokens = queryTemplate.match(/[a-zA-Z0-9-]+:<http:\/\/www.ontotext.com\/graphdb\/similarity\/instance\/>/);
-        prefix = tokens[0].substring(0, prefix.indexOf(':'));
+        prefix = tokens[0].substring(0, tokens[0].indexOf(':'));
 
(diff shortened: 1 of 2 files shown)


sava-savov-ontotext (dismissed):

src/js/angular/similarity/controllers/similarity-list.controller.js
```diff
@@ -251,8 +251,20 @@ function SimilarityCtrl($scope, $interval, toastr, $repositories, $licenseServic
         } else {
             queryTemplate = ($scope.selected.searchQuery) ? $scope.selected.searchQuery : $scope.searchQueries[$scope.selected.type];
         }
+
+        //replace template prefix for PREFIX_INSTANCE in the view mode with actual prefix from the query
+        var array = [];
```
svilenvelikov, line 256: Please, use `let` instead of `var`

svilenvelikov, line 256: Better rename this to `tokens` for example or something more specific.

MartinaDimova:
Jenkins retry build

src/js/angular/similarity/controllers/similarity-list.controller.js
```diff
@@ -253,15 +253,10 @@ function SimilarityCtrl($scope, $interval, toastr, $repositories, $licenseServic
...
-        array = queryTemplate.split(' ');
-        for (let i = 0; i < array.length; i++) {
-            if (array[i].includes(PREFIX_INSTANCE)) {
-                prefix = array[i].substring(0, array[i].indexOf(':'));
-                break;
-            }
-        }
+        let tokens = [];
+        let prefix = '';
+        tokens = queryTemplate.match(/[a-zA-Z0-9-]+:<http:\/\/www.ontotext.com\/graphdb\/similarity\/instance\/>/);
```
svilenvelikov, line 258: Are you sure this now actually work? In any case, you should at least verify that the match actually found something, otherwise you will get the equivalent of the NPE in javascript.

src/js/angular/similarity/controllers/similarity-list.controller.js
```diff
@@ -253,15 +253,10 @@ function SimilarityCtrl($scope, $interval, toastr, $repositories, $licenseServic
...
+        prefix = tokens[0].substring(0, prefix.indexOf(':'));
```
svilenvelikov, line 259: This line won;t do what you might expect. The `prefix.indexOf(':')` will always return `-1` I believe, because `prefix` is not initialized with any actual value at this point yet. I guess you wanted to use `tokens[0].indexOf(':')` maybe.

src/js/angular/similarity/controllers/similarity-list.controller.js
svilenvelikov: OK, then just add a check if the `tokens` is defined before trying to access it below to prevent errors.

MartinaDimova: OK. Thanks.

svilenvelikov (approved):

MartinaDimova:
Jenkins retry build

MartinaDimova:
Jenkins retry build

PR #825, opened 2022-12-20
GDB-7747 Update YASR bundle

yordanalexandrov:
## What?
Update the YASR-Ontotext bundle file so it is the latest

## Why?
There is a new feature in YASR-Ontotext. If translateHeaders option is set to true when creating the YASR instance, it will try to translate the headers in the table view

## How?
Replace YASR-Ontotext bundle file with the latest build


src/js/lib/yasr.bundled.js
```diff
@@ -54121,7 +54121,7 @@ module.exports={
         "spec": ">=1.4.1 <2.0.0",
         "type": "range"
       },
-      "/home/boyan/Git/YASR-Ontotext"
+      "/home/yordan/dev/ontotext/YASR-Ontotext"
```
svilenvelikov, line 54124: what the hell is this?

yordanalexandrov: well ¯\_(ツ)_/¯

svilenvelikov (approved):

PR #838, opened 2023-01-18
GDB-7784 integrate saved query edit action

svilenvelikov:
## What
Integrate the saved queries edit action.

## Why
Edit saved query action exists in current implementation in the GDB WB. It allows the user to edit an existing saved query and to save it again.

## How
* Introduced additional event handler for the edit action as well as a new saved queries related configuration property which is required by the ontotext yasgui component.
* Improved the controller and the sparql rest service.
* Implemented tests.


test-cypress/cypress.json
```diff
@@ -8,6 +8,6 @@
     "supportFile": "support/index.js",
     "videosFolder": "report/videos",
     "video": false,
-    "defaultCommandTimeout": 30000,
+    "defaultCommandTimeout": 10000,
```
svilenvelikov, line 11: Not sure why this was set to 30seconds but I think it's far too big timeout for the commands. Will leave this way for now and see if it'll cause some issues with the builds.

boyan-tonchev (approved):

PR #842, opened 2023-01-23
GDB-7785 integrate share saved query action

svilenvelikov:
## What
Integrate the share saved query action.

## Why
Share saved query action exists in current implementation in the GDB WB. It allows the user to get a shareable link for an existing saved query.

## How
* Introduced additional event handler for the share saved query action.
* Implemented tests and reorganized the old tests a bit to be better structured and readable as well as to reduce some duplication.


boyan-tonchev (approved):

svilenvelikov:
Jenkins retry

svilenvelikov:
Jenkins retry build

PR #845, opened 2023-01-26
GDB-7845 Introduce ability to open shared query

svilenvelikov:
## What
Allow the client to open a query model in a new or existing tab depending on if there is already existing tab with the same name and query.

## Why
The client would needs to open a shared query in editor tab, be it a new or existing one.

## How
* Analyze the url parameters and if there is a savedQueryName parameter then try to load the saved query and open it in the editor.
* Implemented tests.


svilenvelikov:
Jenkins retry build

boyan-tonchev (approved):

PR #846, opened 2023-01-27
GDB-7737 Refactor and implement new resource monitor view

yordanalexandrov:
## What
Refactor the resource monitor view and add new charts for monitoring of system resources

## Why
Refactoring would allow for easier and faster implementation of more charts. The view should help users to better evaluate the system when trying to solve problems

## How
- created a paren ChartData class with methods for visualisation of charts
- implemented ChartData classes for the concrete charts with their specific methods
- refactored the view to show charts in 2 column layout when screen is wider and 1 column for smaller screens
- updated nvd3.js and angular-nvd3.js libraries to the latest available


src/js/angular/resources/chart-models/chart-data.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+    addData(timestamp, data) {
+        this.removeOldData();
+        this.addNewData(timestamp, data);
+        this.updateRange();
+    }
+
+    removeOldData() {
+        if (this.data[0].values.length > this.range) {
+            this.data.forEach((data) => data.values.shift());
```
svilenvelikov, line 21: array.shift is expensive function would you consider something else?

src/js/angular/resources/chart-models/heap-memory-chart.js
```diff
@@ -0,0 +1,50 @@
...
+            key: this.translateService.instant('resource.memory.committed'),
+            values: []
+        }, {
+            key: this.translateService.instant('resource.memory.used'),
+            area: 'true',
+            values: []
+        }];
+    }
+    addNewData(timestamp, data) {
+        const [committed, used] = this.data;
```
svilenvelikov, line 21: Why `this.data` and not just `data`?

src/js/angular/resources/chart-models/chart-data.js
```diff
@@ -0,0 +1,26 @@
...
+        this.data = this.createDataHolder();
+    }
+    chartSetup() {}
+    createDataHolder() {
+        throw new Error('Must implement data holder creation');
+    }
+    addData(timestamp, data) {
+        this.removeOldData();
+        this.addNewData(timestamp, data);
+        this.updateRange();
```
svilenvelikov, line 16: Maybe pass down all necessary stuff to these template methods in order to make the code less confusing. Right now it's not clear, why in some methods you get things from the class scope and others you get as parameters.

svilenvelikov (changes requested):

src/js/angular/resources/chart-models/heap-memory-chart.js
yordanalexandrov: I should probably change the name of `this.data` to `dataHolder` or something so it is not confused with the data from the response.

`this.data` is where I store the properties. `data` is the data from the response.

src/js/angular/resources/chart-models/chart-data.js
yordanalexandrov: I am not sure which will be better in this case.

src/js/angular/resources/controllers.js
```diff
@@ -158,25 +55,52 @@ resourcesCtrl.controller('ResourcesCtrl', ['$scope', 'toastr', '$interval', '$ti
...
+        $scope.resourceMonitorData.cpuLoad = new CPULoadChartData($translate, angular.copy($scope.chartOptions));
+        $scope.resourceMonitorData.fileDescriptors = new FileDescriptorsChart($translate, angular.copy($scope.chartOptions));
+        $scope.resourceMonitorData.heapMemory = new HeapMemoryChart($translate, angular.copy($scope.chartOptions));
+        $scope.resourceMonitorData.offHeapMemory = new NonHeapMemoryChart($translate, angular.copy($scope.chartOptions));
+        $scope.resourceMonitorData.diskStorage = new DiskStorageChart($translate, angular.copy($scope.chartOptions));
+
+        $scope.firstLoad = true;
+        $scope.activeTab = 'resourceMonitor';
+        $scope.error = '';
+        $scope.loader = true;
```
svilenvelikov, line 68: See which of all these variables are really needed to be in the scope, a.k.a. to be accessible through the template and make those which are not needed just some local `private` functions.

svilenvelikov (changes requested):

src/js/angular/resources/chart-models/chart-data.js
yordanalexandrov: Shift seems to be better in this case. I will leave it.

src/js/angular/resources/chart-models/heap-memory-chart.js
yordanalexandrov: I changed `this.data` to `this.dataHolder`

src/js/angular/resources/chart-models/chart-data.js
yordanalexandrov: Refactored it

src/js/angular/resources/chart-models/cpu-load-chart.js
```diff
@@ -0,0 +1,29 @@
+import {ChartData} from "./chart-data";
+
+export class CPULoadChartData extends ChartData {
```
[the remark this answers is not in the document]
yordanalexandrov: It was the first. Changed it.

src/js/angular/resources/controllers.js
```diff
@@ -158,25 +55,52 @@ resourcesCtrl.controller('ResourcesCtrl', ['$scope', 'toastr', '$interval', '$ti
...
+        $scope.getResourceMonitorData = function () {
+            if ($scope.error) {
+                return;
+            }
+            MonitoringRestService.monitorResources().success(function (data) {
```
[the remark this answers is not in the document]
yordanalexandrov: Done

src/js/angular/resources/controllers.js
yordanalexandrov: Done

src/js/angular/resources/controllers.js
```diff
@@ -90,14 +94,17 @@ resourcesCtrl.controller('ResourcesCtrl', ['$scope', 'toastr', '$interval', '$ti
...
-                $scope.error = getError(data);
+            }).catch(function (error) {
+                $scope.error = getError(error);
                 $scope.loader = false;
+            }).finally(() => {
+                pendingRequest = false;
             });
         };
-
         const timer = $interval(function () {
```
svilenvelikov, line 104: Few more thoughts here.
* Create a configuration or at least a constant for this 2000ms interval.
* What if given request takes say 3sec? Then you'll skip current iteration and in fact there will be a 4sec interval in the chart somewhere. Right? Is that OK? I guess that'd be an exceptional case probably but still.
* Why instead of having this `pendingRequest` flag you don't just start a simpler `setTimeout(getResourceMonitorData, 2000)` in the success handler.
* What's the requirement in case an error get's returned? In current implementation you'll just continue to poll the server and get errors. Is that OK? Can we consider some approach for handling this scenario? Say, if there are 5 requests without response and you stop polling and show some info in the widgets. Not sure if this wouldn't mean that   GDB is down for some reason.

yordanalexandrov: Extracted constant, removed `pendingRequest`

svilenvelikov (approved):

yordanalexandrov:
Jenkins retry build

PR #855, opened 2023-02-03
Reorganize the code in query editor directive

svilenvelikov:
This is about to verify if it will work. If it does it'll be much more cleaner and easier to read and refactor.


svilenvelikov:
Jenkins retry build

PR #856, opened 2023-02-03
GDB-7889 Refactor error handling in monitor view

yordanalexandrov:
## What
Refactor the error handling so that an error of one request will not hide whole page

## Why
This will allow the user to see available data, even when some requests fail.

## How
- refactored the error handling and made it more generic
- if an error is encountered, the request is retried several times, set in MAX_RETRIES constant

### DIFF
--- a/src/js/angular/resources/controllers.js
+++ b/src/js/angular/resources/controllers.js
@@ -97,7 +97,6 @@ resourcesCtrl.controller('ResourcesCtrl', ['$scope', '$timeout', 'MonitoringRest
             return $repositories.getActiveRepository();
         };
 
-        // TODO: Organize the following somehow. These are fetcher functions
         const getResourceMonitorData = function() {
             return MonitoringRestService.monitorResources()
                 .then(function(response) {
@@ -157,24 +156,22 @@ resourcesCtrl.controller('ResourcesCtrl', ['$scope', '$timeout', 'MonitoringRest
                 });
         };
 
-
-        // TODO: Organize the following somehow
         const getData = (monitor) => {
             if (hasMonitorError(monitor.error)) {
                 if (monitor.error.retries === MAX_RETRIES) {
                     if (monitor.poll) {
                         $timeout.cancel(monitor.poll);
                     }
-                    return;
+                    return Promise.resolve();
                 }
                 monitor.error.retries++;
             }
             return monitor.fetchFn()
                 .then(() => {
-                    clearError(monitor.error);
+                    clearMonitorError(monitor);
                 })
                 .catch((error) => {
-                    setError(monitor.error, error);
+                    setMonitorError(monitor, error);
                 })
                 .finally(() => {
                     if (monitor.poll) {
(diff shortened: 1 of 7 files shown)


src/js/angular/resources/controllers.js
```diff
@@ -65,211 +66,215 @@ resourcesCtrl.controller('ResourcesCtrl', ['$scope', '$timeout', 'MonitoringRest
...
+            dataSetter(timestamp, data);
+        };
+        const setError = (errorHolder, error) => {
+            errorHolder.hasError = !!error;
+            if (errorHolder.hasError) {
+                errorHolder.message = getError(error);
+            }
+            $scope.error = Object.values($scope.monitors).some((monitor) => monitor.error.hasError);
+        };
+        const clearError = (errorHolder) => {
```
svilenvelikov, line 215: Maybe this should be `updateError` because you may not clear the error. Right?

src/js/angular/resources/controllers.js
```diff
@@ -65,211 +66,215 @@ resourcesCtrl.controller('ResourcesCtrl', ['$scope', '$timeout', 'MonitoringRest
...
-                            $timeout.cancel(timer);
-                        });
+        // TODO: Organize the following somehow
+        const getData = (monitor) => {
+            if (hasMonitorError(monitor.error)) {
+                if (monitor.error.retries === MAX_RETRIES) {
+                    if (monitor.poll) {
+                        $timeout.cancel(monitor.poll);
                     }
+                    return;
```
svilenvelikov, line 168: I'm a bit confused by this. In the other branch, this function is supposed to return a Promise, which means that you has to handle this return somehow.

src/js/angular/resources/controllers.js
```diff
@@ -65,211 +66,215 @@ resourcesCtrl.controller('ResourcesCtrl', ['$scope', '$timeout', 'MonitoringRest
...
+                setChartData(data, dataSetter);
+
+                if ($scope.firstLoad) {
+                    $scope.firstLoad = false;
+
+                    const timer = $timeout(function() {
+                        $scope.loader = false;
+                    }, 500);
+
+                    $scope.$on('$destroy', function() {
```
svilenvelikov, line 198: There are 3 destroy callbacks registered in this controller. I think I know why you did it this way, but it's kind of messy. IMO better register these timers somewhere in a higher scope and clear them at once in in a single destroy.

svilenvelikov (changes requested):

src/js/angular/resources/controllers.js
```diff
@@ -65,211 +66,215 @@ resourcesCtrl.controller('ResourcesCtrl', ['$scope', '$timeout', 'MonitoringRest
...
+                retries: 0
+            },
+            fetchFn: getQueryMonitor,
+            poll: null
+        });
 
-                        const timer = $timeout(function () {
-                            $scope.loader = false;
-                        }, 500);
+        $q.all($scope.monitors.map((monitor) => getData(monitor)))
```
svilenvelikov, line 260: Right now, the monitor view has 3 tabs. When I open the view, all these requests are started even though I may never open onother tab in the view. Isn't it better to group these by tab somehow.

src/js/angular/resources/controllers.js
yordanalexandrov: It should be `clearMonitorError` as what it does is to clear the monitor error. The check at the end is just an update to the `error` flag

src/js/angular/resources/controllers.js
yordanalexandrov: I have made it this way so that you can compare different parameters withing the same time frame

src/js/angular/resources/controllers.js
yordanalexandrov: I will simplify

svilenvelikov (approved):

PR #859, opened 2023-02-07
GDB-7910 Set correct axis values for monitor charts

yordanalexandrov:
## What
Set axis values for ticks to whole integers for some charts

## Why
Some data does not make sense to have the ticks as float numbers

## How
- added custom tick calculation logic to create a range based on the maximum data value


svilenvelikov (dismissed):

yordanalexandrov:
Jenkins retry build

yordanalexandrov:
Jenkins retry build

svilenvelikov (approved):

PR #871, opened 2023-02-22
GDB-7948 Improvements in monitoring UI

yordanalexandrov:
## What
Some things need to be improved in the UI

## Why
To make it better, more informative and to improve the user experience

## How
- changed colors to match the ontotext theme
- added bytes to disk storage chart
- refactored the view to add icons with popover for information
- refactored error handling to show errors in charts sections
- improved by refactoring the data fetching code
- added fragments to url when changing tabs and ability to select specific tab when url has fragment
- renamed some labels


src/js/angular/resources/chart-models/chart-data.js
```diff
@@ -51,25 +55,16 @@ export class ChartData {
...
-
-    setSubTitle(subTitle) {
-        this.chartOptions.subtitle.enable = true;
-        this.chartOptions.subtitle.html = subTitle;
+        const subTitle = keyValues.map((keyValue) => {
+            let values = keyValue.value;
+            if (Array.isArray(keyValue.value)) {
+                values = keyValue.value.join('/');
+            }
+            return keyValue.label + (keyValue.value !== undefined ? `<span class="data-value">${values}</span>` : '');
```
svilenvelikov, line 65: This `keyValue.value !== undefined ? ...` can be just `values ? ...` or `values.length ? ...`

src/js/angular/resources/chart-models/chart-data.js
```diff
@@ -51,25 +55,16 @@ export class ChartData {
...
-        this.chartOptions.title.text = title;
-    }
-
-    getSubTitle() {
-    }
-
-    setSubTitle(subTitle) {
-        this.chartOptions.subtitle.enable = true;
-        this.chartOptions.subtitle.html = subTitle;
+        const subTitle = keyValues.map((keyValue) => {
```
svilenvelikov, line 60: If subTitle is actually an array, you can rename it to plural or something

src/js/angular/resources/chart-models/chart-data.js
yordanalexandrov: added additional logic to return from the method if `value` is array and its length is 0

src/js/angular/resources/chart-models/chart-data.js
yordanalexandrov: renamed

svilenvelikov (approved):

PR #872, opened 2023-02-27
GDB-7844 Update angular bootstrap ui library

yordanalexandrov:
## What
The workbench uses the library's components including popover and must be updated

## Why
With upcoming version 114 of chrome the Popover API will clash with angular bootstrap ui popover component as the names are the same. New version of library adds `uib-` prefixes to components to prevent such future cases

## How
- updated the library to 0.14.3 as the last compatible with the workbench version
- updated components with the new prefix


svilenvelikov (approved):

yordanalexandrov:
Jenkins retry build

svilenvelikov (approved):

PR #881, opened 2023-03-07
GDB-7954: Add pagination control to all yasr plugins

boyan-tonchev:
## What
Added pagination to yasr component.

## Why
Some queries returns a lot of results. To show all of it we need a lot of resource and visualization will be slow.

## How
 - Updates "ontotext-yasgui-web-component".
 - Refactoring query event. It is registered in output handlers.

### DIFF
--- a/src/js/angular/sparql-editor/controllers.js
+++ b/src/js/angular/sparql-editor/controllers.js
@@ -296,57 +296,59 @@ function SparqlEditorCtrl($scope, $location, $jwtAuth, $repositories, toastr, $t
     /**
      * Handles the "query" event emitted by the ontotext-yasgui. The event is fired immediately before sending the
      * request and the request object can be altered here, and it will be sent with these changes.
-     * @param {object} queryResponseEvent - the event payload containing the query and the request object.
+     * @param {QueryRequestEvent} queryResponse - the event payload containing the query and the request object.
      */
-    const queryHandler = (queryResponseEvent) => {
-        updateRequestHeaders(queryResponseEvent.request);
-        changeEndpointByQueryType(queryResponseEvent.queryMode, queryResponseEvent.request);
-        const pageNumber = queryResponseEvent.request._data['pageNumber'] ? parseInt(queryResponseEvent.request._data['pageNumber']) : undefined;
-        queryResponseEvent.request._data['pageNumber'] = undefined;
-        const pageSize = queryResponseEvent.request._data['pageSize'] ? parseInt(queryResponseEvent.request._data['pageSize']) : undefined;
-        queryResponseEvent.request._data['pageSize'] = undefined;
+    const queryHandler = (queryResponse) => {
+        updateRequestHeaders(queryResponse.request);
+        changeEndpointByQueryType(queryResponse.queryMode, queryResponse.request);
+        const pageNumber = queryResponse.getPageNumber();
+        const pageSize = queryResponse.getPageSize();
         if (pageSize && pageNumber) {
-            queryResponseEvent.request._data['offset'] = (pageNumber - 1) * (pageSize -1);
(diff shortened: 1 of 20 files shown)


src/js/angular/sparql-editor/controllers.js
```diff
@@ -300,6 +292,78 @@ function SparqlEditorCtrl($scope, $location, $jwtAuth, $repositories, toastr, $t
...
+
+    /**
+     * Handles the "query" event emitted by the ontotext-yasgui. The event is fired immediately before sending the
+     * request and the request object can be altered here, and it will be sent with these changes.
+     * @param {object} queryResponseEvent - the event payload containing the query and the request object.
+     */
+    function queryHandler(queryResponseEvent) {
+        updateRequestHeaders(queryResponseEvent.request);
+        changeEndpointByQueryType(queryResponseEvent.queryMode, queryResponseEvent.request);
+        const pageNumber = queryResponseEvent.request._data['pageNumber'] ? parseInt(queryResponseEvent.request._data['pageNumber']) : undefined;
```
svilenvelikov, line 304: I wonder, can we have a request object to some model mapper? As far as I can see, do a lot of checks here and there for the properties pageSize, offset, etc. So it might be convenient to have these in a mapper returning some internal representation of this request object. You can check here `src/js/angular/rest/mappers/saved-query-mapper.js` I have created some simple mapping functions.

src/js/angular/sparql-editor/controllers.js
```diff
@@ -300,6 +292,78 @@ function SparqlEditorCtrl($scope, $location, $jwtAuth, $repositories, toastr, $t
     // ================================
     // =     Setup output handlers    =
     // ================================
+
+    /**
+     * Handles the "query" event emitted by the ontotext-yasgui. The event is fired immediately before sending the
+     * request and the request object can be altered here, and it will be sent with these changes.
+     * @param {object} queryResponseEvent - the event payload containing the query and the request object.
+     */
+    function queryHandler(queryResponseEvent) {
```
svilenvelikov, line 301: Please change all these new functions to function expressions. See the other functions in the controller.

src/js/angular/sparql-editor/controllers.js
```diff
@@ -300,6 +292,78 @@ function SparqlEditorCtrl($scope, $location, $jwtAuth, $repositories, toastr, $t
...
+        updateRequestHeaders(countQueryResponseEvent.request);
+        changeEndpointByQueryType(countQueryResponseEvent.queryMode, countQueryResponseEvent.request);
+        countQueryResponseEvent.request._data['pageSize'] = undefined;
+        countQueryResponseEvent.request._data['pageNumber'] = undefined;
+        countQueryResponseEvent.request._data['count'] = 1;
+    }
+    outputHandlers.set(EventDataType.COUNT_QUERY, countQueryResponseHandler);
+
+    function extractTotalElements(countResponse) {
+        if (!countResponse || !countResponse.body) {
```
svilenvelikov, line 330: Here as well we can benefit from a response mapper and a model the parsing and logic can be encapsulated

src/js/angular/utils/yasgui-utils.js
```diff
@@ -8,11 +8,42 @@ export const toYasguiOutputModel = ($event) => {
...
+        case EventDataType.COUNT_QUERY_RESPONSE:
+            return buildCountQueryResponseModel(eventData);
+        case EventDataType.QUERY:
+            return buildQueryModel(eventData);
         default:
             return eventData;
     }
 };
 
+export const buildCountQueryResponseModel = (eventData) => {
```
svilenvelikov, line 22: These function expressions should be declared before their use, e.g. before the toYasguiOutputModel function.

svilenvelikov (approved):

PR #889, opened 2023-03-13
GDB-7829 Add video recording and artefact upload to jenkins

antonsgeorgiev:
- enabled video recording
- added cypress-multi-reporters and mocha-junit-reporter to generate xml reports
- added uploading to jenkins of reports, screenshots and videos

### DIFF
--- a/test-cypress/plugins/index.js
+++ b/test-cypress/plugins/index.js
@@ -12,6 +12,7 @@
 // the project's config changing)
 
 const _ = require("lodash");
+const del = require('del');
 module.exports = (on, config) => {
     // `on` is used to hook into various events Cypress emits
     // `config` is the resolved Cypress config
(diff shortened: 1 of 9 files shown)


test-cypress/plugins/index.js
```diff
@@ -11,6 +11,8 @@
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
 
+const _ = require("lodash");
```
svilenvelikov, line 14: I can't see where the lodash lib comes from? It's not listed in the dependencies in package.json

test-cypress/plugins/index.js
```diff
@@ -26,4 +28,18 @@ module.exports = (on, config) => {
             'cypress-logs|txt': 'txt'
         }
     });
+
+    // keep only the videos for the failed specs
+    on('after:spec', (spec, results) => {
+        if (results && results.video) {
+            // Do we have failures for any retry attempts?
+            const failures = _.some(results.tests, (test) => {
```
svilenvelikov, line 36: You actually don't need lodash for this. You can rewrite it something like this
```javascript
const failures = results.tests.find((test) => {
  return test.attempts.find((attempt) => attempt.state === 'failed');
});
```

yordanalexandrov: Used JS stream functions.

svilenvelikov (approved):

PR #892, opened 2023-03-20
GDB-8049 Change tooltip library

yordanalexandrov:
## What
Change the directive that shows tooltips

## Why
Currently angular-ui-bootstrap provides the directive for tooltips. This implementation has problems with positioning when near the edge

## How
- used a directive from another library which provides a simple implementation of tooltip
- copied and modified the directive so that it repositions if bounding box changes
- added option to accept different triggers

### DIFF
--- a/src/js/angular/core/directives/queryeditor/templates/query-editor.html
+++ b/src/js/angular/core/directives/queryeditor/templates/query-editor.html
@@ -57,9 +57,9 @@
 
                                     <a href="#" ng-click="querySelected(query.body)"><i class="fa fa-1x fa-caret-right"></i>{{query.name}}</a>
                                     <span>
-                                        <i ng-show="canWriteActiveRepo()" uib-tooltip="Edit query" tooltip-trigger="mouseenter" href="#" class="fa fa-pencil-square-o fa-1x editSampleQuery" ng-click="editQuery(query)"></i>
-                                        <i ng-show="canWriteActiveRepo()" uib-tooltip="Delete query" tooltip-trigger="mouseenter" href="#" class="fa fa-trash-o fa-1x deleteSampleQuery" ng-click="deleteQuery(query.name)"></i>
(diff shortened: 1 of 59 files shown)


src/js/angular/core/directives/queryeditor/templates/query-editor.html
```diff
@@ -76,19 +76,15 @@
                                 ng-show="!notoolbarSaved && isUser()"
                                 ng-click="saveTab(currentQuery.id); saveQuery(currentQuery);"
                                 data-container="#queryEditor"
-                                uib-tooltip="{{'core.create.saved.query' | translate}}"
-                                tooltip-append-to-body="true"
-                                tooltip-trigger="mouseenter">
```
svilenvelikov, line 81: How you handle mouseenter? Is it natively handled by the new lib?

src/js/angular/core/templates/inactive-plugin-warning-page.html
```diff
@@ -3,9 +3,8 @@
         {{'inactive.plugin.warning.for.use' | translate}}
     </p>
     <button class="btn btn-primary btn-lg confirm-btn"
-            uib-tooltip="Activate {{humanReadablePluginName}}"
+            gdb-tooltip="Activate {{humanReadablePluginName}}"
             tooltip-placement="top"
```
svilenvelikov, line 7: As I can see, the default position is `top`, can we omit this config where the default is needed?

src/js/angular/core/directives/queryeditor/templates/query-editor.html
yordanalexandrov: It is handled by the directive by adding event listeners for `mouseenter` and `mouseleave` events. Those are default, but I have made it possible to set custom show/hide trigger to be used if needed.

src/js/angular/core/templates/inactive-plugin-warning-page.html
yordanalexandrov: Yes we can. I was wondering if I want to remove them. I didn't change them because, the default before was also top and those are explicitly set.

svilenvelikov (approved):

PR #895, opened 2023-03-21
GDB-8023: extends ontotext-yasgui notification handler

boyan-tonchev:
## What
Added a new type "info" of notification messages emitted by ontotext-yasgui-web-component.

## Why
When user try to open a new tab, but query is running, then a notification of type info have to be shown.

## How
Extends "notificationMessageHandler" to support info messages.

Additional work:
Increase version of "ontotext-yasgui-web-component".

### DIFF
--- a/src/js/angular/sparql-editor/controllers.js
+++ b/src/js/angular/sparql-editor/controllers.js
@@ -1,4 +1,4 @@
-import {isNumber, merge} from "lodash";
+import {isFunction, merge} from "lodash";
 import {
     savedQueriesResponseMapper,
     queryPayloadFromEvent,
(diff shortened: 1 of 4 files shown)


src/js/angular/sparql-editor/controllers.js
```diff
@@ -391,6 +391,8 @@ function SparqlEditorCtrl($scope, $location, $jwtAuth, $repositories, toastr, $t
             toastr.warning(notificationMessageEvent.message);
         } else if (NotificationMessageType.ERROR === notificationMessageEvent.messageType) {
             toastr.error(notificationMessageEvent.message);
+        } else if (NotificationMessageType.INFO === notificationMessageEvent.messageType) {
```
svilenvelikov, line 394: This started moving aout of control. Can't we just do the following and omit this ugly if-else construct entirely? 
```javascript
toastr[notificationMessageEvent.messageType](notificationMessageEvent.message);
```

boyan-tonchev: This was last notification type, but I refactored it how you suggested.

boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #904, opened 2023-04-03
GDB-8085: implement update query flow

boyan-tonchev:
## What
There is difference between the execution flow of "update" and "select" queries. Implements base functionality of update queries.

## Why
The update queries has not results and YASR have to be not visible. Result info message have to describe how many statements have been affected.

## How
- Updates version of "ontotext-yasgui-web-component".
- Implements callback function "getRepositoryStatementsCount", that returns repository statements count.

### DIFF
--- a/src/js/angular/sparql-editor/controllers.js
+++ b/src/js/angular/sparql-editor/controllers.js
@@ -9,9 +9,15 @@ import {QueryMode} from "../utils/query-types";
 import {downloadAsFile, toYasguiOutputModel} from "../utils/yasgui-utils";
 import {YasrPluginName} from "../../../models/ontotext-yasgui/yasr-plugin-name";
 import {EventDataType} from "../../../models/ontotext-yasgui/event-data-type";
+import 'angular/rest/connectors.rest.service';
+
+const modules = [
+    'ui.bootstrap',
+    'graphdb.framework.rest.connectors.service'
+];
 
 angular
-    .module('graphdb.framework.sparql-editor.controllers', ['ui.bootstrap'])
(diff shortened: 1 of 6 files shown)


src/js/angular/sparql-editor/controllers.js
```diff
@@ -79,11 +85,29 @@ function SparqlEditorCtrl($scope,
...
                     }
-                }
+                },
+                getRepositoryStatementsCount: getRepositoryStatementsCount
             };
         }
     };
 
+    const getRepositoryStatementsCount = () => {
+        return new Promise(function (resolve) {
```
svilenvelikov, line 95: This method has a strange construction. Why do you need to wrap the promise returned by `RDF4JRepositoriesRestService.getRepositorySize()` into another promise?

boyan-tonchev: `RDF4JRepositoriesRestService.getRepositorySize()` returns `HttpPromise` which is a bit different than `Promise`. We can not  chain `HttpPromise` as we expected. Here is an article that describes the problems [AngularJS HttpPromise methods break promise chain](https://medium.com/@ExplosionPills/angularjs-httppromise-methods-break-promise-chain-950c85fa1fe7)
Other way to write the function is:
`const getRepositoryStatementsCount = () => {
        return RDF4JRepositoriesRestService.getRepositorySize()
            .then(function (response) {
                const number = parseInt(response.data);
                return $q.when(number);
            })
            .catch(function (data) {
                const params = {
                    repo: $repositories.getActiveRepository(),
                    error: getError(data)
                };
                toastr.warning($translate.instant('query.editor.repo.size.error', params));
            });
    };`

PR #906, opened 2023-04-05
GDB-8115: Implement abort query button

boyan-tonchev:
## What
Added ability to abort a query. When the user clicks on "Run" button, an "Abort query" button will be displayed. If the user click on this button, the running query will be aborted.

## Why
WB has such functionality. This functionality will give to the users opportunity to canceled long-running queries.

## How
- Increases "ontotext-yasgui-web-component" version.
- Implements function "onAbortQuery".

### DIFF
--- a/src/js/services/ontotext-yasgui-web-component.service.js
+++ b/src/js/services/ontotext-yasgui-web-component.service.js
@@ -1,0 +1,38 @@
+const modules = [];
+
+angular
+    .module('graphdb.framework.ontotext-yasgui-web-component', modules)
+    .factory('OntotextYasguiWebComponentService', OntotextYasguiWebComponentService);
+
+OntotextYasguiWebComponentService.$inject = ['MonitoringRestService', 'RDF4JRepositoriesRestService', '$repositories', 'toastr', '$translate'];
+
+function OntotextYasguiWebComponentService(MonitoringRestService, RDF4JRepositoriesRestService, $repositories, toastr, $translate) {
+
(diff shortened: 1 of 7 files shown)


src/js/services/ontotext-yasgui-web-component.service.js
```diff
@@ -0,0 +1,38 @@
+const modules = [];
+
+angular
+    .module('graphdb.framework.ontotext-yasgui-web-component', modules)
+    .factory('OntotextYasguiWebComponentService', OntotextYasguiWebComponentService);
+
+OntotextYasguiWebComponentService.$inject = ['MonitoringRestService', 'RDF4JRepositoriesRestService', '$repositories', 'toastr', '$translate'];
+
+function OntotextYasguiWebComponentService(MonitoringRestService, RDF4JRepositoriesRestService, $repositories, toastr, $translate) {
```
svilenvelikov, line 9: hmm it was a good idea for this service. Maybe not the best name for it, but we will probably definetly need a dedicated service for the sparql editor.

svilenvelikov (approved):

src/js/services/ontotext-yasgui-web-component.service.js
boyan-tonchev: Yes, you are right about the name. I named it that way so we don't confuse our colleagues when we start merging into master. When we clean up the old yasgui code, we can rename it to a simpler name "YasguiService" or something similar.

PR #918, opened 2023-05-04
GDB-8129: Long IRIs are cut from the SPARQL results view

boyan-tonchev:
## What
When IRIs in SPARQL results are shown and don’t fit in the available column size they aren’t wrapped to the next line. This varies between browsers and versions.

## Why
Browsers have different defaults about word-wrapping.

## How
Updates YASR bundled js file that include follow changes:

Added word break rules depends on result binding type.
- For IRIs and blank nodes:
 - adds word-wrap: break-word to all cell with IRI content;
 - inserts <wbr> at strategic places: after :, /, -, _ but treat sequences of those as one thing, e.g. http://<wbr>example…
 - set attribute lang="xx" on the element
- For literals:
 - adds word-wrap: break-word to all cell with literal content;
 - added hyphens: auto (+ -moz- and -webkit- prefixed version) to all cell with literal content;
  - if the literal has a language tag;
   - set attribute lang="<the-tag>" on the element;
  - if literal without language tag;
   - set attribute lang="<the-tag>" on the element;
  - if literal with a datatype;
   - set attribute lang="<the-tag>" on the element;
   - inserts <wbr> before the ^^ of the datatype;
   - inserts <wbr> within the IRI of the datatype following the same rules as for a standalone IRI;

The embedded triples recursively use rules from IRI and literals.


svilenvelikov:
Revert the package-lock.json

src/css/yasr.custom.css
```diff
@@ -98,15 +98,16 @@
 }
 
 .yasr .uri-cell {
-  display: inline;
-  word-break: normal !important;
-  word-wrap: normal !important;
+    display: inline;
+    word-break: normal !important;
+    word-wrap: break-word !important;
```
svilenvelikov, line 103: I don't get this. What's the point in using both `word-break: normal` which dosn't split lenghty words when they overflow the box boundaries and `word-wrap: break-word` which BTW should be replaced with `overflow-wrap`? IMO we should leave only the `overflow-wrap: break-word` here. And same for the line 106 below.

boyan-tonchev: I will change `word-wrap` with `overflow-wrap`. `word-break` is set to normal because it has value break-all from original yasr css style.

svilenvelikov (approved):

PR #920, opened 2023-05-04
GDB-8232: executing a saved query from the welcome page

boyan-tonchev:
## What
When I execute a query from "Saved SPARQL queries" in Welcome page, then "sparql" page is loaded with the query and a confirm dialog is opened. When the dialog is closed no matter how: "Yes", "Cancel" or "x" button and refresh the page dialog is opened again.

## Why
When execute a saved query is from Welcome page the url that triggers "sparql" view be opened contains a parameter "execute". This parameter triggers the opening of a confirmation dialog.

## How
Added code that removes "execute" parameter from the url when dialog is closed.


boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #926, opened 2023-05-11
GDB-8183: different results when clicking sort button in yasr result …

boyan-tonchev:
## What
- When run query that returns results and click on a results table column to sort them, the results are sorted differently than those in WB;
- When run query that returns results and click on a results table column to sort them, the numbers of the first column are not consecutive.

## Why
- The [DataTables](https://datatables.net/) column configuration has a rendering function defined, called to format data table cells. The result of the function is a string that is used as cell data, sorting, and filtering. When the render function is called for visualisation purpose, data from server is formatted. For example if the data is uri it is transformed from full uri to short uri with prefix. When the function is called for sorting and filtering it returns original data from the server. This is the reason for the difference with the WB. In WB, the same formatting is used for visualisation, sorting, and filtering;
- The first column is implemented as a simple data column, and when the table is sorted, the first column is rendered according to the sorting algorithm.

## How
Updates version of "ontotext-yasgui-web-component".


boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

PR #934, opened 2023-05-26
GDB-8332 Changed the DELETED message to the one in backend.

sava-savov-ontotext:
Also changed the evaluation if add or remove node buttons should appear from currentNode.nodeState === NodeState.LEADER -> currentLeader, because leader is needed to process this request.


src/js/angular/clustermanagement/controllers.js
```diff
@@ -30,7 +30,7 @@ export const NodeState = {
     READ_ONLY: 'READ_ONLY',
     RESTRICTED: 'RESTRICTED',
     NO_CLUSTER: 'NO_CLUSTER',
-    DELETED: 'DELETED'
+    DELETED: 'Cluster was deleted on this node.'
```
yordanalexandrov, line 33: This is not a label. It is an object that is used as an Enum. Why did you change it's value?

Is this the correct NodeState value comming from the server!?

yordanalexandrov (dismissed):

yordanalexandrov (approved):

sava-savov-ontotext:
Jenkins retry build

PR #935, opened 2023-05-26
GDB-8252 Fix displaying of additional Ontop properties

yordanalexandrov:
## What
Upon editing a repository, the additional Onto properties are not displayed correct. Instead, [object Object] is shown in the text area.

## Why
The server returns an object which is not parsed and is set in the textbox. The initial talk was for the UI to receive a string so there is no need to parse it, but this has changed.

## How
- added function to convert additional properties values to string

### DIFF
--- a/src/js/angular/repositories/controllers.js
+++ b/src/js/angular/repositories/controllers.js
@@ -44,7 +44,9 @@ const parseMemberParamIfNeeded = function (param) {
 };
 
 const convertAdditionalPropertiesParamIfNeeded = function (params) {
-    if (params.additionalProperties && params.additionalProperties.value) {
+    const values = params && params.additionalProperties && params.additionalProperties.value;
+
+    if (values && values instanceof Object) {
         const values = params.additionalProperties.value;
         const valuesAsStrings = Object.keys(values).map((key) => `${key}=${values[key]}`);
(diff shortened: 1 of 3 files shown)


test-cypress/integration/repository/repositories.spec.js
```diff
@@ -386,7 +386,7 @@ describe('Repositories', () => {
     });
 
     //Create Ontop repository and test ontop functionality
-    it('should create an Ontop repository', () => {
+    it.only('should create an Ontop repository', () => {
```
teodossidossev, line 389: only

src/js/angular/repositories/controllers.js
```diff
@@ -43,6 +43,14 @@ const parseMemberParamIfNeeded = function (param) {
     }
 };
 
+const convertAdditionalPropertiesParamIfNeeded = function (params) {
+    if (params.additionalProperties && params.additionalProperties.value) {
+        const values = params.additionalProperties.value;
```
teodossidossev, line 48: are you sure what object is `values` ?

yordanalexandrov: Added a check if values is an Object

teodossidossev:
lgtm

teodossidossev (approved):

svilenvelikov (approved):

PR #942, opened 2023-06-13
GDB-7640: Make the Ontop edit repository windows bigger and use a mon…

boyan-tonchev:
…ospaced font

## What
- Increase the width and height of update file dialog;
- Change font of the dialog content;
- Added possibility to pass more specific title to "Update file dialog".

## Why
Once an Ontop repository is created the individual files can be edited from the edit repository view. The presented text editor is very small and uses a proportional font.

## How
 - Changed style of dialog.
 - Extended "Update file dialog" with possibility to pass dialog title as parameter.

### DIFF
--- a/src/js/angular/repositories/ontop-repo.directive.js
+++ b/src/js/angular/repositories/ontop-repo.directive.js
@@ -156,13 +156,24 @@ function ontopRepoDirective($uibModal, RepositoriesRestService, toastr, Upload, 
          * @param {OntopFileInfo} ontopFileInfo
          */
         $scope.editFile = (ontopFileInfo) => {
+
+            let fileName = $scope.repositoryInfo.params[ontopFileInfo.type].label;
+            if (fileName.indexOf(' file')) {
+                // Removes the 'file' word from file label if existed, to not duplicate "file" word with label 'update.file.edit.content.header'.
+                fileName = fileName.substring(0, fileName.length - 5);
+            }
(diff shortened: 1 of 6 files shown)


src/js/angular/repositories/ontop-repo.directive.js
```diff
@@ -156,13 +156,24 @@ function ontopRepoDirective($uibModal, RepositoriesRestService, toastr, Upload,
          * @param {OntopFileInfo} ontopFileInfo
          */
         $scope.editFile = (ontopFileInfo) => {
+
+            let fileName = $scope.repositoryInfo.params[ontopFileInfo.type].label;
+            if (fileName.indexOf(' file')) {
```
teodossidossev, line 161: This check to test the end of the string and remove the word 'file' is not correct. IMO it should be like `if (fileName.indexOf(' file') === fileName.length - 5) {`

yordanalexandrov: Doesn't say it removes it from the end.

teodossidossev: > 

Maybe because there can be names without "file" in them.

boyan-tonchev: Because file is not on French :(. I Remove it from the label and translate it in the label.

teodossidossev (dismissed):

yordanalexandrov (approved):

PR #944, opened 2023-06-16
GDB-8209 Hide role and access rights when using external auth

yordanalexandrov:
## What
Hide the role and access rights columns in users catalog when using external auth provider

## Why
The values in the columns, may not be up-to-date and could mislead the user

## How
- added an if to hide the columns


boyan-tonchev (approved):

yordanalexandrov:
Jenkins retry build

PR #945, opened 2023-06-16
GDB-8413: Translate Ontop Form file import fields to french

boyan-tonchev:
## What
The labels of files fields are not translated on French.

## Why
The labels are passed from the BE, but BE don't know about chosen language.

## How
Added labels. They are displayed depends on file type.

Additional work:
 - The "save" and "cancel" buttons have been rearranged on the left side to be as before. The reason is that when some toast message is displayed, the buttons are below the dialog and cannot be clicked. When we have a discussion about the UI, we will move this dependency to the decision.


boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

boyan-tonchev:
Jenkins retry build

PR #948, opened 2023-06-21
GDB-8419: Snowflake Account identifier is not properly mapped in the URL

boyan-tonchev:
## What
The url of Snowflake driver is not populate properly.

## Why
When jdbc driver is Snowflake then url has different url template than other drivers and the generic calculation of url is not applicable to it.

## How
Implemented different calculation function specific for Snowflake.

### DIFF
--- a/src/js/angular/repositories/ontop-repo.directive.js
+++ b/src/js/angular/repositories/ontop-repo.directive.js
@@ -4,6 +4,7 @@ import {OntopConnectionInformation} from "../../../models/ontop/ontop-connection
 import {OntopFileInfo} from "../../../models/ontop/ontop-file-info";
 import {OntopRepositoryError} from "../../../models/ontop/ontop-repository-error";
 import {OntopDriverData} from "../../../models/ontop/ontop-driver-data";
+import {JdbcDriverType} from "../../../models/ontop/jdbc-driver-type";
 
 // A link to Ontop's website with all Ontop configuration keys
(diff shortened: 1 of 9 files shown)


src/js/angular/repositories/ontop-repo.directive.js
```diff
@@ -419,7 +445,7 @@ function ontopRepoDirective($uibModal, RepositoriesRestService, toastr, Upload,
         };
 
         const validateHostName = () => {
-            if ($scope.selectedDriver.driverType !== $scope.genericDriverType) {
+            if ($scope.selectedDriver.driverType !== JdbcDriverType.GENERIC) {
```
svilenvelikov, line 448: Can you create a method like `isGenericDriver(type)` and reuse it as this check is performed too often here.
The method can be in the OntopDriverData class.

svilenvelikov: Then you can also expose a boolean property isGenericDriver which to be used in the template where this check is also performed many times.

boyan-tonchev: Done

boyan-tonchev:
Jenkins retry build

svilenvelikov (dismissed):

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

PR #949, opened 2023-06-22
GDB-8416: title of repository view not translated

boyan-tonchev:
## What
The title of repository view does not translated  when change the language.

## Why
The value of title calculates into controller and is not updated when language is changed.

## How
Added listeners for event "$translateChangeSuccess", that is fired after language is changed.


svilenvelikov (approved):

yordanalexandrov:
Jenkins retry build

PR #950, opened 2023-06-23
GDB-8437: introduce a way to expand error messages in GraphDB workbench

boyan-tonchev:
## What
When exception is thrown after executing of sparql query, the workbench shows the error. But if the error is too big the error area is too long. Most of the users do not need to see the entire message.

## Why
WB displays all error messages no matter of its length.

## How
Added functionality that checks error message length and if it is too long a small part of message is displayed accompanied by a 'Show full error message' button to expand it. When the button is clicked all error will be displayed accompanied by a 'Show less error message' button to show less part of message.

### DIFF
--- a/src/js/angular/core/directives/queryeditor/templates/query-editor.html
+++ b/src/js/angular/core/directives/queryeditor/templates/query-editor.html
@@ -220,20 +220,16 @@
                 </div>
 
                 <div ng-if="!fullErrorMessage && sparqlQueryErrorInfo && sparqlQueryErrorInfo.shortenErrorMessage">
-                    <div class="show-full-message">
-                        <a href="#" ng-click="showFullErrorMessage(true)">
-                            <span class="show-full-message-btn">{{'query.editor.error.show.full.message' | translate}}</span>
-                        </a>
-                    </div>
+                    <a href="#" class="show-full-message-link" ng-click="showFullErrorMessage(true)">
+                        {{'query.editor.error.show.full.message' | translate}}
+                    </a>
                     <div class="plaintext"> {{sparqlQueryErrorInfo.shortenErrorMessage}}</div>
                 </div>
 
                 <div ng-if="fullErrorMessage || sparqlQueryErrorInfo && !sparqlQueryErrorInfo.shortenErrorMessage">
-                    <div class="show-less-message">
-                        <a href="#" ng-click="showFullErrorMessage(false)" ng-if=" sparqlQueryErrorInfo && sparqlQueryErrorInfo.shortenErrorMessage">
-                            <span class="show-less-message-btn">{{'query.editor.error.show.less.message' | translate}}</span>
-                        </a>
-                    </div>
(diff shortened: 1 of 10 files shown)


src/js/angular/core/directives/queryeditor/templates/query-editor.html
```diff
@@ -219,7 +219,24 @@ <h6 class="alert alert-warning no-icon additional" ng-show="!queryIsRunning && !
                     <h6>{{'common.error' | translate}}<span ng-show="yasr.results.getException().status > 0"> {{yasr.results.getException().status}}: {{yasr.results.getException().statusText}}</span></h6>
                 </div>
 
-                <div class="plaintext">{{yasr.results.getException().responseText}}</div>
+                <div ng-if="!fullErrorMessage && sparqlQueryErrorInfo && sparqlQueryErrorInfo.shortenErrorMessage">
+                    <div class="show-full-message">
+                        <a href="#" ng-click="showFullErrorMessage(true)">
+                            <span class="show-full-message-btn">{{'query.editor.error.show.full.message' | translate}}</span>
```
svilenvelikov, line 225: But it's a link actually.

svilenvelikov: Also, why do we need this additional span around the link text?

src/js/angular/core/directives/queryeditor/templates/query-editor.html
```diff
@@ -219,7 +219,24 @@ <h6 class="alert alert-warning no-icon additional" ng-show="!queryIsRunning && !
                     <h6>{{'common.error' | translate}}<span ng-show="yasr.results.getException().status > 0"> {{yasr.results.getException().status}}: {{yasr.results.getException().statusText}}</span></h6>
                 </div>
 
-                <div class="plaintext">{{yasr.results.getException().responseText}}</div>
+                <div ng-if="!fullErrorMessage && sparqlQueryErrorInfo && sparqlQueryErrorInfo.shortenErrorMessage">
+                    <div class="show-full-message">
```
svilenvelikov, line 223: There are too much containers around these links here. Can we simplify this?

svilenvelikov: Also, this `show-full-message` class sounds like something which should be put on some actionable component like button or link, but here it is set on a div tag.

src/js/angular/core/directives/queryeditor/templates/query-editor.html
boyan-tonchev: refactored

src/js/angular/core/directives/queryeditor/templates/query-editor.html
boyan-tonchev: refactored

svilenvelikov (approved):

yordanalexandrov (dismissed):

yordanalexandrov (dismissed):

yordanalexandrov (approved):

PR #951, opened 2023-06-26
GDB-8395: Workbench throwing console warnings for invalid JSON

boyan-tonchev:
## What
- There was a warning message:
```
unable to locate 'src/guides' at '/home/boyan/Git/graphdb-workbench/src/guides'
```
when open a page of workbench.
- There was a warning message:
```
     DevTools failed to load source map: Could not load content for http://localhost:7200/main.35e30fd1f1326e91ed6c.bundle.js.map: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
```
in the Chrome console when opens a page in production.

## Why
- There was legacy configuration of webpack copy plugin from "user interactive guide" functionality.
- During the build process sources are skipped but there are references to the sources into minified files. At the end of every one file has something like "//# sourceMappingURL=184.ed3e402118e64c17a70b.bundle.js.map". This causes Chrome to try loading the source file.

## How
- Removes the legacy configuration.
- Disable devtool on production. After I turned off "devtool" warning messages appeared, informs that assets and entrypoint are big. That's why performance configuration is added. It configured max size of entry points or assets, that can be deployed. The file "swagger-ui.js" is skipped because is too big (2.8MB).

Additional work:
## What
 The test "sparql-error-handling.spec.js" is failing randomly.

## Why
There is a Null Pointer Exception (NPE) occurring with the following error message: "TypeError: Cannot read properties of undefined (reading 'getQueryMode')". This exception is caused by attempting to access the 'getQueryMode' function of Yasqe (window.editor) before it is initialized.

## How
The issue arises when the Cypress test finds the Run button and attempts to click on it. However, sometimes the button becomes visible before Yasqe (window.editor) is fully initialized. During the execution of the 'run query' function, there is a logic check for the query mode that calls the 'getQueryMode' function of Yasqe (window.editor.getQueryMode()). Since the editor is not yet initialized, the exception is thrown due to the undefined property access.

To resolve this issue, it is necessary to ensure that the Yasqe editor is fully initialized before attempting to access its functions or properties, such as 'getQueryMode'.

## How
In the file "query-editor.directive.js," a new boolean flag called "queryEditorIsReady" has been added. This flag is set to true when the query editor is fully prepared, including the creation and overriding of editor functions.

When the "queryEditorIsReady" flag is true, the container that contains various buttons like "Run," "Infer," and "Same as.." is created and displayed accordingly. This ensures that the container and its buttons are only generated and shown when the query editor is fully ready for interaction.

### DIFF
--- a/src/js/angular/core/directives/queryeditor/query-editor.directive.js
+++ b/src/js/angular/core/directives/queryeditor/query-editor.directive.js
@@ -70,6 +70,7 @@ function queryEditorDirective($timeout, $location, toastr, $repositories, Sparql
 
         scope.sparqlQueryErrorInfo = undefined;
         scope.fullErrorMessage = false;
+        scope.queryEditorIsReady = false;
 
         scope.showFullErrorMessage = function (fullErrorMessage) {
             scope.fullErrorMessage = fullErrorMessage;
@@ -85,6 +86,7 @@ function queryEditorDirective($timeout, $location, toastr, $repositories, Sparql
 
         $timeout(function () {
             drawQueryEditor(scope);
+            scope.queryEditorIsReady = true;
         }, 150);
     }
 
@@ -194,7 +196,6 @@ function queryEditorDirective($timeout, $location, toastr, $repositories, Sparql
             scope.currentQuery.query = window.editor.getValue();
             scope.currentQuery.queryType = window.editor.getQueryType();
         });
-
 
         function selectTab(id) {
             $timeout(function () {
--- a/test-cypress/integration/sparql/sparql-error-handling.spec.js
+++ b/test-cypress/integration/sparql/sparql-error-handling.spec.js
@@ -4,7 +4,7 @@ const LONG_ERROR_MESSAGE = "Lorem ipsum dolor sit amet, consectetur adipisicing 
 const MAX_VISIBLE_ERROR_CHARACTERS = 160;
 const SHORTEN_PART_OFF_ERROR_MESSAGE = LONG_ERROR_MESSAGE.substring(0, MAX_VISIBLE_ERROR_CHARACTERS);
 const SHORT_ERROR_MESSAGE = LONG_ERROR_MESSAGE.substring(0, MAX_VISIBLE_ERROR_CHARACTERS - 1);
-describe.skip('Error handling', () => {
+describe('Error handling', () => {
     let repositoryId;
 
     beforeEach(() => {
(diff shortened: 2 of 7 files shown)


webpack.config.prod.js
```diff
@@ -9,6 +9,14 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 
 module.exports = merge(commonConfig, {
     mode: 'production',
+    devtool: '(none)',
```
yordanalexandrov, line 12: Does it have to say `(none)` or you just omit it?

boyan-tonchev: Hm, my .. It just have to be  omitted.  I will remove it.

src/js/angular/core/directives/queryeditor/query-editor.directive.js
```diff
@@ -85,6 +86,7 @@ function queryEditorDirective($timeout, $location, toastr, $repositories, Sparql
 
         $timeout(function () {
             drawQueryEditor(scope);
+            scope.queryEditorIsReady = true;
```
svilenvelikov, line 89: How this guarantees that the editor would be ready?

test-cypress/integration/sparql/sparql-error-handling.spec.js
```diff
@@ -71,10 +71,10 @@ describe.skip('Error handling', () => {
         SparqlSteps.getShowLessExceptionMessage().should('not.exist');
     });
 
-    const stubErrorResponse = (statusCode, errorMessage) => {
-        cy.intercept('POST', '/repositories/' + repositoryId, {
+    const stubErrorResponse = (repositoryId, statusCode, errorMessage) => {
```
svilenvelikov, line 74: We introduced stubbing functions in the yasgui integration branch. Can't we start applying the approach in the main as well?

src/js/angular/core/directives/queryeditor/query-editor.directive.js
boyan-tonchev: after "drawQueryEditor" the window.editor has value the YASQE reference with all overridden functions

svilenvelikov: ahh yes, that freaking function ...

test-cypress/integration/sparql/sparql-error-handling.spec.js
boyan-tonchev: Done

svilenvelikov (approved):

PR #955, opened 2023-07-04
Moves before update query functionality

boyan-tonchev:
## What
Moves the before update functionality from yasgui-component.directive.js to sparql view controller.

## Why
At this point this functionality is sparql view specific.

## How

Additional work:
- removes unnecessary service ontotext-yasgui-web-component.service.js;
- moves all models from root directory to js/angular.


boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #957, opened 2023-07-05
GEDB-8479 Refactoring of similarity indexes view

boyan-tonchev:
## What
 - Reorganize functions in "similarity-list.controller.js";
 - Optimizes loading of view;
 - Added repository change listener;
 - Optimizes the checks can write active repo and is active repo GraphDB repo;
 - Optimizes loading of similarity indexes.
 
## Why
Cleans the code before updating the yasgui. Reduces the changes that have to be made.

## How
 - Separates the functions depends on them usage and scope;
 - The view is initialized when active repository is set.
 - Added repository change listener. When the active repository is changed, the list has to be reloaded.
 - Removes functions "getActiveRepository", "canWriteActiveRepo", "isActiveRepoOntopType" and "isActiveRepoFedXType" from the view template. Added property that holds the value of those functions. The values are calculated when component is loaded and when the active repository changes.
 - Added a check to prevent reinitialization of the loading interval if it is already initialized.

### DIFF
--- a/src/js/angular/similarity/controllers/similarity-list.controller.js
+++ b/src/js/angular/similarity/controllers/similarity-list.controller.js
@@ -17,49 +17,18 @@ function SimilarityCtrl($scope, $interval, toastr, $repositories, $licenseServic
     $scope.pluginName = 'similarity';
     $scope.pluginIsActive = true;
 
-    $scope.setPluginIsActive = function (isPluginActive) {
-        $scope.pluginIsActive = isPluginActive;
-    };
+    $scope.loading = false;
 
-    const literalForQuery = function (literal) {
-        return '"' + literal + '"';
-    };
-
-    // TODO: Fix cases when this function is called with undefined
-    const iriForQuery = function (iri) {
-        // Do not put brackets on nested triples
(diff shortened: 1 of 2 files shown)


src/js/angular/similarity/controllers/similarity-list.controller.js
```diff
@@ -75,76 +47,17 @@ function SimilarityCtrl($scope, $interval, toastr, $repositories, $licenseServic
...
-                }).error(function (data) {
-                    toastr.error(getError(data), $translate.instant('get.namespaces.error.msg'));
-                });
-        }
-    });
-
-    function checkAutocompleteStatus() {
-        if ($licenseService.isLicenseValid()) {
-            $scope.getAutocompletePromise = AutocompleteRestService.checkAutocompleteStatus();
+    $scope.updateSimilarityIndexes = function () {
```
svilenvelikov, line 50: You can just use arrow functions everywhere for consistency.

svilenvelikov (approved):

src/js/angular/similarity/controllers/similarity-list.controller.js
boyan-tonchev: Ok, I will do it.

PR #968, opened 2023-07-17
GDB-8369 Cluster View broken toast error for lack of user permissions

Emskiqq:
GDB-8369 Ascertaining whether the error is a 403 and calls getError(error.data, error.status) if it is

### DIFF
--- a/src/js/angular/clustermanagement/controllers.js
+++ b/src/js/angular/clustermanagement/controllers.js
@@ -241,7 +241,7 @@ function ClusterManagementCtrl($scope, $http, $q, toastr, $repositories, $uibMod
                 })
                 .catch((error) => {
                     const failMessage = $translate.instant('cluster_management.delete_cluster_dialog.notifications.fail_delete');
-                    handleAddRemoveErrors(error.data, error.status, failMessage);
+                    handleErrors(error.data, error.status, failMessage);
                 })
                 .finally(() => {
                     $scope.setLoader(false);
@@ -323,7 +323,7 @@ function ClusterManagementCtrl($scope, $http, $q, toastr, $repositories, $uibMod
                 })
                 .catch((error) => {
                     const failMessageTitle = $translate.instant('cluster_management.cluster_page.notifications.add_nodes_fail');
-                    handleAddRemoveErrors(error.data, error.status, failMessageTitle);
+                    handleErrors(error.data, error.status, failMessageTitle);
                 })
                 .finally(() => {
                     $scope.setLoader(false);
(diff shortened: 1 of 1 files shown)


Emskiqq:
Jenkins retry build

src/js/angular/clustermanagement/controllers.js
```diff
@@ -241,7 +241,7 @@ function ClusterManagementCtrl($scope, $http, $q, toastr, $repositories, $uibMod
                 })
                 .catch((error) => {
                     const failMessage = $translate.instant('cluster_management.delete_cluster_dialog.notifications.fail_delete');
-                    const failedNodesList = Object.keys(error.data)
+                    const failedNodesList = error.status === 403 ? getError(error.data, error.status) : Object.keys(error.data)
```
yordanalexandrov, line 244: Is it only 403 that needs to be handled like this?

IMO all errors should be handled with `getError()`, except the one that returns the failed delete from the cluster.

yordanalexandrov, line 244: Also change the name of the variable if it might contain different messages

yordanalexandrov, line 244: The other actions in this view have this case handled. You can the same aproach here as well, so it is consistent.

yordanalexandrov:
Also please change the name of the MR as it has nothing to do with what the changes are.

src/js/angular/clustermanagement/controllers.js
yordanalexandrov, line 244: The other actions in this view have this case handled. You can the same aproach here as well, so it is consistent.

yordanalexandrov (changes requested):

src/js/angular/clustermanagement/controllers.js
```diff
@@ -241,9 +241,7 @@ function ClusterManagementCtrl($scope, $http, $q, toastr, $repositories, $uibMod
                 })
                 .catch((error) => {
                     const failMessage = $translate.instant('cluster_management.delete_cluster_dialog.notifications.fail_delete');
-                    const failedNodesList = error.status === 403 ? getError(error.data, error.status) : Object.keys(error.data)
-                        .reduce((message, key) => message += `<div>${key} - ${error.data[key]}</div>`, '');
-                    toastr.error(failedNodesList, failMessage, {allowHtml: true});
+                    handleAddRemoveErrors(error.data, error.status, failMessage);
```
yordanalexandrov, line 244: Rename the function to something more generic `handleErrors` perhaps

yordanalexandrov (changes requested):

yordanalexandrov (approved):

PR #973, opened 2023-07-20
GDB-8442 Allow access to cluster view for all authenticated users

yordanalexandrov:
## What
Allow access to cluster view from the menu or by url to all authenticated users and restrict actions to admin users only

## Why
All authenticated users should have access to the cluster view and only admin users should be able to perform actions

## How
- added a check to see if user has admin access to create a cluster and changed view
- added ng-if to action buttons to hide them is user is not admin
- changed main menu item role to ROLE_USER to allow all authenticated users to see it

### DIFF
--- a/src/js/angular/clustermanagement/cluster-drawing.service.js
+++ b/src/js/angular/clustermanagement/cluster-drawing.service.js
@@ -37,7 +37,7 @@ export function setCreateClusterZone(hasCluster, clusterZone, translationsMap, h
         textGroup
             .append('text')
             .attr('id', 'no-cluster-label')
-            .text(translationsMap.no_cluster_configured || 'No cluster group configured')
+            .text(translationsMap.no_cluster_configured)
             .attr('y', -50)
             .classed('h2', true)
             .style('text-anchor', "middle");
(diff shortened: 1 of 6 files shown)


boyan-tonchev (approved):

src/js/angular/clustermanagement/cluster-drawing.service.js
```diff
@@ -40,11 +41,13 @@ export function setCreateClusterZone(hasCluster, clusterZone, translationsMap) {
...
-        textGroup.append('text')
-            .attr('id', 'create-cluster-label')
-            .classed('h3', true)
-            .text(translationsMap.create_cluster_btn || 'Click here to create a cluster')
-            .style('text-anchor', "middle");
+        if (hasAccess ){
+            textGroup.append('text')
+                .attr('id', 'create-cluster-label')
+                .classed('h3', true)
+                .text(translationsMap.create_cluster_btn || 'Click here to create a cluster')
```
svilenvelikov, line 48: shouldn't this be translated?

yordanalexandrov: I guess they are some sort of fallbacks that got to the master.

I removed them.

All translations are in the `translationsMap`

boyan-tonchev (approved):

PR #978, opened 2023-07-25
Added error icon to similarity query tabs

boyan-tonchev:
## What
 - Added an icon to the top-right corner of the query type tab when the tab's query is invalid.

## Why
 - The similarity indexes consist of two or three queries that are visualized on different tab. If there is an invalid query in a tab that is not currently visible and  the user clicks the "Create" button, they might not easily understand why the index is not being created. The added icon provides a visual indication of the error.

## How
 - An error icon was added to the top-right corner of the query type tab when the tab's query contains an error.

### DIFF
--- a/src/css/bootstrap-graphdb-theme.css
+++ b/src/css/bootstrap-graphdb-theme.css
@@ -1313,6 +1313,17 @@ footer.footer .container-fluid.main-container p {
     content: "\e920";
 }
 
+.tab-error:after {
+    font-family: 'icomoon', sans-serif !important;
+    font-size: 1em;
+    vertical-align: middle;
+    line-height: 1;
+    position: absolute;
+    left: 0.1em;
+    content: "\e920";
+    color: var(--onto-orange);
+}
+
 .alert-success:before {
     content: "\e914";
 }
--- a/src/js/angular/similarity/controllers/create-index.controller.js
+++ b/src/js/angular/similarity/controllers/create-index.controller.js
@@ -184,6 +184,7 @@ function CreateSimilarityIdxCtrl(
     }
 
     $scope.saveSearchQuery = function () {
+        $scope.saveOrUpdateExecuted = true;
         if (!isDirty) {
             goToSimilarityIndexesView();
         }
@@ -192,6 +193,8 @@ function CreateSimilarityIdxCtrl(
             .then(validateSimilarityIndexName)
             .then(validateQuery)
             .then(validateQueryType)
+            .then(validateSearchQuery)
+            .then(validateAnalogicalQuery)
             .then(saveQuery)
             .then(notifySaveSuccess)
             .catch((error) => {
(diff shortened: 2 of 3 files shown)


src/css/bootstrap-graphdb-theme.css
```diff
@@ -1313,6 +1313,17 @@ footer.footer .container-fluid.main-container p {
...
 
+.tab-error:after {
+    font-family: 'icomoon', sans-serif !important;
+    font-size: 1em;
+    vertical-align: middle;
+    line-height: 1;
+    position: absolute;
+    left: 0.1em;
+    content: "\e920";
+    color: var(--onto-orange);
```
yordanalexandrov, line 1324: Shoundn't it be some red color or this is what we use in the WB for errors?

src/js/angular/similarity/controllers/create-index.controller.js
```diff
@@ -277,6 +280,69 @@ function CreateSimilarityIdxCtrl(
...
     }
 
+    $scope.getTabErrorMessage = (similarityQueryTab) => {
+
+        if (!$scope.similarityIndexInfo) {
+            return;
+        }
+
+        if (SimilarityQueryType.DATA === similarityQueryTab) {
+           return getSelectQueryErrorMessage();
```
yordanalexandrov, line 290: Whay not just pass the type and have one method? Build the error key inside based on the type?

src/css/bootstrap-graphdb-theme.css
boyan-tonchev: I couldn't find a constant for errors. I can create a new one, and we can use it in the future."

src/js/angular/similarity/controllers/create-index.controller.js
boyan-tonchev: The method will have to many "if else" operators

src/js/angular/similarity/controllers/create-index.controller.js
```diff
@@ -277,6 +280,69 @@ function CreateSimilarityIdxCtrl(
...
+            return $translate.instant('similarity.error.query.invalid');
+        }
+    }
+
+    const getSearchQueryErrorMessage = () => {
+        if (!$scope.similarityIndexInfo.hasSearchQuery()) {
+            return $translate.instant('similarity.empty.search.query.error');
+        }
+
+        if ($scope.similarityIndexInfo.invalidSelectQueryType) {
```
yordanalexandrov, line 323: Is this correct to be `invalidSelectQueryType` or it should be `search`?

src/js/angular/similarity/controllers/create-index.controller.js
yordanalexandrov: Well, not sure that you would need more than 2 ifs with 3 branches each.

yordanalexandrov: One to check the type and set some properties and the other to check if properties are true and return the correct message.

src/css/bootstrap-graphdb-theme.css
svilenvelikov: No. You shouldn't. Just use the bootstrap's danger color. There is an overriden bootstrap theme which is used here.

boyan-tonchev: Done as @svilenvelikov suggested.

src/js/angular/similarity/controllers/create-index.controller.js
boyan-tonchev: Refactored

src/js/angular/similarity/controllers/create-index.controller.js
boyan-tonchev: fixed

boyan-tonchev:
Jenkins retry build

yordanalexandrov (approved):

svilenvelikov (approved):

PR #985, opened 2023-08-03
GDB-8598 unify buttons position and order in all views

svilenvelikov:
## What
Unifies the order and position of any ok/confirm/cancel like buttons in views and dailogs throughout the workbench.

## Why
We need to improve the consistency and adhere to the user experience where it needs to always be obvious where to expect such buttons to appear and what will be their order.

## How

Clean up the modal service and controller and removed a TODO


svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

PR #986, opened 2023-08-04
GDB-8610 reimplemented the sparql results download as functionality

svilenvelikov:
## What
Reimplemented the sparql results download as functionality to go trough a post request instead of get

## Why
The old implementation used to send a get request with form data containing the sparql query string which happened to break the browser limit for get request size.

## How
* Removed the old form used for sending request data needed for the download and replaced with an actual http.post request and saved result according to provided file type.


boyan-tonchev (approved):

boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

PR #990, opened 2023-08-10
GDB-8601: No execution confirmation for saved queries triggered from the WB home page

boyan-tonchev:
## What
- There is no confirmation dialog before execute query when sparql view is opened from WB home page.
- There are doubled initialization of yasgui component when a page used it is loaded.

## Why
- The feature was not implemented;
- There were two listeners for repository change event.

## How
- Checks if url parameters contains execute param. If it is present and query tha have to be executed is update query, then confirm dialog will be displayed;
- Removes duplication of the listeners.

Additional work:
- Expose tab query model;
- Expose a YasguiComponent class that wraps all available method of ontotext-yasgui-web-component.


boyan-tonchev:
Jenkins retry build

src/js/angular/sparql-editor/controllers.js
```diff
@@ -68,14 +75,21 @@ function SparqlEditorCtrl($scope,
...
+    const initViewFromUrlParams = () => {
+        $scope.updateConfig();
         const queryParams = $location.search();
         if (queryParams.hasOwnProperty(RouteConstants.savedQueryName)) {
             // init new tab from shared saved query link
             initTabFromSavedQuery(queryParams);
         } else if (queryParams.hasOwnProperty(RouteConstants.query)) {
             // init new tab from shared query link
             initTabFromSharedQuery(queryParams);
+        } else if (GuidesService.isActive()) {
```
svilenvelikov, line 90: I don't quite understand the idea here, but Its kind of quirky to me

boyan-tonchev: User guides have steps that explain SPARQL functionality. When the guide is started and the view is open, we always open a new tab to ensure that we don't alter the user's query, as it could be important for them.

svilenvelikov: OK, but why sparql view should know about the existence of the guides? IMO this should not be here

boyan-tonchev: Okay. At this moment, the user guides are broken due to the changes in Yasgui. When I start working on the guides, I will refactor this.

svilenvelikov: Can you create a task for it so that we don't forget to remove it from here.

svilenvelikov (approved):

src/js/angular/sparql-editor/controllers.js
boyan-tonchev: Done [GDB-8625](https://ontotext.atlassian.net/browse/GDB-8625)

PR #993, opened 2023-08-11
GDB-8516 Changed the endpoint configuration to be a function instead of a string.

boyan-tonchev:
## What
The endpoint configuration has been converted to a function rather than a string.

## Why
When endpoint is a string it is persisted in browser local store. When a query is run the value is taken from the store, no matter that the endpoint configuration is changed. If endpoint is function then the function is called every time query is executed.

## How
Changed endpoint to be function.


boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #997, opened 2023-08-15
GDB-8620 sparql view should not be empty when there is no active repository

svilenvelikov:
## What
Sparql view should not be empty when there is no active repository.

## Why
When there is no selected active repository the sparql view is not initialized correctly as it was in the old sparql view. The old implementation used to show a list with the available repositories for the user to select from. Aside from that the yasgui component threw an error because it was unconditionally included in the view while its config was missing in the case of missing repository.

## How
Added check for the yasgui config object before including it in the view. Also added the generic core-errors directive which used to render the repositories list as in the old implementation.


boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

yordanalexandrov (approved):

PR #1006, opened 2023-08-22
GDB-8641: Empty tab when execute RDF* query via resource view

boyan-tonchev:
## What
When clicking on the title of a triple resource in the resource view, a new tab opens in the SPARQL view with an empty tab name. The tab name must be "Unnamed <next number>"

## Why
The addTab function in YASGUI takes a second argument, which is the configuration for the new tab. This configuration includes the tab name. When the configuration is missing, YASGUI generates a tab name. In our case, from WB, we pass an empty string as the tab name configuration, which YASGUI treats as a valid name and uses it.

## How
The default value of the tab name configuration has been changed to be undefined instead of an empty string.


boyan-tonchev:
Jenkins retry build

PR #1011, opened 2023-08-31
GDB-8540: Contrast/color and themes

avataar:
- Adjusted some colors after feedback from Dilyana and some experimentation
- Better separation of themes, including dark mode
- Added theme support for adjusting the colors of the logo
- Brought "onto original" to be more like the actual original theme before all the changes

Fixed GDB-8717 (fallback to default theme when theme not present in settings or theme not available)


svilenvelikov (dismissed):

svilenvelikov:
Recreated in another branch after master was reset back to its 2.3 state

PR #1012, opened 2023-09-12
GDB-8709 Remove option to add roles to admin and repo-manager users

Emskiqq:


src/js/angular/security/templates/user.html
```diff
@@ -181,7 +181,7 @@ <h3>{{'security.workbench.settings.title' | translate}}
 
 			<div class="col-md-6 us-right-col" ng-hide="!isLocalAuthentication()">
 				<div class="card user-role" id="user-roles">
-					<div class="card-block">
+					<div class="card-block" ng-click="resetCustomRoleField()">
```
svilenvelikov, line 184: Only links and buttons should be clickable. Refactor this to have a button.

svilenvelikov: Now after diving deeper in the code, it turned out that current approach would force this click handler to be triggered on any click inside this container which holds the checkboxes with the user roles. This might work as it is, but might lead to hidden bugs in future. Better approach would be to have a watcher for the $scope.userType property and to do this logic in there.

svilenvelikov (approved):

svilenvelikov:
Recreated in another branch after master was reset back to its 2.3 state

PR #1014, opened 2023-09-13
GDB-8760: Implement backup and restore monitoring view in WB

boyan-tonchev:
## What
A new view, 'Backup and Restore' has been added to display running backup or restore operations.

## Why
We aim to provide our clients with the opportunity to easily check if there are ongoing backup or restore operations.

## How
The view displays all currently running backup and restore operations in a table format. The page automatically refreshes every 2 seconds.


boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

src/js/angular/backup-and-restore/controllers.js
```diff
@@ -0,0 +1,65 @@
...
+    $scope.loading = false;
+    $scope.initialized = false;
+
+    /**
+     * @type {BackupAndRestoreInfo[]}
+     */
+    $scope.backupAndRestoreInfos = undefined;
+    $scope.hasClusterOperation = false;
+    let timer = undefined;
+    const UPDATE_BACKUP_AND_RESTORE_INFO_DATA_TIME_INTERVAL = 2000;
```
yordanalexandrov, line 20: I would rather have this on top for easy access.

src/js/angular/backup-and-restore/controllers.js
```diff
@@ -0,0 +1,65 @@
...
+    // =========================
+    // Event handlers
+    // =========================
+    $scope.$on('$destroy', removeAllListeners);
+
+    // =========================
+    // Initializes controller.
+    // =========================
+    loadBackupAndRestoreData();
+    timer = $interval(() => loadBackupAndRestoreData(), UPDATE_BACKUP_AND_RESTORE_INFO_DATA_TIME_INTERVAL);
```
yordanalexandrov, line 64: If the monitor endpoint stalls for some reason, there will be multiple requests, which will resolve out of order and might cause some inconsistencies.

You can chain the next call to the previous. I think in the cluster or in the system monitor controller it is done like this, to prevent the issue.

src/pages/monitor/backup-and-restore.html
```diff
@@ -0,0 +1,74 @@
...
+            <tr>
+                <th class="id-header">{{'view.monitoring.backup_and_restore.id.header' | translate}}</th>
+                <th class="username-header">{{'view.monitoring.backup_and_restore.username.header' | translate}}</th>
+                <th class="recovery-operation-header">{{'view.monitoring.backup_and_restore.recovery_operation_type.header' | translate}}
+                </th>
+                <th class="affected-repository-header">{{'view.monitoring.backup_and_restore.affected_repositories.header' | translate}}
+                </th>
+                <th class="lifetime-header">{{'view.monitoring.backup_and_restore.lifetime.header' | translate}}</th>
+                <th class="snapshot-options-header">{{'view.monitoring.backup_and_restore.snapshot_options.header' | translate}}</th>
+                <th ng-if="hasClusterOperation" class="node-performing-cluster-backup-header">
```
yordanalexandrov, line 34: Maybe we should always display the columns regardless if they have info. If some operation starts it will bounce the table.

src/pages/monitor/backup-and-restore.html
```diff
@@ -0,0 +1,74 @@
...
+    <h1>
+        {{title}}
+        <span class="btn btn-link"
+              uib-popover-template="'js/angular/templates/titlePopoverTemplate.html'"
+              popover-trigger="mouseenter"
+              popover-placement="bottom-right"
+              popover-append-to-body="true">
+            <span class="icon-info text-tertiary"></span>
+        </span>
+        <span onto-loader-new class="backup-and-restore-updater-icon" size="25">
```
yordanalexandrov, line 15: There is one loader here and one at the end of the page. Is this the same way as other pages' heading?

src/js/angular/backup-and-restore/controllers.js
yordanalexandrov: aa, you have a loading flag so this will not happen.

src/js/angular/backup-and-restore/controllers.js
boyan-tonchev: done

src/pages/monitor/backup-and-restore.html
boyan-tonchev: The column contains information only for operations executed on the cluster. Currently, we support only one backup or restore operation at a time, resulting in a single type of operation that either shows or does not show this column. After discussing it with the team, we decided to display it only when necessary.

yordanalexandrov: So this could not spawn while you are on the page without doing anything in this view?

src/pages/monitor/backup-and-restore.html
yordanalexandrov: Ok, perhaps we should create an issue to update all other pages that show information which is constantly updated.

src/pages/monitor/backup-and-restore.html
boyan-tonchev: Yes.

yordanalexandrov (approved):

PR #1020, opened 2023-09-14
GDB-8753 Implemented ACL management page

svilenvelikov:
## What
Implemented an ACL management page.

## Why
By requirement.

## How
* Implemented the page, services, tests.
* Recreated the package-lock.json due to inconsistencies.

### DIFF
--- a/src/i18n/locale-fr.json
+++ b/src/i18n/locale-fr.json
@@ -127,6 +127,31 @@
             "content": "Sélectionner les nœuds à retirer du cluster existant"
         }
     },
+    "acl_management": {
+        "rulestable": {
+            "column": {
+                "index": "#",
+                "reorder": "",
+                "subject": "Sujet",
+                "predicate": "Prédicat",
+                "object": "Objet",
+                "context": "Contexte",
+                "role": "Rôle",
+                "policy": "Politique",
+                "actions": ""
+            },
+            "actions": {
+                "move_up": "Déplacer la règle vers le haut",
+                "move_down": "Déplacer la règle vers le bas",
+                "delete_rule": "Supprimer la règle",
+                "add_rule": "Ajouter une nouvelle règle",
+                "edit_rule": "Modifier la règle"
+            }
+        },
+        "errors": {
+            "loading_rules": "Erreur lors du chargement des règles ACL"
+        }
+    },
     "config.name.label": "Nom de configuration",
     "graph.config.required": "Ma configuration de graphe (obligatoire)",
     "description.text": "La description",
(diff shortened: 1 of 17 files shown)


src/i18n/locale-fr.json
```diff
@@ -119,6 +119,31 @@
             "content": "Sélectionner les nœuds à retirer du cluster existant"
         }
     },
+    "acl_management": {
+        "rulestable": {
+            "column": {
+                "index": "#",
+                "reorder": "",
```
yordanalexandrov, line 126: What is this for?

src/js/angular/aclmanagement/controllers.js
```diff
@@ -0,0 +1,122 @@
...
+    };
+
+    //
+    // Private functions
+    //
+
+    const loadRules = () => {
+        $scope.loading = true;
+        const repositoryId = getActiveRepository();
+        AclManagementRestService.getRules(repositoryId).success((response) => {
```
yordanalexandrov: Did we not talk at some point that it is better to use `.then` instead of `.success`?

Since they are depricated (not that we are going to update AngularJS, but still)

src/pages/aclmanagement.html
```diff
@@ -0,0 +1,87 @@
...
+                                gdb-tooltip="{{'acl_management.rulestable.actions.move_down' | translate}}">
+                            <i class="fa fa-caret-down"></i>
+                        </button>
+                    </td>
+                    <td>{{rule.subject}}</td>
+                    <td>{{rule.predicate}}</td>
+                    <td>{{rule.object}}</td>
+                    <td>{{rule.context}}</td>
+                    <td>{{rule.role}}</td>
+                    <td>{{rule.policy}}</td>
```
yordanalexandrov, line 72: Shouldn't this also be translated? Or you want all data to be shown as is regardless if it is an enumeration of some kind?

test-cypress/fixtures/locale-en.json
```diff
@@ -52,8 +52,8 @@
             "message_size_kb_tooltip": "Message size concerns the size of the data blocks transferred during data replication streaming through the RPC protocol.",
             "verification_timeout": "Verification timeout",
             "verification_timeout_tooltip": "Verification timeout concerns the amount of time a follower node would wait until attempting to verify last committed entry when the first verification is unsuccessful.",
-            "transaction_log_max_size_gb": "Transaction log maximum size",
-            "transaction_log_max_size_gb_tooltip": "Maximum size of the transaction log in GBs that will trigger an automatic truncation of the transaction log if it becomes bigger. The minimum transaction log size is 1 GB. Negative numbers will disable the automatic transaction log truncation."
+            "transaction_log_maximum_size_gb": "Transaction log maximum size GB",
```
yordanalexandrov, line 55: Shouldn't this be in a separate commit?

yordanalexandrov: Hm.. maybe a rebase on master is in order?

yordanalexandrov: This is just an update of the test fixture, so it is ok

src/pages/aclmanagement.html
svilenvelikov: Exactly. This is data and is displayed as is. Pavel insisted that we don't want to represent it with some component like a checkbox for example.

src/i18n/locale-fr.json
svilenvelikov: The label for the reorder actions column :)
Maybe I should remove it

src/js/angular/aclmanagement/controllers.js
svilenvelikov: Yes. My bad. I tested something and forgot to revert this.

yordanalexandrov (approved):

PR #1021, opened 2023-09-15
GDB-8280 Add missing extensions to import file selector

yordanalexandrov:
## What
Added support for additional file formats

## Why
To enhance the import feature's compatibility and usability, this commit introduces support for several new file formats, including 'xml', 'rdfs', 'ndjsonld', 'ndjson', and 'jsonl'.

## How
Updated the 'controllers.js' file to include the newly supported formats in the 'fileFormats' array, ensuring that the import feature can now handle these formats seamlessly. This expansion of supported formats increases the versatility of the import functionality.

### DIFF
--- a/src/js/angular/import/controllers.js
+++ b/src/js/angular/import/controllers.js
@@ -48,7 +48,8 @@ importCtrl.controller('CommonCtrl', ['$scope', '$http', 'toastr', '$interval', '
 
         $scope.getAppData();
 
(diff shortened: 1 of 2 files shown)


svilenvelikov:
It seems like here you should update the tests as well

yordanalexandrov:
> It seems like here you should update the tests as well

No need to change them for the new addition. But amended them to fix a problem that was bound to happen.

When the import button is hovered a popover with information is shown and this popover is supposed to hide after the button is pressed.

In the cypress step this is done forcefully with a trigger to `mouseout` event, which is not what the popover is listening for. Changed to `mouseleave`. Probably this change came with some update of the libraries and did not cause problem until now.

svilenvelikov (approved):

PR #1036, opened 2023-09-21
GDB-8819 fix sparql result download in json and jsonld format

svilenvelikov:
## What
Downloading the sparql result in json and jsonld was broken. This should be fixing the problem.

## Why
There is a branch in the code where the sparql construct queries are handled differently and the rest client for the download was invoked with the repository id, but the rest client expected repository info object which resulted in `undefined` as repository id value and failing download request.

## How
Fixed the rest client to expect simple repository id as a string. The only controller which uses it is sending the id only and not the repository object.


svilenvelikov:
Jenkins retry build

yordanalexandrov (approved):

PR #1038, opened 2023-09-25
GDB-8813 Refactor so token is fetched per request

yordanalexandrov:
## What
Refactor the application so that auth token is fetched on each request and not cached

## Why
So that if user logs out, he will not be able to continue working as the user in another tab/window

## How
- added an auth token service to persist and fetch the auth token
- refactored code to use this service


src/js/angular/core/services.js
```diff
@@ -286,3 +287,25 @@ function ClassInstanceDetailsService($http) {
         }
     }
 }
+
+function AuthTokenService() {
```
boyan-tonchev, line 291: It is a good idea to introduce an authentication service, but it would be better if you use the `LocalStorageAdapter` instead of `localStorage`. For example, you can see how the `WorkbenchSettingsStorageService` (workbench-settings-storage-service.js) is implemented.

src/js/angular/core/services.js
```diff
@@ -286,3 +287,25 @@ function ClassInstanceDetailsService($http) {
...
+    const authStorageName = 'com.ontotext.graphdb.auth';
```
boyan-tonchev, line 292: Add this constant as `LSKeys` option in `LocalStorageAdapter`.

src/js/angular/core/services/jwt-auth.service.js
```diff
@@ -278,30 +269,23 @@ angular.module('graphdb.framework.core.services.jwtauth', [
...
-            this.setAuthHeaders = function () {
-                return new Promise(resolve => {
-                    const auth = this.auth ? this.auth : undefined;
-                    $http.defaults.headers.common['Authorization'] = auth;
-                    // Angular doesn't send this header by default and we need it to detect XHR requests
+            this.setAuthHeaders = function() {
+                return new Promise((resolve) => {
+                    // Angular doesn't send this header by default, and we need it to detect XHR requests
                     // so that we don't advertise Basic auth with them.
                     $http.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest';
```
boyan-tonchev, line 276: I am not sure, but I think that this header can be removed too, because the header is added in `unauthorized.interceptor.js`.

src/js/angular/core/services.js
yordanalexandrov: Well, there is an authentication service, it is called `$jwtAuth`, but it uses `$http` and having this inside causes a cyclic dependency. This is why I extracted it as a separate service.

As far as using the `LocalStorageAdapter`, I was going to but it prefixes all keys and as this key in particular is used separate, I didn't want to change it, so I left it like that.

The key is used by other libraries when they make requests. `swagger-ui.js` is one of them and it does not have access to `LocalStorageAdapter` so if the prefix is changed in the workbench config, it would stop working.

src/js/angular/core/services/jwt-auth.service.js
yordanalexandrov: Yes, I removed it from here along with the method

boyan-tonchev (approved):

src/js/angular/core/services/jwt-auth.service.js
```diff
@@ -228,12 +221,9 @@ angular.module('graphdb.framework.core.services.jwtauth', [
...
             };
 
-            this.getAuthToken = function () {
-                return this.auth;
-            };
-
             this.authTokenIsType = function (type) {
-                return this.auth && this.auth.startsWith(type);
+                const token = AuthTokenService.getAuthToken();
+                return token && token.startsWith(type);
```
svilenvelikov, line 226: It'd have been better if we moved all such checks inside the service: `token.startsWith('GDB')`, `token.startsWith('Bearer')`, etc.

PR #1043, opened 2023-09-27
GDB-8644 Fix repository handling when login changes

yordanalexandrov:
## What
Handling of selected repository causes errors if logged user does not have read rights

## Why
The logic did not cover the case where a user would not have read rights to a preselected repository upon login

## How
- moved adding of authorization header and repository header to a separate interceptor
- added logic to remove selected repo if user does not have rights for it
- removed the caching of the repository and refactored, so it is always fetched from the local storage
- added keys to LSKey for repository id and location
- did some refactoring along the way, mainly changed success/error with then/catch for promises


src/js/angular/core/services/jwt-auth.service.js
```diff
@@ -280,6 +280,15 @@ angular.module('graphdb.framework.core.services.jwtauth', [
                     $rootScope.deniedPermissions = {};
                     this.securityInitialized = true;
 
+                    const selectedRepo = {
+                        id: LocalStorageAdapter.get(LSKeys.REPOSITORY_ID) || '',
+                        location: LocalStorageAdapter.get(LSKeys.REPOSITORY_LOCATION) || ''
+                    };
+
+                    if (!jwtAuth.canReadRepo(selectedRepo)) {
+                        LocalStorageAdapter.remove(LSKeys.REPOSITORY_ID);
```
boyan-tonchev, line 289: We probably need to remove the `REPOSITORY_LOCATION`.

src/js/angular/core/services/repositories.service.js
```diff
@@ -57,13 +52,17 @@ repositories.service('$repositories', ['$http', 'toastr', '$rootScope', '$timeou
         };
 
         this.resetActiveRepository = function () {
-            let repositoriesFromLocation = that.repositories.get(this.repository.location);
+            const repository = {
+                id: LocalStorageAdapter.get(LSKeys.REPOSITORY_ID) || '',
+                location: LocalStorageAdapter.get(LSKeys.REPOSITORY_LOCATION) || ''
+            };
```
boyan-tonchev, line 58: The repository creation object from local store can be extracted as a function and can be used everywhere

src/js/angular/core/services/jwt-auth.service.js
```diff
@@ -280,6 +280,15 @@ angular.module('graphdb.framework.core.services.jwtauth', [
...
+                    }
+
```
boyan-tonchev, line 291: We probable need to reset and `$rootScope.deniedPermissions`. Can you check this. You can see comment in `repositories.servcice.js setRepository` function.

boyan-tonchev (approved):

svilenvelikov (approved):

PR #1045, opened 2023-09-28
GDB-8757 Prevent some operations and leaving the page when ACL model is changed

svilenvelikov:
## What
* Prevent save/revert ACL when there are no changes in the ACL model.
* Prevent leaving the page when there are changes in the ACL model. Doing this will raise confirmation dialog for the user to choose to stay on the page or to leave and loose his data.

## Why
This is an UX improvement that makes the work in the ACL management page more friendly and error prone.

## How
Added a modelIsDirty flag that is set on each operation. The flag is used to control the actions visibility and to show confirmation dialog when the user tries to leave the page before saving the changes in the model.

### DIFF
--- a/src/js/angular/aclmanagement/controllers.js
+++ b/src/js/angular/aclmanagement/controllers.js
@@ -1,6 +1,7 @@
 import 'angular/rest/plugins.rest.service';
 import 'angular/rest/aclmanagement.rest.service';
 import {mapAclRulesResponse} from "../rest/mappers/aclmanagement-mapper";
+import {isEqual} from 'lodash';
 
 const modules = [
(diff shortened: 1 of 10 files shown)


src/js/angular/aclmanagement/controllers.js
```diff
@@ -200,4 +257,55 @@ function AclManagementCtrl($scope, toastr, AclManagementRestService, $repositori
     }, function () {
```
boyan-tonchev, line 257: `$scope.$watch` also returns a deregistration function, so it has to be added to the `subscriptions` array.

boyan-tonchev (approved):

PR #1051, opened 2023-10-03
GDB-8863, GDB-8864 Fixes authentication issues

boyan-tonchev:
## What
- When a user logs out in one tab, it doesn't log them out of all tabs. For example, if security is enabled, free access is disabled, and another tab is displaying the welcome page, that page will still be visible;
- When security is enabled, free access is disabled, and a user logs out, the main menu and status bar will remain visible in all other tabs;

## Why
- We log the user out and redirect them to the login page if a request returns a 401 (Unauthorized) status. For pages that do not execute requests after they are loaded, such as the welcome page, the page will remain visible;
- The condition for displaying the status bar and main menu is incorrect;

## How
- A local store change event listener has been added. When the event occurred for the 'authorization' value, the page is updated based on the new authorization status. Redirects the user to the login page if the new value is unauthorized. Reloads the last visited page if the user is authorized again.
- Fixes the condition check when status bar and main menu have to be displayed.


src/js/angular/controllers.js
```diff
@@ -461,12 +461,53 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repos
...
             // otherwise show login screen if security is on
             $rootScope.redirectToLogin();
         }
     };
 
+    $scope.showMainManuAndStatusBar = () => {
+        return $jwtAuth.isAuthenticated() || $scope.isSecurityEnabled() && $scope.isFreeAccessEnabled();
+    };
+
+    let isAuthenticated = undefined;
```
svilenvelikov, line 475: Would there be a problem if this is initialized with `false`?

src/js/angular/core/services/jwt-auth.service.js
```diff
@@ -25,6 +25,12 @@ angular.module('graphdb.framework.core.services.jwtauth', [
                 return !$rootScope.deniedPermissions[path];
             };
 
+            $rootScope.updateReturnUrl = () => {
```
svilenvelikov, line 28: Why this is in the $rootScope?

svilenvelikov (changes requested):

src/js/angular/controllers.js
boyan-tonchev: In the initial implementation, it had to be undefined, but now it can be false. I will verify and set it.

boyan-tonchev: Actually, it has to be undefined, I wrote an explanatory comment to describe the initialization of the variable.

src/js/angular/core/services/jwt-auth.service.js
boyan-tonchev: Done

PR #1066, opened 2023-10-10
GDB-8914 Fix OpenId authentication process

yordanalexandrov:
## What
Fix invalid client authentication scheme error when trying to log in with OpenID

## Why
When authenticating with OpenId, the calls that are made during the auth process were amended by the authentication interceptor, which caused failure of getting tokens. Also the OpenId access token, was not set in the AuthTokenService

## How
- moved the OpenId configuration to AuthTokenService
- skipped headers change in authentication interceptor for OpenId requests
- set OpenId access token in AuthTokenService

### DIFF
--- a/src/js/angular/core/interceptors/authentication.interceptor.js
+++ b/src/js/angular/core/interceptors/authentication.interceptor.js
@@ -8,6 +8,14 @@ angular.module('graphdb.framework.core.interceptors.authentication', [
             return {
                 'request': function(config) {
                     const headers = config.headers || {};
+
+                    // When using OpenID, during authentication process, the additional headers modification must be skipped
+                    const openIDUrls = [AuthTokenService.OPENID_CONFIG.openIdKeysUri, AuthTokenService.OPENID_CONFIG.openIdTokenUrl];
+                    const isOpenIdUrl = openIDUrls.some((url) => config.url.indexOf(url) > -1);
+                    if (isOpenIdUrl) {
+                        return config;
+                    }
+
                     // Angular doesn't send this header by default, and we need it to detect XHR requests
                     // so that we don't advertise Basic auth with them.
                     headers['X-Requested-With'] = 'XMLHttpRequest';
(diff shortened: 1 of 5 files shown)


src/js/angular/core/interceptors/authentication.interceptor.js
```diff
@@ -8,6 +8,14 @@ angular.module('graphdb.framework.core.interceptors.authentication', [
             return {
                 'request': function(config) {
                     const headers = config.headers || {};
+
+                    // When using OpenID, during authentication process, the additional headers modification must be skipped
```
boyan-tonchev, line 12: Do repository headers have to be skipped?

src/js/angular/core/services.js
```diff
@@ -290,12 +290,14 @@ function ClassInstanceDetailsService($http) {
 
 function AuthTokenService() {
     const authStorageName = 'com.ontotext.graphdb.auth';
+    const OPENID_CONFIG = {};
```
boyan-tonchev, line 293: All open ID configurations will be in memory. I think this will break some logout flows when the user logged out from a second tab.

src/js/angular/core/interceptors/authentication.interceptor.js
yordanalexandrov: Yes, those requests are to the Auth service and not GDB, so the repo information is not needed.

src/js/angular/core/services.js
yordanalexandrov: They were in momory before also. They are fetched on security init.
I just moved them so that they are accessable from other places (`authentication.interceptor`)

svilenvelikov (approved):

src/js/angular/core/services.js
yordanalexandrov: There was a problem and I have done a fix

boyan-tonchev (approved):

PR #1070, opened 2023-10-16
GDB-8891 disable the ask gpt button while previous question is in flight

svilenvelikov:
## What
Disable the ask gpt button while previous question is in flight.

## Why
If the button is not disabled and another question is asked while the previous one is not already answered, then the answers in the pane are mismatched.

## How
Disabled the button during the request is in progress using the `loading` flag.


svilenvelikov:
Jenkins retry build

PR #1071, opened 2023-10-16
GDB-8687: The "isLiteral field" of the similarity index is not properly set

boyan-tonchev:
## What
When I create similarity indexes from existing one, the index type of result is not same as source index.

## Why
There was an initialization mistake in setting the "options" field of the SimilarityIndex object when tha view is loading. The field was being set on the SimilarityIndexInfo object instead of the SimilarityIndex.

## How
Wrong setting of "options" field has been corrected.


svilenvelikov (approved):

boyan-tonchev:
Jenkins retry build

PR #1073, opened 2023-10-16
GDB-8711 Allowing users to copy paste roles with space between them, instead of replacing with a dash

Emskiqq:
Add parameters to the tags-input to enable users to paste multiple roles with space between them


svilenvelikov (approved):

yordanalexandrov:
Jenkins retry build

PR #1076, opened 2023-10-17
GDB-8686 Disable the save sparql template button if the query is not changed

svilenvelikov:
## What
Disable the save sparql template button if the query is not changed.

## Why
It should be possible to perform sparql template saving only if there is a change in the existing query in the editor. Otherwise the backend throws error. It's not clear why this error happens but it is what it is for now.

## How
Extended the ontotext-yasgui facade component to do dirty checking for the query value and emit a `queryChanged` event. The event is then handled in the sparql template edit controller where the `isDirty` flag is toggled. Also the save button is conditionally disabled or enabled according to the `isDirty` flag value.

### DIFF
--- a/src/js/angular/core/directives/yasgui-component/yasgui-component.directive.js
+++ b/src/js/angular/core/directives/yasgui-component/yasgui-component.directive.js
@@ -8,7 +8,6 @@ import {QueryMode} from "../../../models/ontotext-yasgui/query-mode";
 import {YasrPluginName} from "../../../models/ontotext-yasgui/yasr-plugin-name";
 import {isEqual} from "lodash/lang";
 import {QueryType} from "../../../models/ontotext-yasgui/query-type";
-import {saveAs} from 'lib/FileSaver-patch';
 import {YasguiComponent} from "../../../models/yasgui-component";
 
 const modules = [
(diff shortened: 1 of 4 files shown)


src/js/angular/core/directives/yasgui-component/yasgui-component.directive.js
```diff
@@ -320,6 +337,8 @@ function yasguiComponentDirective(
 
                     $scope.ontotextYasguiConfig = config;
 
+                    setInitialQueryState();
```
boyan-tonchev, line 340: Would be better if the `setInitialQueryState` returns a promise and when it is resolved then the `afterInit` to be called.

boyan-tonchev (approved):

src/js/angular/core/directives/yasgui-component/yasgui-component.directive.js
svilenvelikov: yeah sure, I thought about it, but left it as is. Will think about it again and eventually change it.

svilenvelikov: fixed

svilenvelikov:
Jenkins retry build

PR #1098, opened 2023-10-20
Gdb 8267_GDB-8286

boyan-tonchev:
## What
- GDB-8267: When repository is virtual. the inferred and sameAs buttons have to be active and should not be toggleable.
- GDB-8286: A new configuration property "language" has been introduced.

## Why
- GDB-8267: Missed requirement during development;
-  GDB-8286: There was no way to set the language when the component is created. The default language was English, so if the client wanted to open the component with French, for instance, they had to open it with English and then refresh it as set language on French. This flow caused component to be redrawn twice.

## How
- GDB-8267,  GDB-8286:  The version of "ontotext-yasgui-web-component" has been updated.
- GDB-8267: Added styling when infer and sameAs are disabled.

Additional work:
- Added ontotext-yasgui-web-component configuration model.

### DIFF
--- a/src/js/angular/models/ontotext-yasgui/yasgui-config.js
+++ b/src/js/angular/models/ontotext-yasgui/yasgui-config.js
@@ -1,0 +1,357 @@
+import {RenderingMode} from "./rendering-mode";
+import {YasguiOrientation} from "./yasgui-orientation";
+import {YasguiQueryHttpMethod} from "./yasgui-query-http-method";
+import {YasqeMode} from "./yasqe-mode";
+import {YasrPluginName} from "./yasr-plugin-name";
+
+/**
+ * Holds all configurations related with ontotext-yasgui-web-component.
+ */
+export class YasguiConfig {
+    constructor() {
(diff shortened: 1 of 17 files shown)


src/js/angular/models/ontotext-yasgui/yasgui-config.js
```diff
@@ -0,0 +1,357 @@
+import {RenderingMode} from "./rendering-mode";
+import {YasguiOrientation} from "./yasgui-orientation";
+import {YasguiQueryHttpMethod} from "./yasgui-query-http-method";
+import {YasqeMode} from "./yasqe-mode";
+import {YasrPluginName} from "./yasr-plugin-name";
+
+/**
+ * Holds all configurations related with ontotext-yasgui-web-component.
+ */
+export class YasguiConfig {
```
svilenvelikov, line 10: Cool. Maybe just rename it to OntotextYasguiConfig

svilenvelikov (approved):

PR #1100, opened 2023-10-20
GDB-9013: Link underlining issues in Workbench and the documentation

avataar:
In 10.4 we introduced underlining of links:
Dotted for button links in Workbench and internal links in the documentation
Solid for external links in both (was already there)

It looks like the default auto thickness and offset of the underline are a bit of a fringe value. Safari does wonderfully, Firefox has too thin lines on smaller font sizes, Chrome is a bit of an oddball here, especially with the dotted links – too thick and too close to the text. See attached screenshots.
We aim to reproduce a behaviour that is as close as possible to the default auto algorithm in Safari/Firefox on all browsers.
Specifying this seems to do the trick for the most part in a font-size neutral way:
```css
text-decoration-thickness: clamp(1px, 0.06em, 10px);
text-underline-offset: clamp(1px, 0.06em, 10px);
```
While this works well for Workbench and the documentation, Chrome is again the oddball by producing a different (smaller) offset in Workbench when compared to the documentation with exactly the same styling.
Thus, for Workbench it’s this incantation:
```css
text-decoration-thickness: clamp(1px, 6%, 10px);
text-underline-offset: clamp(1px, 6%, 10px);
```
It produces identical output in both Workbench and the documentation in all 3 browsers.
That 0.06em vs 6% is a really weird issue as 1) they should do exactly the same, 2) at font sizes 14 and 16 the clamp should return 1px anyway. Putting 6% on the documentation in Chrome makes the offset larger than the other browsers.

A secondary issue just with Workbench: when a button has an icon and a link, the underlining spills out between the icon and the link. As far as I can tell there are only 2 such buttons, the Create repository shown when no repo is selected and the Help button in the Import view.

Also added two missing icon classes that are used in the documentation (to keep the theme file synced).

### DIFF
--- a/src/css/bootstrap-graphdb-theme.css
+++ b/src/css/bootstrap-graphdb-theme.css
@@ -15,6 +15,7 @@
     --menu-width: var(--menu-width-expanded);
     /* max usable content width not considering the menu and margins */
     --max-width-minus-menu: calc(100% - var(--menu-width));
+    /* Link decoration, see GDB-9013 before you touch these */
     --link-decoration-thickness: clamp(1px, 6%, 10px);
     --link-underline-offset: clamp(1px, 6%, 10px);
 }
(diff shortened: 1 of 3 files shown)


avataar:
Jenkins retry build

svilenvelikov (approved):

PR #1104, opened 2023-10-23
GDB-8989 Change the line numbers color in yasqe

svilenvelikov:
## What
Change the line numbers color in yasqe.

## Why
Try to match the existing workbench look and feel as much as possible,

## How
Remove the overriding of the yasqe line numbers colors.


svilenvelikov:
Jenkins retry build

PR #1114, opened 2023-11-01
GDB-9019 Loading a non-existent predicate for filtering in the RDF Rank seems to break the Workbench view

MartinaDimova:
What

Users should be able to add and delete IRIs in following ways without receiving an error: 
* with shortened predicate when it exists as a namespace, with or without <>. For example: rdfs:label.
* like absolute IRI, with or without <>. For example: http://www.w3.org/2000/01/rdf-schema#label.
* like an irregular syntax IRI with <>, containing “:”. For example: <mex:type>.

Why

RDF Rank and Autocomplete views to accept and display IRIs consistently. Some short IRI's are not expanded, but percived as text even it is a namespace, for others we receive an error when trying to remove it, the interface does not make difference between shortened and a simple text IRI.

How

* Made a check for a newly added namespace when refreshing filtering configs.
* Allowed adding of an absolute IRI, with or without <>.
* Allowed adding of a shortened IRI, with or without <>.
* Made shortened IRIs displyed as absolute.
* Check for existing namespace and throwng error of IRI is not bracketed with <>.


MartinaDimova:
Jenkins retry build

svilenvelikov:
Please follow the MR description format we use in all web projects. See for example https://github.com/Ontotext-AD/graphdb-workbench/pull/1111

src/js/angular/autocomplete/controllers.js
```diff
@@ -257,6 +263,20 @@ function AutocompleteCtrl($scope, $interval, toastr, $repositories, $licenseServ
         pullStatus();
     });
 
+    const initNamespaces = function () {
+        RDF4JRepositoriesRestService.getNamespaces($repositories.getActiveRepository())
+            .success(function (data) {
+                const nss = _.map(data.results.bindings, function (o) {
```
svilenvelikov, line 269: I'd use the namespace mapper utility  for this src/js/angular/rest/mappers/namespaces-mapper.js utility

MartinaDimova: Will do also in the RDFRank app.js and push.

svilenvelikov (approved):

PR #1128, opened 2023-11-09
GDB-9077 fix download sparql results operation in the sparql view

svilenvelikov:
## What
Fix download sparql results operation in the sparql view.

## Why
The `download as` operation seems to be broken for a while because we removed the form used for submitting the parameters needed for the export as well as there was another issue where the sparql endpoint was impossible to be resolved the way it was implemented because the `endpoint` configuration can be a string or a get function which is expected to be evaluated in the context of the YASGUI component and not in the workbench.

## How
* Re-introduced the download as parameters form again.
* Refactored the obtaining of the sparql endpoint based on the query mode so that it can be used correctly in the context of the workbench.


boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

PR #1132, opened 2023-11-10
GDB-9118 hide all yasr plugin tabs if configured via showResultTabs

svilenvelikov:
## What
When `showResultTabs` flag for the yasgui component is set to `false` then all yasr plugins should be hidden.

## Why
There are some integration points like the explore (resource) view where only the yasr results table renderer should be displayed and no tabs should be visible. In the latest version of the yasgu component were introduced two new yasr plugins: charts and pivot table. These two plugins were not registered in the service which used to hide them when the flag is falsy.

## How
* The component was updated and the new version is integrated in the workbench
* Also enabled tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
                 "ng-file-upload": "^12.2.13",
                 "ng-tags-input": "^3.2.0",
                 "oclazyload": "^1.1.0",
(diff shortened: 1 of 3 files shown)


boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

PR #1135, opened 2023-11-13
GDB-8994 reposition yasqe action buttons

svilenvelikov:
## What
Reposition yasqe action buttons to prevent overlapping with the vertical scrollbar when it's present in the yasqe.

## Why
Yasqe buttons were placed too close to the right-side border of the editor which led to overlapping between the vertical scrollbar and the run query button.

## How
Increased the space between the buttons and the right-side editor border


svilenvelikov:
Jenkins retry build

PR #1136, opened 2023-11-13
GDB-8996 improve styling of yasr plugin tabs

svilenvelikov:
## What
Improve styling of yasr plugin tabs.

## Why
Make it consistent with the old yasgui.

## How
Put some explicit margin-right on tabs to make it similar to old yasgui


svilenvelikov:
Jenkins retry build

PR #1145, opened 2023-11-21
GDB-9194: Typo in the alert message for accessible repositories in Workbench when using french

DanchoIvanov:
## What
Now workbench displays correct alert message for accessible repositories in French.

## Why
Workbench had a typo in the alert message for accessible repositories in French.

## How
* Changed the template rendering the alert message because the word order of the sentences in both English and French is different.
* Changed the labels in both locale files.


yordanalexandrov:
Please follow the convention for describing issues in the Workbench.
You can see [here](https://github.com/Ontotext-AD/graphdb-workbench/pull/1144) as an example.

src/i18n/locale-fr.json
```diff
@@ -928,8 +928,8 @@
     "core.errors.select.repository.warning.msg": "Cliquez sur l'un des dépôts ci-dessous pour vous y connecter",
     "core.errors.or.create.repository.warning.msg": "ou créez un nouveau dépôt",
     "core.errors.no.accessible.warning.msg": "Il n'y a pas de dépôt accessible",
-    "core.errors.writable": "accessible en écriture",
-    "core.errors.repositories": "accessible en écriture",
+    "core.errors.writable": "",
```
yordanalexandrov, line 931: This will not work like this.

As can be seen in the template this is a compound label that in english looks like this:
```
"There are no accessible" + "writable" + "repositories"
```
Where the `writable` part is optional if there is a repository but it is restricted.

This can not be translated directly in french with.
The labels need to be merged into two. One without `writable` and one with.
The template would then need to be modified with an if condition and showing the apropriate label.

DanchoIvanov: image

DanchoIvanov: We send the same message so you don't need to add additional word in case the repository is not write accessible

yordanalexandrov: Having an optional part `writable` makes the sentences two, which show under different conditions.

```
There are no accessible repositories
```
and
```
There are no accessible writable repositories
```

DanchoIvanov: Changed the template and labels as suggested.

src/js/angular/core/templates/core-errors.html
```diff
@@ -19,8 +19,10 @@
                         </span>
 
                         <span ng-hide="getAccessibleRepositories().length">
-                            {{'core.errors.no.accessible.warning.msg' | translate}}<span ng-if="isRestricted">{{'core.errors.writable' | translate}}
-                        </span>{{'core.errors.repositories' | translate}}<span ng-if="canManageRepositories()">{{'core.errors.create.repository.warning.msg' | translate}}</span>.
+                            {{(isRestricted
```
yordanalexandrov, line 22: This will evaluate on each digest cycle.
It is far better to use a span with an ng-if on it with `isRestricted` as a variable as it will evaluate only on variable change.

DanchoIvanov: Fixed

boyan-tonchev (approved):

yordanalexandrov (approved):

PR #1146, opened 2023-11-22
GDB-9197 Fix issues with charts

yordanalexandrov:
## What
Fix issues with charts when no series are selected

## Why
Causes the axis labels to show strange things

## How
- added an if condition to not update chart if all series are disabled


svilenvelikov (approved):

yordanalexandrov:
Jenkins retry build

PR #1154, opened 2023-11-27
GDB-9164 and GDB-9159 fix save and cancel buttons position and order

svilenvelikov:
## What
Fix save and cancel buttons position and order broken during a revert some time ago.

## Why
The order and position of these buttons are expected to be this way in all views and dialogs.

## How
Just reordered the buttons in the template.


boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

PR #1155, opened 2023-11-27
GDB-9167: Different color on info icon in a SPARQL Query & Update

boyan-tonchev:
## What
- Different color on info icon in a SPARQL Query & Update;
- There is a randomly failing test 'Should prevent leaving with confirmation when expand query is changed';
- The test "should abort query when click on "Abort query" button." frequently fails.

## Why
- The theme changes has been added after implementation of sparql editor page and the theme color is not set to it;
- There is a NPE - "cons:error ✘ TypeError: Cannot read properties of undefined (reading 'getPropertyDescription')" when try to retrieve sample name of a query;
- When the "Abort query" button is clicked, the value of the button and the tooltip are changed. However, the time between clicking and disappearance is too short, leading to the test failure with "AssertionError: Timed out retrying after 15000ms: Expected to find element: div[data-tippy-root], but never found it.".

## How
- added missed style class to the info icon;
- The checks if sample query exist before fetching the name has been added;
- Remove those checks because they are not important for the aborting query behaviour.

### DIFF
--- a/src/js/angular/graphexplore/controllers/graphs-config.controller.js
+++ b/src/js/angular/graphexplore/controllers/graphs-config.controller.js
@@ -173,6 +173,9 @@ function GraphConfigCtrl(
      * @return {string|*}
      */
     $scope.getSampleName = (sample, property) => {
+        if (!sample) {
+            return '';
+        }
         const propertyDescription = sample.getPropertyDescription(property);
         if (propertyDescription) {
             // Sample has description, use it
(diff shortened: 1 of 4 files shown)


svilenvelikov (approved):

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

PR #1157, opened 2023-11-28
Gdb 9168 wrong styling of explain plan

boyan-tonchev:
## What
- The font of the explain plan query differs from the old yasr;
- The error plugin tests frequently failed.

## Why
- The "font-family" property was not set;
- We don't know exactly why. But the test fail on verification if full error message displayed.

## How
- The "font-family" has been set to the same value as in the old yasr;
- The error text verification has been changed.

Additional work:
Updates the ontotext-yasgui-web-component version because there is a fix for the explain plan query. The result looks like plain text.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
         "ng-file-upload": "^12.2.13",
         "ng-tags-input": "^3.2.0",
         "oclazyload": "^1.1.0",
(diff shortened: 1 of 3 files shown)


boyan-tonchev:
Jenkins retry build

PR #1162, opened 2023-11-29
GDB-9232 improve replace cluster nodes dialog

svilenvelikov:
## What
Improve replace cluster nodes dialog.

## Why
The replace cluster nodes dialog needs some improvements for better UX like changing the cluster nodes list label and the selected nodes for adding label. Also changing the condition which is used to disable the OK button. Now it is allowed to only select nodes to be removed or added and not both.

## How
* Changed the labels of the two lists.
* Changed the condition that disables the OK button


yordanalexandrov (approved):

yordanalexandrov (approved):

svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

PR #1167, opened 2023-12-05
The visual button has different styling than old implementation

boyan-tonchev:
## What
The visual button has small space between icon/button-label and corner of button.

## Why
There was set different styling than old implementation.

## How
Added padding on left and right side of button.


svilenvelikov (approved):

boyan-tonchev:
Jenkins retry build

PR #1173, opened 2023-12-12
GDB-9275: CLuster view cut out "Building a snapshot for" text

boyan-tonchev:
## What
Sometimes, messages describing the recovery status of a node appear incomplete.

## Why
The messages are dynamically generated, and if the backend (BE) does not return an affected node (for some reason), the messages look incomplete.

## How
Additional checks have been added. When rendering a message, if affected nodes are sent, then the full message "Building a snapshot for {{node}}" is used. If not, a shorter version "Building a snapshot" will be used. This check will be added to all messages that may encounter this scenario.


src/js/angular/clustermanagement/directives.js
```diff
@@ -338,7 +339,33 @@ clusterManagementDirectives.directive('clusterGraphicalView', ['$window', 'Local
                     nodes.forEach((node) => {
                         node.hostname = UriUtils.shortenIri(node.endpoint);
                         if (!_.isEmpty(node.recoveryStatus)) {
-                            const messageLabelKey = 'cluster_management.cluster_graphical_view.recovery_state.' + node.recoveryStatus.state.toLowerCase();
+                            let messageLabelKey = 'cluster_management.cluster_graphical_view.recovery_state.';
+                            const hasAffectedNodes = node.recoveryStatus.affectedNodes && node.recoveryStatus.affectedNodes.length > 0;
+                            const nodeRecoveryState = node.recoveryStatus.state;
+
+                            if (hasAffectedNodes) {
+                                switch (nodeRecoveryState) {
```
svilenvelikov, line 347: Can you please replace this switch with the  more simple! Reorganize the label keys so that they allow simple resolution by concatenating `messageLabelKey + state.toLoverCase` without any additional conditional expressions.

boyan-tonchev: Done

svilenvelikov (approved):

PR #1175, opened 2023-12-13
GDB-9108 removed node replacement operation for GDB10.5 version

svilenvelikov:
## What
Removed node replacement operation for GDB10.5 version.

## Why
Removed it due to instability of the cluster in the backend after introducing the feature.

## How
Hide the button that opens the replace node dialog and skipped the tests.


boyan-tonchev:
May be is better to remove the code instead comment it, because we don't sure if the functionality will be released in 10.5

svilenvelikov:
> May be is better to remove the code instead comment it, because we don't sure if the functionality will be released in 10.5

@boyan-tonchev It won't be released in 10.5 for sure. But I don't get it what throubles you?

PR #1183, opened 2024-01-08
GDB-9211: Add Pop-Up Menus for JSON-LD Export Selection

VPetrova00:
### **What**
New JSONLD export settings view

### **Why**
In order to export repository in JSONLD format with additional settings.

### **How**
Created a modal for new settings.
Created a new controller for additional settings view.
Created an export service.
Added translations to locale-en.json.


VPetrova00:
What remains to be done:
- a cypress test
- translations in French
- fix an errors with the http get request

VPetrova00:
> What remains to be done:
> 
> * a cypress test
> * translations in French
> * fix an errors with the http get request

Everything is addressed.

src/js/angular/import/templates/exportSettingsModal.html
```diff
@@ -0,0 +1,57 @@
...
+                </label>
+                <div class="col-lg-9">
+                    <select class="form-control" id="wb-JSONLD-mode"
+                            ng-model="currentMode"
+                            ng-options="mode.link for mode in JSONLDModes">
+                    </select>
+                </div>
+            </div>
+
+            <div class="form-group row" ng-if="currentMode.name == 'framed' || currentMode.name == 'frame'">
```
svilenvelikov, line 25: You have already defined the JSONLDModesNames model which should be the source of truth and yet here you hardcode these types. Why? And something more. comparison in javascript must be done always by `===` or `!==`. There should be a really special case in order to use `==`.
So you can extract the names in a enum like object like this
``` javascript
$scope.JSONLDModesNames = $scope.JSONLDModes.reduce(function (acc, cur) {
   acc[cur.name] = cur.name;
   return acc;
}, {});
```
and then check against it here
```
 ng-if="currentMode.name === JSONLDModesNames.framed || currentMode.name == JSONLDModesNames.frame"
```
It'd be even better if we can omit these ^ checks here by having two lists with the repective types like 
```
$scope.JSONLDFramedModes = [JSONLDModesNames.framed, JSONLDModesNames.frame];
$scope.JSONLDContextModes = [JSONLDModesNames.context, JSONLDModesNames.compacted, JSONLDModesNames.flattened];
```
and then here you can just do
```
ng-if="JSONLDFramedModes.indexOf(currentMode.name) > -1"
```

src/js/angular/rest/export.rest.service.js
```diff
@@ -0,0 +1,50 @@
...
+        let url;
+        if (forSelectedGraphs) {
+            url = `${REPOSITORIES_ENDPOINT}/${repo.id}/statements?infer=false&${context}`;
+        } else {
+            if (context) {
+                url = `${REPOSITORIES_ENDPOINT}/${repo.id}/statements?infer=false&context=${graphsByValue[context.value].exportUri}&location=${encodeURIComponent(repo.location)}`;
+            } else {
+                url = `${REPOSITORIES_ENDPOINT}/${repo.id}/statements?infer=false&location=${encodeURIComponent(repo.location)}`;
+            }
+        }
```
svilenvelikov, line 23: There is quite a lot of repetition here. Can you try to simplify this by extracting the url base which is repeated in all branches and then just add on top of it the other parameters.

src/js/angular/rest/export.rest.service.js
```diff
@@ -0,0 +1,50 @@
...
+
+        return $http({
+            url: url,
+            method: 'GET',
+            headers: {
+                'Accept': acceptHeader,
+                'Link': linkHeader
+            },
+            responseType: "blob"
+        }, {timeout: 100}).then(function (res) {
```
svilenvelikov, line 41: Why this timeout was needed?

src/js/angular/rest/export.rest.service.js
```diff
@@ -0,0 +1,50 @@
...
+
+ExportRestService.$inject = ['$http', '$repositories', '$translate'];
+
+const REPOSITORIES_ENDPOINT = 'repositories';
+function ExportRestService($http, $repositories, $translate) {
+    return {
+        getExportedStatementsAsJSONLD
+    };
+
+    function getExportedStatementsAsJSONLD(context, repo, graphsByValue, acceptHeader, linkHeader, auth, forSelectedGraphs) {
```
svilenvelikov, line 13: Document and reorder these parameters a bit. It'd have been better if they were fewer but still.

svilenvelikov (changes requested):

VPetrova00:
Jenkins retry build

VPetrova00:
Jenkins retry build

svilenvelikov (approved):

PR #1191, opened 2024-01-18
GDB-8998 make line numbers color in yasgui consistent

svilenvelikov:
## What
Make line numbers color in yasgui consistent.

## Why
Line numbers in yasqe and yasr should have same color as well as the color needs to have better contrast.

## How
Add explicit styling of the line numbers and used the proposed rdgb color


boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

PR #1205, opened 2024-01-22
feat(GDB-9098): Edit all kinds of ACLs in Workbench

teodossidossev:
## WHAT:
- Updated the Workbench UI to support comprehensive ACL management across various scopes: Statement, Clear Graph, Plugin, and System.
- Enhanced the UI to accommodate different operation types (read, write) and enable users to specify attributes associated with each ACL scope, such as subjects, predicates, objects, contexts, plugins, roles, and policies.
- Added new tabbed interface and corresponding table structures in the ACL management section to distinctly represent and manage ACLs of different scopes.
- Implemented changes in the model, controller, view, and localizations to support this extended functionality.
- Implemented scope dirty handling with indicator in the tab title.

## WHY:
- To ensure that users can easily create, edit, and delete ACLs, and clearly understand the scope and impact of their access control settings.

## HOW:
- Refactored ACListModel in aclmanagement/model.js to use a Map for handling ACL rules by scope, and introduced specific models for different types of ACL rules.
- Updated aclmanagement/controllers.js to include functions for managing ACLs in different scopes and for tab navigation.
- Enhanced aclmanagement.html with a new tabbed layout and dynamic table structures to display and edit ACLs based on the selected scope.
- Adjusted CSS in aclmanagement.css to support the new layout and ensure a consistent and user-friendly interface.
- Adapted the REST mapper in aclmanagement-mapper.js to correctly process and map the new ACL rule structure.
- Modified JSON files to include new translations for tabs and columns, facilitating clearer communication and localization.


svilenvelikov:
image

src/js/angular/aclmanagement/controllers.js
```diff
@@ -212,6 +251,20 @@ function AclManagementCtrl($scope, $location, toastr, AclManagementRestService,
...
+     * Function to switch between tabs in the user interface.
+     *
+     * @param {string} scope
+     */
+    $scope.switchTab = (scope) => {
+        if ($scope.editedRuleIndex !== undefined) {
+            $scope.cancelEditing($scope.activeTabScope, $scope.editedRuleIndex)
+        }
+        $scope.activeTabScope = scope;
+        $scope.ruleKeys = Object.keys($scope.rulesModel.filterByScope($scope.activeTabScope)[0] || {});
```
svilenvelikov, line 265: Can you move the obtaining of the keys for the rule of given type in a method in the model too?

src/js/angular/aclmanagement/model.js
```diff
@@ -101,25 +162,77 @@ export class ACListModel {
     set aclRules(value) {
         this._aclRules = value;
     }
+
+    filterByScope(scope) {
```
svilenvelikov, line 166: This `filterByScope` is kind of OK as a name given what it does, although you are technically not filtering but just getting some value from the map by key, but in the places where you use it, it sounds confusing. The missing jsdoc contributes to this confusion too.

src/js/angular/aclmanagement/model.js
```diff
@@ -4,94 +4,155 @@ export class ACListModel {
      * @param {ACRuleModel[]} aclRules
      */
     constructor(aclRules = []) {
-        this._aclRules = aclRules;
+        this._aclRules = new Map();
```
svilenvelikov, line 7: The missing types for the key and the value for this Map leads to weird type resolutions in other places where you use it.

src/js/angular/aclmanagement/model.js
```diff
@@ -4,94 +4,155 @@ export class ACListModel {
...
-     * @param {string} role
-     * @param {string} policy
+     * @param {string} plugin
      */
-    appendNewRule(subject, predicate, object, context, role, policy) {
-        this.aclRules.push(new ACRuleModel(subject, predicate, object, context, role, policy));
+    appendNewRule(scope, policy, role, operation, subject, predicate, object, context, plugin) {
+        const rule = ACRuleFactory.createRule(scope, policy, role, operation, subject, predicate, object, context, plugin);
+        if (!this._aclRules.has(scope)) {
+            this._aclRules.set(scope, []);
```
svilenvelikov, line 37: Now this tends to propagate in multiple places which I don't like. I mean setting the default [] value and also getting the rules for given scope and defaulting to []. As far as the model is now a Map which is not typed, this leads to confusion and potential mistakes. That's why I tend to avoid using Map in plain js, but rather some custom models with methods.

src/js/angular/aclmanagement/model.js
```diff
@@ -4,94 +4,155 @@ export class ACListModel {
...
+        }
     }
 
-    getRule(index) {
-        return this.aclRules[index];
+    /**
+     * Returns the rule at the specified index for the given scope.
+     * @param {string} scope - The scope for which to retrieve the rule.
+     * @param {number} index - The index of the rule to retrieve.
+     * @returns {object|null} - The rule at the specified index for the given scope, or null if no rule exists.
```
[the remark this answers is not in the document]
svilenvelikov: Also, why `object` and not `ACRuleModel` maybe. Can you please have a look and add proper types.

src/js/angular/aclmanagement/model.js
```diff
@@ -4,94 +4,155 @@ export class ACListModel {
...
-            if (indexRuleToBeChecked !== index) {
+    isRuleDuplicated(scope, indexRuleToBeChecked) {
+        const checkedRule = this.getRule(scope, indexRuleToBeChecked);
+        if (!checkedRule) {
+            return false;
+        }
+
+        return this._aclRules.get(scope).some((rule, index) => {
+            if (index !== indexRuleToBeChecked) {
                 return angular.equals(checkedRule, rule);
```
svilenvelikov, line 108: Get rid if this `angular.equals`. Either use lodash or something custom

src/js/angular/aclmanagement/model.js
```diff
@@ -154,24 +267,187 @@ export class ACRuleModel {
...
+     * @param {string} operation - The operation associated with the rule.
+     * @param {string} subject - The subject associated with the rule.
+     * @param {string} predicate - The predicate associated with the rule.
+     * @param {string} object - The object associated with the rule.
+     * @param {string} context - The context associated with the rule.
+     * @param {string} plugin - The plugin associated with the rule.
+     * @returns {ACRuleModel} - The created access control rule.
+     * @throws {Error} - Throws an error if an invalid scope is provided.
+     */
+    static createRule(scope, policy, role, operation, subject, predicate, object, context, plugin) {
```
svilenvelikov, line 422: As far as I can see, all prams except scope are optional.

src/js/angular/aclmanagement/model.js
```diff
@@ -16,7 +16,7 @@ export class ACListModel {
 
 
     size(scope) {
-        return this._aclRules.get(scope).length;
+        return this._aclRules.get(scope) && this._aclRules.get(scope).length;
```
svilenvelikov, line 19: And what value type will this return if there are no rule for that scope?

src/js/angular/rest/mappers/aclmanagement-mapper.js
```diff
@@ -7,24 +7,28 @@ import {ACListModel} from "../../aclmanagement/model";
...
-                rule.policy,
-                rule.role,
-                rule.operation,
-                rule.subject,
-                rule.predicate,
-                rule.object,
-                rule.context,
-                rule.plugin
-            );
+            if (rule.scope) {
```
svilenvelikov, line 19: I'm not sure when this can actually happen, but isn't it better if we report such cases somehow?

src/pages/aclmanagement.html
```diff
@@ -14,155 +14,552 @@ <h1 id="acl-management-view-title">
...
-                        <em ng-if="!ruleData.$valid" class="icon-save icon-lg save-rule-disabled-btn"
-                            gdb-tooltip="{{'acl_management.rulestable.messages.invalid_form' | translate}}"></em>
-                    </td>
-                </tr>
-                <tr ng-repeat-end ng-if="0"></tr>
-                </tbody>
-            </table>
-        </form>
-
+        <div class="acl-management-container">
```
svilenvelikov, line 17: I wonder if it isn't about time to split this to smaller components/directives/

svilenvelikov:
There is no validation of the IRI fields and when I try to save the model, backend returns error 
image

svilenvelikov:
The role field has no default `CUSTOM_` prefix populated in the field in all context tabs, but only for statements.

svilenvelikov:
The column widths and alignment are inconsistent on separate tabs. This should be communicated with Donka maybe and fixed somehow.
image

svilenvelikov:
There is no default policy preselected when new rule is being created except for the statements rules

svilenvelikov:
There is some weird behavior where I can open for edit/create rules on more than one tab and when I save one of them, the edit forms on each tabs are closed, but the tab on which I pressed the save button for the rule still appears as dirty. We should either not allow editing more than one context at a time or improve the behavior as it is right now.

svilenvelikov:
Even if I have saved the model successfuly I still see the save button and if I try to leave the page or refresh I get the confirmation 
image

src/js/angular/aclmanagement/controllers.js
teodossidossev: done

src/js/angular/aclmanagement/model.js
teodossidossev: added map types

src/js/angular/aclmanagement/model.js
teodossidossev: done

src/js/angular/aclmanagement/model.js
teodossidossev: added comment and return types

src/pages/aclmanagement.html
teodossidossev: I was thinking of making the form dynamic based on the different scopes and their properties, but that would take time. We may plan it in the next iterations

src/js/angular/aclmanagement/model.js
```diff
@@ -154,24 +267,187 @@ export class ACRuleModel {
...
+        return {
+            scope: this.scope,
+            policy: this.policy,
+            role: this.role,
+            operation: this.operation,
+            subject: this.subject,
+            predicate: this.predicate,
+            object: this.object,
+            context: this.context
+        }
```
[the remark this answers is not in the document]
teodossidossev: done

src/js/angular/aclmanagement/model.js
teodossidossev: fixed the types of the Map and all is fine now

src/js/angular/rest/mappers/aclmanagement-mapper.js
teodossidossev: It turned out it couldn't happen. I removed the check

src/js/angular/aclmanagement/model.js
teodossidossev: fixed the documentation

src/js/angular/aclmanagement/model.js
teodossidossev: renamed and added documentation

teodossidossev:
Jenkins retry build

svilenvelikov (approved):

svilenvelikov (approved):

PR #1207, opened 2024-01-22
GDB-8998 yasr table borders should be collapsed

svilenvelikov:
## What
Yasr table borders should be collapsed.

## Why
The datatable plugin used in yasr comes with the style for `table-collapse: separate` which makes the borders thicker which is not consistent with current workbench

## How
Add style `border-collapse: collapse` to data table in yasr.


svilenvelikov:
Jenkins retry build

PR #1209, opened 2024-01-23
GDB-9310 make yasqe height consistent with current workbench

svilenvelikov:
## What
Make yasqe height consistent with current workbench.

## Why
For consistency with current workbench.

## How
* Changed the height of the codemirror inside yasqe.
* Also removed the fluid container class from the yasgui wrapper to prevent the excessive padding comming from it.
* Applied darker border color to codemirror to be consistent with the one in the workbench.


svilenvelikov:
Jenkins retry build

svilenvelikov:
recreated against master after yasgui integration branch was merged

PR #1213, opened 2024-01-24
GDB-9366: Drop-down list in Resource view should not be accessible from Context tab

boyan-tonchev:
## What
When opening the "Resource" view and clicking on the "inference-select" dropdown, there is an empty option. If a new option is selected, the empty option disappear.

## Why
This can happen if the select does not match any option with the value of "nb-model". [More explanation](https://stackoverflow .com/questions/12654631/why-does-angularjs-include-an-empty-option-in-select/12654812#12654812). The problem in our code is that the value of the select is an id of a context type, but during initialization, an instance of the context type is used instead of the id. This causes the select not to match the initial value with a passed option.

## How
Fixed the initial value to be the ID of the context type instead of the instance. Renamed the variable "currentContextType" to "currentContextTypeId" for more clarity regarding the expected value.

### DIFF
--- a/src/js/angular/explore/controllers.js
+++ b/src/js/angular/explore/controllers.js
@@ -59,7 +59,7 @@ function ExploreCtrl(
 
     $scope.ContextTypes = ContextTypes;
     $scope.contextTypes = ContextType.getAllType();
-    $scope.currentContextType = ContextTypes.EXPLICIT;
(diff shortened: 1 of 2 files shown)


boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1217, opened 2024-01-26
GDB-9361 show copy of sample query in user query section in graphs config

svilenvelikov:
## What
Show copy of sample query in user query section in graphs config.

## Why
There was a wrong condition which prevented showing the query copy only on the first page.

## How
Fixed the condition allowing to show the selected sample queries from previous configs to appear in the user queries section.


boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

PR #1223, opened 2024-01-30
GDB-9458: Abort the count query when the component is being destroyed or a tab is closed.

boyan-tonchev:
# GDB-9458: Abort the count query when the component is being destroyed or a tab is closed.

## What
Aborts all ongoing count queries when leaving a view with the component on it.

## Why
When leaving the page, the processing of select queries needs to be aborted as the result is not processed. This will prevent the backend from being burdened with obsolete queries.

## How
Called Be to cancel count queries when a tab is closed, goes to another page, or the current one is reloaded.

# Additional work
### What
When hovering the mouse over a result cell with small width, the 'Copy Link' button was being displayed outside the cell.

### Why
The button was hidden, when the table was displayed, its width was not considered in the cell width calculation.

### How
Increased the "ontotext-yasgui-web-component".


svilenvelikov (dismissed):

svilenvelikov (approved):

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1230, opened 2024-02-01
fix(GDB-9484): Fix ACL functional issues

teodossidossev:
## What
Fixing functional issues in the ACL view.

## Why
For better perceiving and user experience.

## HOW
- Adjusting the positioning of the autocomplete dropdown to prevent overflow outside the viewport.
- Fixing an off-by-one error in the delete confirmation dialog.
- Modifying the behavior of the Enter key to prevent unintended actions in the input fields.
- Correcting the outline display for required fields when empty.
- Preventing toast errors related to undefined repositories.
- Updating the 'required' field handling in the autocomplete directive.
- The adjustments in the autocomplete directive ensure that dropdowns are positioned correctly and that the Enter key behaves as expected across different input scenarios.

### DIFF
--- a/src/js/angular/core/directives/autocomplete/autocomplete.directive.js
+++ b/src/js/angular/core/directives/autocomplete/autocomplete.directive.js
@@ -19,7 +19,9 @@ function autocomplete($location, toastr, ClassInstanceDetailsService, Autocomple
             styleClass: '@',
             multiline: '@',
             defaultresults: '=',
-            onModelChange: '&'
+            onModelChange: '&',
+            keypress: '&',
+            required: '=?'
         },
         templateUrl: 'js/angular/core/directives/autocomplete/templates/autocomplete.html',
         link: linkFunction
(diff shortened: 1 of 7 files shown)


src/js/angular/core/directives/autocomplete/autocomplete.directive.js
```diff
@@ -160,6 +167,42 @@ function autocomplete($location, toastr, ClassInstanceDetailsService, Autocomple
...
+         * edge with the left edge of the input field. This ensures that the dropdown is fully
+         * visible within the current viewport.
+         *
+         * The function is wrapped inside a `$timeout` to ensure that the DOM has finished
+         * updating before calculating the positions and applying any adjustments.
+         *
+         * Additionally, an angular watch is set up on `autoCompleteUriResults` to reposition
+         * the dropdown whenever the autocomplete results change.
+         */
+        $scope.positionAutocompleteWrapper = function() {
```
svilenvelikov, line 186: This behaves quite strange. When I type in the autocomplete dropdown appears and on each keystroke it starts to shift from left aligned to right which is quite awkward

src/js/angular/core/directives/autocomplete/autocomplete.directive.js
```diff
@@ -160,6 +167,42 @@ function autocomplete($location, toastr, ClassInstanceDetailsService, Autocomple
...
+            $timeout(() => {
+                const bodyRect = document.body.getBoundingClientRect();
+                const dropdown = element.find('.autocomplete-results-wrapper').get(0);
+                const dropdownRect = dropdown.getBoundingClientRect();
+
+                if (dropdown && dropdownRect.right > bodyRect.width) {
+                    dropdown.style.left = 'auto';
+                    dropdown.style.right = '0px';
+                } else {
+                    dropdown.style.right = 'auto';
```
svilenvelikov, line 196: This is incorrect. If you end up in the `else` clause, then `dropdown` might be missing according to the `if` clause above and setting the style would throw an error. I know that this can't happen probably because of the timeout above, but still it'd be better if you fix the condition because right now it's confusing.

src/js/angular/core/directives/autocomplete/autocomplete.directive.js
teodossidossev: fixed

src/js/angular/core/directives/autocomplete/autocomplete.directive.js
teodossidossev: check added

svilenvelikov (approved):

PR #1232, opened 2024-02-02
GDB-9488 make save and cancel buttons sticky in ACL view

svilenvelikov:
## What
Make save and cancel buttons sticky in ACL view.

## Why
In order to allow the user to have fast and easy access to those operations.

## How
Styled the panel with the buttons to be sticky and have a background. Also added a message to warn the user that acl rules are not applied until the acl list is saved.

### DIFF
--- a/src/css/aclmanagement.css
+++ b/src/css/aclmanagement.css
@@ -2,6 +2,18 @@
     table-layout: fixed;
 }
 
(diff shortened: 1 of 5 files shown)


teodossidossev (approved):

svilenvelikov:
Jenkins retry build

PR #1236, opened 2024-02-05
Fixes GDB-9475, GDB-9476 and GDB-9483

boyan-tonchev:
## What
 - GDB-9476: this issue arises when there are no triples matching the query, leading to improper display empty row message;
 - GDB-9483: fix query explain plan styling when displayed in multiple yasgui tabs;
 - GDB-9475: There is an issue with the display of extended-table plugin controls. The problem can be reproduced with the following scenario:
  - execute a select query;
  - execute an update query.
 - GDB-9475: The info message that describes the result of executed queries is differently positioned than in the old implementation of YASGUI.

## Why
 - GDB-9476: the empty result message is displayed as row of datatable (jquery datatable behaviour). But the first column (row number) can be hidden and this causes the issues;
 - GDB-9483: there was a wrong selector from the document for the explain plan html element on which the styling should be applied which in this case was always the first yasgui tab where there is one. In result the rest tabs with displayed explain plan remained unstyled;
 - GDB-9475: The plugin controls are cleared only if the plugin is changed. In this case, even though the plugin remains the same, no plugin will be displayed;
 - GDB-9475: The header of the result has two sections. The first section is where the rendered plugin can insert some controls related to its functionality, and the second section is the message describing the processed query. Between these sections, there is an element that takes up all available space. In the result, the first section is on the left, and the second is on the right side. The message from the issue is displayed for an update query. The update queries do not return a result, and there is no registered plugin to be rendered. In this case, the old implementation showed the message on the left. However, in the new implementation, the message is pushed out from the spacer on the right side.

## How
Increased the "ontotext-yasgui-web-component" version.

# Additional work
 - Added translation for no results message.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
                 "ng-file-upload": "^12.2.13",
                 "ng-tags-input": "^3.2.0",
                 "oclazyload": "^1.1.0",
-                "ontotext-yasgui-web-component": "1.1.24",
+                "ontotext-yasgui-web-component": "1.1.25",
                 "shepherd.js": "^11.1.1"
             },
             "devDependencies": {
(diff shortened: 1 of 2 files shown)


boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1239, opened 2024-02-05
GDB-9471 enable inferred data and expand on same as property buttons in visual graph config

svilenvelikov:
## What
Enable inferred data and expand on same as property buttons in visual graph config view.

## Why
The user should be able to configure these two properties in the visual graph config.

## How
Provided configuration for enabling these two buttons in the yasqe.

(cherry picked from commit 66631591eb36020b3f97ccdfe7dd09eaddc218e5)


svilenvelikov:
Jenkins retry build

PR #1244, opened 2024-02-07
Increase yasgui version to 1.2.1

svilenvelikov:
## What
Increase yasgui version to 1.2.1

## Why
Includes fixes for:
* GDB-9500 fix explain plan styling

## What
Increased version


svilenvelikov:
Jenkins retry build

PR #1246, opened 2024-02-07
GDB-8888 - Add tooltip to query monitoring "Pause" button

DesiBorisova:
## What?
The "Pause" button on the Monitor -> Queries and Updates page has a different tooltip when in the paused state.
image

## Why?
The tooltip was the same as in the unpaused state, possibly leading to some confusion.

## How?
I added a condition check for the button state before the tooltip is set.


src/i18n/locale-en.json
```diff
@@ -264,7 +264,8 @@
     "next.btn": "Next",
     "closes.config.no.save": "Closes the configuration without saving the changes.",
     "refresh.to.retry": "Refresh the page to retry.",
-    "temp.pause": "Temporarily pauses query monitoring so you can copy text",
+    "temp.pause": "Temporarily pauses query monitoring, so you can copy text",
+    "monitoring.restart": "Unpauses query monitoring, so you you can see query status updates",
```
boyan-tonchev, line 268: double "you"

boyan-tonchev (approved):

PR #1247, opened 2024-02-08
GDB-9504: Create rule button not translated to french

boyan-tonchev:
## What
The "Create rule" button is not translated to French.

## Why
The label was hardcoded, it missed to be translated.

## How
Added translation to the button.

### DIFF
--- a/src/i18n/locale-fr.json
+++ b/src/i18n/locale-fr.json
@@ -143,6 +143,9 @@
         }
     },
     "acl_management": {
+        "buttons": {
+            "create_rule": "Créer une règle"
+        },
         "rulestable": {
             "tab": {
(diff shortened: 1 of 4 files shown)


src/i18n/locale-fr.json
```diff
@@ -143,6 +143,9 @@
         }
     },
     "acl_management": {
+        "buttons": {
```
svilenvelikov, line 146: There is an `actions` section below. Move the label there pls

boyan-tonchev: Done

boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1248, opened 2024-02-08
Cherry-pick from 2.6 GDB-9504: Create rule button not translated to french

boyan-tonchev:
## What
The "Create rule" button is not translated to French.

## Why
The label was hardcoded, it missed to be translated.

## How
Added translation to the button.


svilenvelikov (approved):

boyan-tonchev:
Jenkins retry build

PR #1253, opened 2024-02-13
GDB-9513: Saved queries are not displayed on the welcome page

boyan-tonchev:
## What
Saved SPARQL queries are not loaded when the user is logged in.

## Why
The authentication headers are still not set correctly when a request that loads saved queries is called and the $unauthorizedInterceptor rejects the request.

## How
There was another issue (GDB-8091) with the same problem. Then, a setTimeout was added, which in this case is not enough for the authentication header to be properly set. The loading of saved queries was moved after the initialization of the authentication principal.

** Created GDB-9547 issue for tracking the effort for improving the implementation instead of just finding workarounds


boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1261, opened 2024-02-16
GDB-9549 fix sparql query endpoint resolving

svilenvelikov:
## What
Fix sparql query endpoint resolving.

## Why
When GDB WB is deployed behind a context path requests to all endpoints which the WB calls should start without `/` otherwise this might brake the request routing.

## How
Removed the leading `/` from the sparql query endpoint used by the sparql editor and also by the download as feature.


svilenvelikov:
Jenkins retry build

PR #1268, opened 2024-02-21
Fix UI on RDF import page for Google Chrome browsers for GraphDB 9

ww-jermaine:
# What

When I navigate to the RDF import section of the GraphDB interface, the styling does not appear as expected. It seems that the CSS is broken, which impacts the usability of the page. This happens when using Chrome browser and the GraphDB version is 9.

image

# Why

The styling might be messing up the layout.

# How

I added a fix adjusting the sizes and spacing of each buttons.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "graphdb-workbench",
-    "version": "1.11.1",
+    "version": "1.11.2",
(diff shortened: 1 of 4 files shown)


hudson-onto:

PR #1270, opened 2024-02-21
GDB-9516 clear the rule highlighting on tab switch

svilenvelikov:
## What
Clear the rule highlighting on tab switch.

## Why
It's annoying bug where the selected rule index remains uncleared in the controller when tab is switched which causes a flickering of previously moved rule.

## How
* Cleared the selected rule index in the controller to prevent flickering on tab switch.
* Clear the selected rule property when moving rules up/down and the model appears to be pristine.


boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

PR #1288, opened 2024-02-23
GDB-9548 - Add color in dark theme for keyboard shortcut text

DesiBorisova:
## What?
The keyboard shortcuts in dark theme on the Visual graph view and SPARQL view will match the color of the main menu text.

## Why?
The keyboard shortcuts would remain the same color as they were in the light theme.

## How?
I added a new variable in the dark theme for the shortcut text.

## Screenshots?
**Both views in dark mode:**

Screenshot from 2024-02-23 11-49-06
Screenshot from 2024-02-23 11-48-48

### DIFF
--- a/src/themes/default/plugin.js
+++ b/src/themes/default/plugin.js
@@ -58,7 +58,8 @@ PluginRegistry.add('themes', {
             'primary-color-lightness': '60%',
             'secondary-color-saturation': '70%',
             'color-warning-light': 'hsla(var(--primary-color-hsl), 0.15)',
-            'logo-color': 'var(--primary-color-dark)'
+            'logo-color': 'var(--primary-color-dark)',
+            'keyboard-shortcuts': 'var(--secondary-color)'
         },
         // CSS properties, "foo: bar" becomes "foo: bar"
         'properties': {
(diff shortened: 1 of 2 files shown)


src/themes/default/plugin.js
```diff
@@ -58,7 +58,8 @@ PluginRegistry.add('themes', {
             'primary-color-lightness': '60%',
             'secondary-color-saturation': '70%',
             'color-warning-light': 'hsla(var(--primary-color-hsl), 0.15)',
-            'logo-color': 'var(--primary-color-dark)'
+            'logo-color': 'var(--primary-color-dark)',
+            'keyboard-shortcuts': 'var(--secondary-color)'
```
svilenvelikov, line 62: Why did this was added here? Can't we just use the secondary-color var in the dark theme stylesheet?

DesiBorisova:
```
.hotkeys-text, .keyboard-shortcut-description-item-description {
    color: var(--secondary-color);
}
```
Adding only this to the bootstrap-graphdb-theme-dark-auto.css applies the color in light theme as well.

PR #1290, opened 2024-02-23
GDB 9466 9211 9198 JSON-LD UI

VPetrova00:
### **What**

- Create import JSON-LD context input element in the import settings modal.
- Create export JSON-LD settings modal (it appears when exporting in graph overview and when downloading as in the resource view and after execution of construct query).
- Added some tests.

### **Why**
In order exporting in JSON-LD format to be more user friendly.

### **How**

- Added JSON-LD context input element when importing from text snippet, file and URL when JSON-LD or NDJSON-LD format is selected (or in other situations described in detail here - [https://docs.google.com/document/d/19uOg_mC4iFWbeyn5H-Y89Q_x77qhRWO2TofutzT8cvw/edit](https://docs.google.com/document/d/19uOg_mC4iFWbeyn5H-Y89Q_x77qhRWO2TofutzT8cvw/edit)

- Created a new modal for export settings.
- Added tests for some of the functionality.


src/pages/sparql-editor.html
```diff
@@ -21,6 +21,7 @@ <h1 id="sparql-query-update-title-label">
         <input id="wb-download-infer" name="infer" type="hidden"/>
         <input id="wb-download-sameAs" name="sameAs" type="hidden"/>
         <input id="wb-download-accept" name="Accept" type="hidden"/>
+        <input id="wb-download-link" name="Link" type="hidden">
```
svilenvelikov, line 24: Now this form isn't used anymore, is it?

svilenvelikov (changes requested):

src/js/angular/import/controllers.js
```diff
@@ -186,9 +186,22 @@ importCtrl.controller('CommonCtrl', ['$scope', '$http', 'toastr', '$interval', '
...
             $scope.settingsFor = fileName;
             $scope.settings = $scope.getSettingsFor(fileName, withDefaultSettings);
+            if (fileName === "" ||
+                format === "application/ld+json" ||
+                format === "application/x-ld+ndjson" ||
+                fileName.endsWith("jsonld") ||
+                fileName.endsWith("zip") ||
+                fileName.endsWith("gz") ||
+                $scope.settings.type === "directory") {
+                $scope.settings.isJSONLD = true;
```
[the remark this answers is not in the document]
VPetrova00: We have discussed with Sava that it's not necessary to check if an archive or a directory contains jsonld files when importing them. And that is why I set isJSONLD variable to be true in these cases. If there are jsonld files the context will be used, in other cases nothing will happen.

webpack.config.common.js
```diff
@@ -181,12 +181,8 @@ module.exports = {
...
             },
             {
-                from: 'src/js/angular/core/templates/modal',
-                to: 'js/angular/core/templates/modal'
-            },
-            {
-                from: 'src/js/angular/templates',
-                to: 'js/angular/templates'
+                from: 'src/js/angular/core/components/export-settings-modal',
+                to: 'js/angular/core/components/export-settings-modal'
```
svilenvelikov, line 185: Can you also put this `transform: replaceVersion` at the end! This ensures that a new hash is added in the end of the file name on each change of the template so that the browser to know that there is a change in the file and it should be reloaded.

svilenvelikov:
Jenkins retry build

svilenvelikov (approved):

PR #1301, opened 2024-02-29
feat(GDB-9634) Enhance node info display with dynamic sizing and interactivity

teodossidossev:
## WHAT
To improve the user experience in cluster management, the display of node statuses has been enhanced to dynamically adjust the size based on the content and to allow interactivity for viewing longer messages

## HOW:
- Introduced `font` constants for icon usage as we use icons from different fonts
- Modified the `updateNodesIcon` function to dynamically select icons based on the node's recovery state
- Implemented a new approach for displaying node recovery messages using `foreignObject` in SVG, allowing HTML content to dynamically resize and wrap text. This includes calculating the size of the text to set the `foreignObject` dimensions appropriately and updating these dimensions upon user interaction
- Added `recovery_operation_failure_warning` to the translations

cherry-pick from https://github.com/Ontotext-AD/graphdb-workbench/pull/1296


src/js/angular/clustermanagement/services/cluster-drawing.service.js
```diff
@@ -7,6 +7,13 @@ export const clusterColors = {
     ontoGrey: 'var(--gray-color)'
 };
 
+export const font = {
```
yordanalexandrov, line 10: Should be uppercase as a constant

src/js/angular/clustermanagement/services/cluster-drawing.service.js
```diff
@@ -177,68 +180,149 @@ function getNodeInfoIconType(node) {
 
     switch (node.recoveryStatus.state) {
         case RecoveryState.SEARCHING_FOR_NODE:
```
yordanalexandrov, line 182: For both functions you can construct an object before the switch and just set the `icon` value. Then return the object. It would be more readable IMO.

src/js/angular/clustermanagement/services/cluster-drawing.service.js
```diff
@@ -177,68 +180,149 @@ function getNodeInfoIconType(node) {
...
+
+function extractShortMessageFromNode(node) {
+    return node.recoveryStatus.message && node.recoveryStatus.message.length > shortMessageLimit ?
+        node.recoveryStatus.message.substring(0, shortMessageLimit) + '...' : node.recoveryStatus.message || '';
+}
+
+function calculateElementSizeByText(text) {
+    const tmp = document.createElement('div');
+    tmp.innerHTML = text;
+    tmp.style.padding = '4px';
```
yordanalexandrov, line 282: Would have saved this if it was a class

svilenvelikov (approved):

src/js/angular/clustermanagement/services/cluster-drawing.service.js
teodossidossev: This is a cherry pick. This and the other comments will be fixed with the next cluster task.

PR #1313, opened 2024-03-08
GDB-9126: Can't download json result from construct query

boyan-tonchev:
## What
There is an error when try to export results of query, that contains data.

## Why
The service which is responsible for downloading results makes a POST call to BE with "Content-Type" "application/x-www-form-urlencoded". The queries which causes the issue includes a data, for instance "FILTER (xsd:dateTime(?dateOfBirth) > "1999-03-22T14:17:31.290+02:00"^^xsd:dateTime)". The browser replace the character "+" with " " because of value of content type and this is the reason error appeared in BE. It can't parse data correctly.

## How
Implemented query encoding before sending it to the backend. This encoding ensures correct parsing on the server side.


boyan-tonchev:
Jenkins retry build

PR #1318, opened 2024-03-14
GDB-9238 - hide remote locations already added to cluster node list

DesiBorisova:
## What?
If a node goes offline, the add/replace node dialogs won't show it as an available remote location, until the node is removed first.

## Why?
Previously, the node would appear as both added to the cluster and as an available location, which would cause some confusion.

## How?
The RPC address was checked previously and nodes without connection had no RPC address, meaning they didn't match the cluster node list and appeared as an available location.
I used an array to extract just the cluster node endpoints and used them to compare the locations against.


yordanalexandrov (approved):

DesiBorisova:
Jenkins retry build

DesiBorisova:
Jenkins retry build

PR #1319, opened 2024-03-15
GDB-9948 - Add NDJSON-LD download to views where missing

DesiBorisova:
## What?
The option to download as `NDJSON-LD` is now present where `JSON-LD` is present.

## Why?
The option was missing in `Graphs overview` and `Resource` view.

## How?
I used the same logic as the `JSON-LD` download. The dialog is reused for both.
I added dynamic labels within the dialog by passing in the file format in all views, which open it.
I restructured the list HTML elements, so that the file types can appear in the same order in all the drop-downs across views.
I edited test steps, where the CSS classes needed to be updated.


src/pages/export.html
```diff
@@ -98,16 +96,14 @@ <h1>
 										<button type="button" class="btn btn-link dropdown-toggle" uib-dropdown-toggle gdb-tooltip="{{'export.selected.graphs.tooltip' | translate}}" ng-disabled="!hasMultipleSelected()">
 											<span class="icon-export"></span></button>
 										<ul class="dropdown-menu" role="menu">
-                                            <li ng-repeat="format in exportFormats" class="export-repo-format-{{format.name}}"
-                                                ng-if="format.name === 'JSON-LD';"
-                                                ng-click="openJSONLDExportSettingsForSelectedGraphs(format);">
-                                                <span class="dropdown-item">{{format.name}}</span>
+                                            <li ng-repeat="format in exportFormats">
+                                                <span ng-click="openJSONLDExportSettingsForSelectedGraphs(format)"
```
svilenvelikov, line 100: Do you mind replacing these spans with links in order the html to be semantically valid?

src/pages/export.html
```diff
@@ -150,16 +146,14 @@ <h1>
 											ng-disabled="!graph.contextID.uri">
 										<span class="icon-export"></span></button>
 									<ul class="dropdown-menu" role="menu">
-                                        <li ng-repeat="format in exportFormats" class="export-repo-format-{{format.name}}"
-                                            ng-if="format.name === 'JSON-LD';"
-                                            ng-click="openJSONLDExportSettings(format, graph.contextID, false);">
-                                            <span class="dropdown-item">{{format.name}}</span>
+                                        <li ng-repeat="format in exportFormats">
+                                            <span ng-click="openJSONLDExportSettings(format, graph.contextID, false)"
```
svilenvelikov, line 150: Same here!

svilenvelikov:
When I'm exporting in ndjson-ld format I still see `Export settings: JSON-LD` in the dialog title. 
image

src/pages/export.html
DesiBorisova: Sure, I'll replace all of them.

src/js/angular/core/components/export-settings-modal/controller.js
```diff
@@ -23,7 +23,7 @@ export function ExportSettingsCtrl($scope, $uibModalInstance, LocalStorageAdapte
     $scope.defaultMode = $scope.JSONLDModes.find((mode) => mode.name === 'expanded');
     $scope.currentMode = $scope.defaultMode;
     $scope.link = null;
-
+    $scope.fileFormat = $scope.$parent.format.toLowerCase();
```
svilenvelikov, line 26: I don't like this touching things in the parent scope. Can't you just pass this as an option to the  modal on open?

DesiBorisova: Please see the proposed change I've made

svilenvelikov (approved):

PR #1332, opened 2024-03-28
GDB-10004 Fixed issues in export.rest.service.js related with the additional encoding by us. Also fixed issue when multiple graphs are selected for export in both formats.

sava-savov-ontotext:
What?
Fixed bug in exporting multiple graphs. Also removed additional encoding.

Why?
This update was done, because the passed parameters were encoded once by us and afterwards they are encoded by http method. Also when multiple graphs were selected for export instead of passing array of the selected items we were passing concatenated string of the graphs with &context=, which is totally wrong.

How?
I decode the already encoded value(s) of the context(s) and do not encode location. Also instead of concatenating the selected contexts I return array of the values.

### DIFF
--- a/src/js/angular/export/controllers.js
+++ b/src/js/angular/export/controllers.js
@@ -146,9 +146,8 @@ exportCtrl.controller('ExportCtrl',
              *
              * @method downloadJSONLDExport
              * @param {String} data format
-             * @param {String} string context if there is any (or string from multiple contexts if there are multiple selected graphs)
+             * @param [String] array of string contexts if multiple graphs were selected or single graph object
              * @param {String} context/frame link
-             * @param {Boolean} true if the method is invoked for export of multiple selected graphs
              * @param {Object} current repository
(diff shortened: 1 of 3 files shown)


src/js/angular/export/controllers.js
```diff
@@ -153,13 +153,13 @@ exportCtrl.controller('ExportCtrl',
              * @param {Object} graphsByValue
              * @param {Object} JSONLDMode (name and mode link)
              */
-            function downloadJSONLDExport(format, context, link, forSelectedGraphs, repo, graphsByValue, JSONLDMode) {
```
svilenvelikov, line 156: Can you just update the jsdoc to account changes in the parameters list. Do this for all methods you changed

svilenvelikov:
Add description of the MR following the format we use. See here for example https://github.com/Ontotext-AD/graphdb-workbench/pull/1331

svilenvelikov (approved):

PR #1336, opened 2024-03-29
Draft: GDB-10046 - Properly update recovery messages for cluster nodes

DesiBorisova:
## What?
The messages for each cluster node during recovery will update correctly and according to the icons for that node.

## Why?
The user needed to click the message for it to update. Hence, the icon would have changed on its own, but without user interaction, the message would remain out of date.

## How?
I removed the click listener and added some validation for the case with no message. I added the missing height attribute, preventing the messages from being seen without a page reload. I also refactored some of the existing code.


src/js/angular/clustermanagement/services/cluster-drawing.service.js
```diff
@@ -174,28 +174,23 @@ function hasRecoveryState(node) {
...
-            return {icon: '\uf0ee', font: font.FONT_AWESOME};
-        case RecoveryState.RECOVERY_OPERATION_FAILURE_WARNING:
-            return {icon: '\ue920', font: font.ICOMOON};
-        default:
-            return {icon: '', font: ''};
+    if (!status || !status.state) {
+        return {icon: '', font: ''};
     }
+
+    const iconMap = {
```
svilenvelikov, line 183: Move this out of the method otherwise it'd be created on each invocation.

DesiBorisova:
Jenkins retry build

PR #1337, opened 2024-04-01
GDB-9972 change file upload workflow

svilenvelikov:
## What
Change the upload files workflow in the user data tab on the import view.

## Why
The new upload workflow allows the user to decide if newly selected for upload files should override existing ones if filename collision occurs. If override is rejected, then all duplicated filenames are prefixed with a numerical index.

## How
* Extended the import controller  and template so that when new files are selected to check for duplications and open a confirmation modal.
* Implemented additional utilities to calculate the file prefixes.
* Added new tests

### DIFF
--- a/src/js/angular/import/controllers.js
+++ b/src/js/angular/import/controllers.js
@@ -11,6 +11,8 @@ import {FileUtils} from "../utils/file-utils";
 import {DateUtils} from "../utils/date-utils";
 import {toImportResource, toImportServerResource} from "../rest/mappers/import-mapper";
 import {ImportResourceTreeElement} from "../models/import/import-resource-tree-element";
+import {decodeHTML} from "../../../app";
+import {FilePrefixRegistry} from "./file-prefix-registry";
 
 const modules = [
     'ui.bootstrap',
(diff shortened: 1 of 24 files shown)


src/js/angular/import/controllers.js
```diff
@@ -695,20 +793,43 @@ importViewModule.controller('UploadCtrl', ['$scope', 'toastr', '$controller', '$
...
+     */
+    const buildPrefixesRegistry = (files) => {
+        files.forEach((file) => {
+            // file name is in format 123-file-name.txt
+            // split the number from the file name
+            const separatorIndex = file.name.indexOf('-');
+            let index = file.name.substr(0, separatorIndex);
+            const filename = file.name.substr(separatorIndex + 1);
+            if (index) {
+                index = index * 1;
```
boyan-tonchev, line 820: IMO `index = parseInt(index, 10)` is more readable.

src/js/angular/import/controllers.js
```diff
@@ -695,20 +793,43 @@ importViewModule.controller('UploadCtrl', ['$scope', 'toastr', '$controller', '$
...
+     * Builds a registry of filenames mapped to numerical indices.
+     * The registry is in format <code>{filename: index}</code>.
+     * @param {[object]} files - list of files to be used when building the registry.
+     */
+    const buildPrefixesRegistry = (files) => {
+        files.forEach((file) => {
+            // file name is in format 123-file-name.txt
+            // split the number from the file name
+            const separatorIndex = file.name.indexOf('-');
+            let index = file.name.substr(0, separatorIndex);
```
boyan-tonchev, line 817: Maybe the separator is not correct. Try to import an RDF text snippet, its name not start with number and "-".

boyan-tonchev (dismissed):

svilenvelikov:
Jenkins retry build

PR #1343, opened 2024-04-02
Gdb 9971 implements import resources buttons

boyan-tonchev:
## What
- Added "Reset Status" and "Remove" buttons to the import-resource-tree directive;
- Added the "Import" button to the import-resource-tree directive;
- Added a dropdown menu for selecting files based on their status.

## Why
- This addition provides users with the ability to reset the status of imported resources and to remove files if they were uploaded by the user;
- This addition provides users with the ability to trigger the import of files with or without changing settings
- This addition provides users with the ability to select files based on their status.

### DIFF
--- a/src/i18n/locale-en.json
+++ b/src/i18n/locale-en.json
@@ -993,6 +993,12 @@
                 "modified": "Modified",
                 "imported": "Imported",
                 "context": "Context"
+            },
+            "status": {
+                "ALL": "All",
+                "NONE": "None",
+                "IMPORTED": "Imported",
+                "NOT_IMPORTED": "Not imported"
             }
         }
     },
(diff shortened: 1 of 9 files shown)


src/i18n/locale-en.json
```diff
@@ -984,6 +984,12 @@
                 "modified": "Modified",
                 "imported": "Imported",
                 "context": "Context"
+            },
```
svilenvelikov, line 987: update the bundle in test fixtures

svilenvelikov (approved):

src/i18n/locale-en.json
boyan-tonchev: Done

svilenvelikov (approved):

PR #1345, opened 2024-04-03
GDB-9972 Automatic import after file upload

svilenvelikov:
## What
After a file upload the import process should start immediately. This should happen either for a single or multiple uploaded files. If the import is rejected, then files are only uploaded instead.

## Why
This would save the user some additional steps which are rarely needed.

## How
Extended the logic in the import controller so that the file import process to start automatically after file selection. 
Renamed test specs for to be more meaningful.
Implemented more tests to cover the functionality.


src/js/angular/import/controllers.js
```diff
@@ -143,30 +144,37 @@ importViewModule.controller('ImportViewCtrl', ['$scope', 'toastr', '$interval',
...
+                    $scope.settings = settings;
+                    if ($scope.settingsFor === '') {
+                        $scope.importSelected();
+                    } else {
+                        $scope.importFile($scope.settingsFor, true);
+                    }
+                },
+                (settings) => {
+                    $scope.settings = settings;
+                    onImportRejectHandler && onImportRejectHandler();
```
yordanalexandrov, line 163: Doesn't sonar complain about this kind of assesments? In other projects it does.

src/js/angular/import/controllers.js
```diff
@@ -715,9 +734,9 @@ importViewModule.controller('UploadCtrl', ['$scope', 'toastr', '$controller', '$
         $scope.settings.name = file.name;
         const data = UploadRestService.createUploadPayload(file, $scope.settings);
         const uploader = startImport ? UploadRestService.uploadUserDataFile : UploadRestService.updateUserDataFile;
-        uploader($repositories.getActiveRepository(), file, data).success(function () {
+        uploader($repositories.getActiveRepository(), file, data).then(() => {
             $scope.updateList();
-        }).error(function (data) {
+        }).catch((data) => {
```
yordanalexandrov, line 739: When mixing `success/error` and `then/catch` keep in mind that the first returns only the body and the latter returns the whole respones object.

yordanalexandrov (approved):

boyan-tonchev (dismissed):

src/js/angular/import/controllers.js
svilenvelikov: It does. It suggests optional chaining instead. I'll change it.

src/js/angular/import/controllers.js
svilenvelikov: `getError(data)` seems to handle this

src/js/angular/import/controllers.js
yordanalexandrov: Using an optional chaining would also show as a code smell.
You should check with an `if` statement

svilenvelikov: > Using an optional chaining would also show as a code smell.

Really?! Why they are suggesting it then? :)
Anyway. I did it with an if clause because I can't do the optional chaining in this case.

yordanalexandrov (approved):

PR #1346, opened 2024-04-03
GDB-9971: Implement Three-State Checkbox

boyan-tonchev:
## What
Implements an additional selection state for the "import-resource-tree" directive. This state is applied to directories indicates that not all children are selected.

## Why
To enhance user experience.

## How
Implemented a new "propIndeterminate" directive that marks a checkbox as indeterminate. The "import-resource-tree" directive utilizes this directive to mark directories where some, but not all, children are selected.

### DIFF
--- a/src/js/angular/core/directives/prop-indeterminate/prop-indeterminate.directive.js
+++ b/src/js/angular/core/directives/prop-indeterminate/prop-indeterminate.directive.js
@@ -1,0 +1,52 @@
+/**
+ * @ngdoc directive
+ * @name propIndeterminate
+ * @module graphdb.framework.core.directives.prop-indeterminate
+ * @restrict A
+ *
+ * @description
+ * The `propIndeterminate` directive updates the `indeterminate` property of an HTMLInputElement.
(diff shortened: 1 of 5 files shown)


svilenvelikov (approved):

src/js/angular/core/directives/prop-indeterminate/prop-indeterminate.directive.js
```diff
@@ -0,0 +1,52 @@
...
+
+function propIndeterminate() {
+    return {
+        restrict: 'A',
+        scope: {
+            propIndeterminate: '='
+        },
+        link: ($scope, element, attributes) => {
+            const indeterminateChanged = () => {
+                element.prop('indeterminate', !!$scope.propIndeterminate);
```
yordanalexandrov, line 34: Does it get the `true/false` as a boolean or as a string?

yordanalexandrov: If used as in the given example

yordanalexandrov (approved):

PR #1347, opened 2024-04-03
GDB-10070 improve import settings dialog

svilenvelikov:
## What
Reorganize and improve styles and validation in the import settings dialog.

## Why
This is expected to improve the user experience.

## How
* Moved the base uri field in the advanced options section as it's rarely used.
* Aligned labels with tier respective checkboxes.
* Fixed IRI validation for the context, base iri and jsonld context link.
* Added tests for the settings modal

### DIFF
--- a/src/js/angular/utils/uri-utils.js
+++ b/src/js/angular/utils/uri-utils.js
@@ -31,10 +31,17 @@ function UriUtils(ClassInstanceDetailsService) {
         return validProtocol && validPath;
     }
 
+    /**
+     * Validates if the given text is a valid IRI.
+     * @param {object} value
+     * @param {string} valueText
+     * @return {boolean} Returns true if the value is a valid IRI and false otherwise.
+     */
     function isValidIri(value, valueText) {
         if (value !== undefined && valueText !==undefined && (validateRdfUri(valueText) || !!valueText.match(iriRegExp))) {
             return true;
         }
+        return false;
     }
 
     function shortenIri(iri) {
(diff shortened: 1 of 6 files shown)


yordanalexandrov (dismissed):

src/js/angular/utils/uri-utils.js
```diff
@@ -31,10 +31,17 @@ function UriUtils(ClassInstanceDetailsService) {
...
+     * Validates if the given text is a valid IRI.
+     * @param {object} value
+     * @param {string} valueText
+     * @return {boolean} Returns true if the value is a valid IRI and false otherwise.
+     */
     function isValidIri(value, valueText) {
         if (value !== undefined && valueText !==undefined && (validateRdfUri(valueText) || !!valueText.match(iriRegExp))) {
             return true;
         }
+        return false;
```
svilenvelikov, line 44: Should return true/false in order to be more predictable. Additionally, the `undefined` value returned by this function instead of `false` turned out to brake the validation directive which expects true/false

test-cypress/integration/import/import-user-data-settings-dialog.spec.js
```diff
@@ -0,0 +1,71 @@
...
+        cy.presetRepository(repositoryId);
+        ImportSteps.visitImport('user', repositoryId);
+    });
+
+    afterEach(() => {
+        cy.deleteUploadedFile(repositoryId, testFiles);
+        cy.deleteRepository(repositoryId);
+    });
+
+    it.skip('Should prevent import when there are missing or invalid settings', () => {
```
svilenvelikov, line 35: Will be implemented when my other branch gets merged

yordanalexandrov (approved):

PR #1349, opened 2024-04-05
Gdb 9971 add stylint to import resource messages

boyan-tonchev:
## What
- Icons and colors have been added to resource messages;
- Implemented a reset status button for each resource in the tree.

## Why
- To enhance the user experience;
- This provides users with the opportunity to reset the status of individual files without having to click on checkboxes and use the global reset button.

## How
- Different icons and colors have been applied to resource messages based on the import resource status. For example, errors are shown in red, successful operations in green, etc.
- The button has been added to each row in the import resource tree. When clicked, it calls the onReset function of the "import-resource-tree" directive client.

### DIFF
--- a/src/js/angular/import/templates/import-resource-tree.html
+++ b/src/js/angular/import/templates/import-resource-tree.html
@@ -115,6 +115,13 @@
             <td class="import-resource-cell">
                 <div class="import-resource-actions">
                     <div class="icon-info text-tertiary import-resource-action-info"></div>
+                    <button role="button" class="btn btn-link btn-inline secondary pr-0"
+                            ng-click="resetStatus(resource)"
+                            ng-show="resource.importResource.status === ImportResourceStatus.DONE || resource.importResource.status === ImportResourceStatus.ERROR"
+                            ng-disabled="selectedResources.length > 1"
+                            gdb-tooltip="{{'import.reset.status' | translate}}">
+                        <span class="icon-close"></span>
+                    </button>
                     <button class="btn btn-link secondary import-resource-action-remove" ng-click="onDelete({resource: resource})">
                         <em class="icon-trash"></em>
                     </button>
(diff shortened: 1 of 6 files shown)


src/js/angular/import/templates/import-resource-tree.html
```diff
@@ -128,7 +135,22 @@
             <td class="import-resource-cell import-resource-cell-select"></td>
             <td ng-style="{'padding-left': (filterByType !== TYPE_FILTER_OPTIONS.FILE ? resource.indent : '8px')}"
                 class="import-resource-cell import-resource-message">
-                {{ resource.importResource.message || ''}}
+                <span ng-if="resource.importResource.status === ImportResourceStatus.NONE && resource.importResource.message" class="text-info">
+                    <small>{{resource.importResource.message || 'sssss'}}</small>
```
svilenvelikov, line 139: ?

src/js/angular/import/templates/import-resource-tree.html
```diff
@@ -128,7 +135,22 @@
             <td class="import-resource-cell import-resource-cell-select"></td>
             <td ng-style="{'padding-left': (filterByType !== TYPE_FILTER_OPTIONS.FILE ? resource.indent : '8px')}"
                 class="import-resource-cell import-resource-message">
-                {{ resource.importResource.message || ''}}
```
svilenvelikov, line 131: split the whole thing in separate component maybe

src/js/angular/import/templates/import-resource-tree.html
boyan-tonchev: Ups. I will remove it.

boyan-tonchev: Done

src/js/angular/import/templates/import-resource-tree.html
boyan-tonchev: Done

PR #1350, opened 2024-04-05
GDB-10046 - show full recovery message on mouseover

DesiBorisova:
## What?
The shortened recovery messages in the cluster view will appear full length on `mouseover`. They will become shortened on `mouseout`.

## Why?
There was no way to see the full recovery message. 

## How?
I added event listeners in the service and used the original message size in the calculations.

## Screenshots?
On mouse over:
image

### DIFF
--- a/src/js/angular/clustermanagement/services/cluster-drawing.service.js
+++ b/src/js/angular/clustermanagement/services/cluster-drawing.service.js
@@ -294,6 +294,23 @@ function calculateElementSizeByText(text) {
     return {height, width};
 }
 
+function addRecoveryLabelListeners(object, message, width, height, displayMessage) {
+    object.on('mouseover.tooltip', () => {
+        object
+            .attr('width', calculateElementSizeByText(message).width)
+            .attr('height', calculateElementSizeByText(message).height)
+            .attr('x', -calculateElementSizeByText(message).width / 2)
+            .html(`<div>${message}</div>`);
+    });
+    object.on('mouseout.tooltip', () => {
+        object
+            .attr('width', width)
+            .attr('height', height)
+            .attr('x', -width / 2)
+            .html(`<div>${displayMessage}</div>`);
+    });
+}
+
 function resizeLabelDynamic(nodes) {
     nodes.select('.node-info-fo')
         .each(function (d) {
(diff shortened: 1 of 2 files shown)


src/js/angular/clustermanagement/services/cluster-drawing.service.js
```diff
@@ -294,6 +294,23 @@ function calculateElementSizeByText(text) {
     return {height, width};
 }
 
+function addRecoveryLabelListeners(object, message, width, height, displayMessage) {
+    object.on('mouseover.tooltip', () => {
+        object
+            .attr('width', calculateElementSizeByText(message).width)
+            .attr('height', calculateElementSizeByText(message).height)
+            .attr('x', -calculateElementSizeByText(message).width / 2)
```
yordanalexandrov, line 302: Is there some way to skip calculating the same thing twice?

src/js/angular/clustermanagement/services/cluster-drawing.service.js
```diff
@@ -307,6 +324,12 @@ function resizeLabelDynamic(nodes) {
             const displayMessage = isShorten ? message : d.recoveryStatus.message;
             const width = calculateElementSizeByText(displayMessage).width;
             const height = calculateElementSizeByText(displayMessage).height;
+            object.on('.tooltip', null);
+            if (isShorten) {
+                addRecoveryLabelListeners(object, d.recoveryStatus.message, width, height, displayMessage);
```
yordanalexandrov, line 329: The way D3 works with datasets is that you define `enter/exit` functions which are applied to action on entries in the dataset.

When you add node, or when you remove a node.

This in mind I think you should unsubscribe in the `exit` phase, because the dataset could have changed.

@teodossidossev WDYT?

src/js/angular/clustermanagement/services/cluster-drawing.service.js
DesiBorisova: Possibly. Assigning them to a field would need to happen within the listener, or maybe in the method which calls this one.

src/js/angular/clustermanagement/services/cluster-drawing.service.js
teodossidossev: Absolutely. If you don't unsubscribe on exit, another listener will be added to the object every time you enter it again. This accumulation will lead to unpredicted behavior.

yordanalexandrov (approved):

src/js/angular/clustermanagement/services/cluster-drawing.service.js
DesiBorisova: The documentation on event handling in `D3` (https://d3js.org/d3-selection/events) doesn't give any recommendation about where the listeners should be removed. Trying to remove them in the `exit` phase caused other problems, where node removal did not happen as it should. 
After testing the current code, it appears that these listeners are removed correctly.

PR #1352, opened 2024-04-08
Gdb 9971 implement import resource file acitons

boyan-tonchev:
## What:
- Implements the action to reset the import status of a single or multiple resources;
- Implements import single RDF resource;
- Implements interrupt single rdf resource import;
- Hides the "Delete" button when the server tab is active.

## Why:
- This allows users to clear the status of either a single imported resource or all imported resources;
- This gives the user the opportunity to import a single RDF resource;
- This gives the user the opportunity to abort a rdf import if it is not finished yet;
- Server files can't be deleted from the workbench.

## How:
- Added a BE call to reset the status of either specified or all resources when the user clicks the "Reset status" button;
- Added a BE call to import a single RDF resource when the user clicks the "Import" button;
- Added BE call to abort a single rdf resource import when the user clicks on the "Abort" button;
- Added a check to see if the "onDelete" function is passed. If not, the button is hidden. Removed the "onDelete" function from the "import-resource-tree" directive when it's instantiated for server import.

### DIFF
--- a/src/js/angular/import/directives/import-resource-tree.directive.js
+++ b/src/js/angular/import/directives/import-resource-tree.directive.js
@@ -57,6 +57,7 @@ function importResourceTreeDirective($timeout) {
             $scope.filterByFileName = '';
             $scope.STATUS_OPTIONS = STATUS_OPTIONS;
             $scope.selectedByStatus = undefined;
+            $scope.ImportResourceStatus = ImportResourceStatus;
 
             // =========================
             // Public functions
@@ -93,13 +94,17 @@ function importResourceTreeDirective($timeout) {
             };
 
             $scope.onResetStatus = () => {
-                if ($scope.selectedResources && $scope.selectedResources.length > 0) {
-                    $scope.onReset({selectedResources: $scope.selectedResources});
+                if (!$scope.selectedResources) {
+                    return;
+                }
+                const files = $scope.selectedResources.filter((resource) => !resource.isDirectory());
+                if (files.length > 0) {
+                    $scope.onReset({resources: files});
                 }
             };
 
             $scope.resetStatus = (importResource) => {
-                $scope.onReset({selectedResources: [importResource]});
+                $scope.onReset({resources: [importResource]});
             };
 
             $scope.onRemoveResources = () => {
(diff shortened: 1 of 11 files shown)


src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -93,13 +95,17 @@ function importResourceTreeDirective($timeout) {
             };
 
             $scope.onResetStatus = () => {
-                if ($scope.selectedResources && $scope.selectedResources.length > 0) {
-                    $scope.onReset({selectedResources: $scope.selectedResources});
+                if (!$scope.selectedResources) {
+                    return;
+                }
+                const files = $scope.selectedResources.filter((resource) => !resource.isDirectory());
```
svilenvelikov, line 101: I've added additional methods here for getting the files and filenames only. But you can leave this now. We can change it later

src/js/angular/import/templates/import-resource-tree.html
```diff
@@ -149,14 +149,23 @@
                             gdb-tooltip="{{'import.reset.status' | translate}}">
                         <span class="icon-close"></span>
                     </button>
-                    <button class="btn btn-link btn-sm secondary import-resource-action-remove"
+                    <button ng-if="!onDelete" class="btn btn-link btn-sm secondary import-resource-action-remove"
                             ng-click="onDelete({resource: resource})">
                         <em class="icon-trash"></em>
                     </button>
                     <button class="btn btn-primary btn-sm import-resource-action-import"
+                            ng-if="resource.isFile() && resource.importResource.status !== ImportResourceStatus.IMPORTING && resource.importResource.status !== ImportResourceStatus.UPLOADING && resource.importResource.status !== ImportResourceStatus.PENDING && resource.importResource.status !== ImportResourceStatus.INTERRUPTING"
```
svilenvelikov, line 157: I wonder if we can precalculate these after each load of the list. Although the status can't be changed in other means. The idea is to have a property `isImportable` in the resource model for example wich to be precalculated and just refered here.

svilenvelikov (dismissed):

svilenvelikov (approved):

PR #1355, opened 2024-04-09
GDB-9971: Added import status popover.

boyan-tonchev:
What:
When hovering over a resource's info icon, a popover dialog is displayed with more information about the importing process.

Why:
In table view, the user sees only a few details of the import data. Opening the dialog gives the user more information about the import process.

How:
The info icon and popover are implemented as a directive.

### DIFF
--- a/src/i18n/locale-fr.json
+++ b/src/i18n/locale-fr.json
@@ -935,7 +935,6 @@
         "remove.selected": "Supprimer les éléments sélectionnés de la liste",
         "remove.btn": "Supprimer",
         "remove.confirm.msg": "Êtes-vous sûr de vouloir supprimer les fichiers sélectionnés: {{name}}",
-        "last.import.settings": "Paramètres de la dernière importation",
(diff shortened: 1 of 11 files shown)


src/i18n/locale-fr.json
```diff
@@ -1004,6 +1003,15 @@
                 "interrupt_import": "Interrompre l'importation",
                 "abort": "Abandonner"
             }
+        },
+        "status-info": {
+            "last_import_settings": "Paramètres de la dernière importation",
+            "context": "Contexte",
+            "replaced_graphs": "Graphes remplacés",
+            "base_iri": "Base IRI",
```
svilenvelikov, line 1011: is this translated?

svilenvelikov (approved):

src/i18n/locale-fr.json
boyan-tonchev: I took it from another property.

PR #1359, opened 2024-04-10
GDB-9971: Make 'importedOn' or 'modifiedOn' bold depending on which is bigger.

boyan-tonchev:
## What
Changes the display of 'importedOn' and 'modifiedOn' to be bold depending on which is bigger.

## Why
It will be easier for the user to see any changes between the two dates.

## How
Applied the 'font-weight-bold' class to 'importedOn' or 'modifiedOn' depending on which is bigger.


boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

PR #1362, opened 2024-04-11
Gdb 9970 add opportunity to user to change rdf text snipets

boyan-tonchev:
## What
Clicking on the name link of an RDF text resource in the import-resource-tree.directive now opens a dialog containing the resource data, enabling users to update the imported RDF text data.

## Why
This functionality allows users to make modifications to the imported RDF data.

## How
A listener for the "onEditResource" event from the import-resource-tree.directive has been added. When this event occurs, a dialog displaying the resource data is opened, allowing users to update the RDF data and import it into the GraphDB database.

# Additional work
- Fixes unsubscribing functionality of import-resource-status-info directive;
- Removes duplicated import resource status constants.

### DIFF
--- a/src/js/angular/import/controllers.js
+++ b/src/js/angular/import/controllers.js
@@ -4,7 +4,6 @@ import 'angular/rest/import.rest.service';
 import 'angular/rest/upload.rest.service';
 import 'angular/import/import-context.service';
 import 'angular/import/import-view-storage.service';
-import {FILE_STATUS} from "../models/import/file-status";
 import {FileFormats} from "../models/import/file-formats";
 import * as stringUtils from "../utils/string-utils";
 import {FileUtils} from "../utils/file-utils";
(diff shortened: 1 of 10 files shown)


src/js/angular/import/controllers.js
```diff
@@ -355,6 +355,12 @@ importViewModule.controller('ImportViewCtrl', ['$scope', 'toastr', '$interval',
             $scope.setSettingsFor('', withoutChangingSettings, undefined);
         };
 
+        $scope.onEditResource = (resource) => {
```
svilenvelikov, line 358: put jsdoc on public methods at least

svilenvelikov (dismissed):

src/js/angular/import/controllers.js
boyan-tonchev: Done

boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1363, opened 2024-04-12
Gdb 9971 fixses loader functionality

boyan-tonchev:
## What
Implemented a loader to display during the initialization of the import view.

## Why
After integration of the new resource list, the loader had to be changed to work again as expected.

## How
Introduced a 'loader' flag to control the display of the loader. Upon page initialization, the loader is set to true, causing it to be displayed while hiding the import-resource-tree.directive. Once resources are loaded for the first time, the loader is set to false, triggering the display of resources and hiding the loader.

# Additional work
- The 'viewType' parameter was renamed to 'activeTabId' and moved to the import context service;
- The 'resources' parameter was moved to the import context service.

### DIFF
--- a/src/js/angular/import/controllers.js
+++ b/src/js/angular/import/controllers.js
@@ -75,7 +75,6 @@ importViewModule.controller('ImportViewCtrl', ['$scope', 'toastr', '$interval', 
         $scope.fileFormatsHuman = FileFormats.getFileFormatsHuman() + $translate.instant('import.gz.zip');
         $scope.textFileFormatsHuman = FileFormats.getTextFileFormatsHuman();
         $scope.maxUploadFileSizeMB = 0;
-        $scope.resources = new ImportResourceTreeElement();
         $scope.SORTING_TYPES = SortingType;
 
         // =========================
@@ -247,7 +246,7 @@ importViewModule.controller('ImportViewCtrl', ['$scope', 'toastr', '$interval', 
          * @return {string[]}
          */
         $scope.getSelectedFiles = () => {
-            return $scope.resources.getAllSelectedFilesNames();
+            return ImportContextService.getResources().getAllSelectedFilesNames();
         };
 
         $scope.importSelected = (overrideSettings) => {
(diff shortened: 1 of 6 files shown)


src/js/angular/import/controllers.js
```diff
@@ -1030,41 +1030,44 @@ importViewModule.controller('TabCtrl', ['$scope', '$location', 'ImportViewStorag
...
-    };
-    init();
+    subscriptions.push(ImportContextService.onFilesUpdated(onFilesUpdated));
+
+    const removeAllListeners = () => subscriptions.forEach((subscription) => subscription());
+
+    $scope.$on('$destroy', removeAllListeners);
+
+    // Updates the active tab id from tha url.
+    ImportContextService.updateActiveTabId($location.hash() || TABS.USER);
```
svilenvelikov, line 1070: Let's try to keep things tidy. There should not be any expression outside of functions.

src/js/angular/import/controllers.js
```diff
@@ -31,16 +32,11 @@ const modules = [
 
 const importViewModule = angular.module('graphdb.framework.impex.import.controllers', modules);
 
-const OPERATION = {
+export const OPERATION = {
```
svilenvelikov, line 35: Why this is exported? As far as I can see, it is used only here?

boyan-tonchev: I forgot to remove this constant. The method where it is used has to use 'TABS'. This is the reason the tests failed.

src/js/angular/import/controllers.js
boyan-tonchev: Done

src/js/angular/import/controllers.js
boyan-tonchev: Fixed.

src/js/angular/import/controllers.js
svilenvelikov: Actually it's not done. See how the other controllers are made. Move all initialization logic inside the init function which you actually removed here I think. Then just call the init.

boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

boyan-tonchev:
Jenkins retry build

PR #1373, opened 2024-04-17
GDB-9304 - Translate labels in Google chart config and Pivot table

DesiBorisova:
## What?
When the user changes the language while in the SPARQL view, the page will require permission to reload.

## Why?
The configs are loaded with the language passed in during initialization, it cannot be changed afterwards. Hence, the page needs to reload in order to have the right labels in the config window.

## How?
I handle the dialog open and page reload in the `yasgui-component-directive`. If the user is on the SPARQL editor page, the dialog will appear.

**Dependency**: this MR depends on changes in the YASGUI component with MR https://github.com/Ontotext-AD/ontotext-yasgui/pull/265

### DIFF
--- a/src/js/angular/core/directives/languageselector/language-selector.directive.js
+++ b/src/js/angular/core/directives/languageselector/language-selector.directive.js
@@ -1,14 +1,15 @@
 import 'angular/utils/local-storage-adapter';
+import 'angular/core/services/event-emitter-service';
 angular
     .module('graphdb.framework.core.directives.languageselector.languageselector', [
-        'graphdb.framework.utils.localstorageadapter'
+        'graphdb.framework.utils.localstorageadapter', 'graphdb.framework.utils.event-emitter-service'
     ])
     .directive('languageSelector', languageSelector)
     .service('$languageService', [languageService]);
 
-languageSelector.$inject = ['$translate', 'LocalStorageAdapter', 'LSKeys', '$languageService'];
+languageSelector.$inject = ['$translate', 'LocalStorageAdapter', 'LSKeys', '$languageService', 'EventEmitterService'];
 
-function languageSelector($translate, LocalStorageAdapter, LSKeys, $languageService) {
+function languageSelector($translate, LocalStorageAdapter, LSKeys, $languageService, eventEmitterService) {
     return {
         templateUrl: 'js/angular/core/directives/languageselector/templates/languageSelector.html',
         restrict: 'E',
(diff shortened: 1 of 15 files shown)


src/js/angular/core/directives/languageselector/language-selector.directive.js
```diff
@@ -31,12 +33,44 @@ function languageSelector($translate, LocalStorageAdapter, LSKeys, $languageServ
...
+            }
+
             $scope.changeLanguage = function (lang) {
-                $scope.selectedLang = lang
-                $translate.use(lang.key);
-                LocalStorageAdapter.set(LSKeys.PREFERRED_LANG, lang.key);
-                $languageService.setLanguage($scope.selectedLang.key);
-                $scope.$broadcast('language-changed', {locale: lang.key});
+                const yasgui = YasguiComponentDirectiveUtil.getOntotextYasguiElement(QUERY_EDITOR_ID);
+                const promise = yasgui ? yasgui.yasguiComponent.getActivePluginTab() : Promise.resolve(null);
```
svilenvelikov, line 42: Should you do `Promise.reject()` if yasgui is not found on the page? This way you can omit some of the checks you do in the handler below.

src/js/angular/core/directives/languageselector/language-selector.directive.js
```diff
@@ -31,12 +33,44 @@ function languageSelector($translate, LocalStorageAdapter, LSKeys, $languageServ
...
+                        $languageService.setLanguage($scope.selectedLang.key);
+                        $scope.$broadcast('language-changed', {locale: lang.key});
+                    }
+                });
+
+                if (!yasgui && $languageService.getLanguage() !== lang.key) {
+                    $scope.selectedLang = lang;
+                    $translate.use(lang.key);
+                    LocalStorageAdapter.set(LSKeys.PREFERRED_LANG, lang.key);
+                    $languageService.setLanguage($scope.selectedLang.key);
```
svilenvelikov, line 72: Is there a reason for repeating these 4 lines three times? Think of a way to improve this!

src/js/angular/core/directives/languageselector/language-selector.directive.js
DesiBorisova: OK. I made some changes to this.

boyan-tonchev: Why is this check here? I think it would be better placed in the yasgui-component.directive. My suggestion is to fire an event via EventEmitter here, and in the yasgui directive, cancel it if necessary. This approach was used when a repository is changed.

src/js/angular/core/directives/languageselector/language-selector.directive.js
```diff
@@ -31,12 +32,29 @@ function languageSelector($translate, LocalStorageAdapter, LSKeys, $languageServ
...
                 $translate.use(lang.key);
                 LocalStorageAdapter.set(LSKeys.PREFERRED_LANG, lang.key);
                 $languageService.setLanguage($scope.selectedLang.key);
-                $scope.$broadcast('language-changed', {locale: lang.key});
+            }
+
+            $scope.changeLanguage = function (lang) {
+                const eventData = {locale: lang.key, cancel: false};
+                eventEmitterService.emit('language-changed', eventData, (eventData) => {
+                    if (eventData && !eventData.cancel && $languageService.getLanguage() !== lang.key) {
```
boyan-tonchev, line 45: The check $languageService.getLanguage() !== lang.key is the same in both cases. It is better if this check is made before the 'language-changed' event is emitted; this will prevent unnecessary checks in the yasgui-component directive.

DesiBorisova: OK. I will extract it.

boyan-tonchev (approved):

src/js/angular/core/directives/yasgui-component/yasgui-component.directive.js
```diff
@@ -393,6 +393,7 @@ function yasguiComponentDirective(
                         return;
                     }
                 }
+                $scope.language = $languageService.getLanguage();
```
svilenvelikov, line 396: You must declare the language variable in the variables section on the top.

teodossidossev:
lgtm

PR #1376, opened 2024-04-18
GDB-10040-Update-server-responses-related-to-import-functionality

boyan-tonchev:
## What
Updated models to reflect changes in the backend.

## Why
The backend API now returns additional information about import resources.

## How
The model has been updated.

# Additional work
- Fixes conditions when importedOn and modifiedOn have to be bolded;
- Changed icons based on the type of import resource.

### DIFF
--- a/src/js/angular/rest/mappers/import-mapper.js
+++ b/src/js/angular/rest/mappers/import-mapper.js
@@ -1,6 +1,7 @@
 import {ImportResourceTreeElement} from "../../models/import/import-resource-tree-element";
 import {ImportResource} from "../../models/import/import-resource";
 import {ImportResourceStatus} from "../../models/import/import-resource-status";
+import {ImportResourceType} from "../../models/import/import-resource-type";
 
 export const toImportResource = (importResourcesServerData) => {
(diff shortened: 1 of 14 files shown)


src/js/angular/rest/mappers/import-mapper.js
```diff
@@ -63,8 +71,30 @@ const addResourceToTree = (root, resource) => {
...
+        importServerResource.iconClass = getIconClass(resource);
+        return importServerResource;
     } else {
         resourceParent.importResource = resource;
+        resourceParent.iconClass = getIconClass(resource);
+        return resourceParent;
+    }
+};
+
+const getIconClass = (resource) => {
```
svilenvelikov, line 83: Can you just create a simple mapping of resource types to icons instead of this function.

boyan-tonchev: Done

svilenvelikov: I didn't mean to use a Map here but just a plain object. Anyway, this will do the job as well

svilenvelikov (approved):

PR #1378, opened 2024-04-19
GDB-10161 Redesign cluster configuration panel

svilenvelikov:
## What
Redesign cluster configuration panel.

## Why
Hide the delete cluster button inside and simplify the main cluster-management.controller.js by extracting the cluster configuration in separate component.

## How
Extract cluster configuration in separate component and use it in the cluster view side panel. Introduced cluster-view-context.service.js for storing some context variables and allow inter-component communication.

### DIFF
--- a/src/js/angular/clustermanagement/directives/cluster-configuration.directive.js
+++ b/src/js/angular/clustermanagement/directives/cluster-configuration.directive.js
@@ -1,0 +1,64 @@
+import {DELETE_CLUSTER, UPDATE_CLUSTER} from "../events";
+
+const modules = [];
+
+angular
+    .module('graphdb.framework.clustermanagement.directives.cluster-configuration', modules)
+    .directive('clusterConfiguration', ClusterConfiguration);
+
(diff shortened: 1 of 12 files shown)


src/js/angular/clustermanagement/directives/cluster-configuration.directive.js
```diff
@@ -0,0 +1,61 @@
...
+            clusterModel: '=',
+            clusterConfiguration: '='
+        },
+        link: ($scope) => {
+
+            $scope.closeClusterConfigurationPanel = () => {
+                ClusterViewContextService.hideClusterConfigurationPanel();
+            };
+
+            $scope.isAdmin = () => {
```
boyan-tonchev, line 26: IMO this can be passed as argument to directive or be calculated once when directive is created. We can't change admin when the directive is displayed.

boyan-tonchev (approved):

src/js/angular/clustermanagement/directives/cluster-configuration.directive.js
svilenvelikov: I will check this out

boyan-tonchev (dismissed):

boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

PR #1380, opened 2024-04-23
GDB-6329 - Hide toastr and info messages when license invalid

DesiBorisova:
## What?
When the license is invalid in the `Class hierarchy view` and the `Plugins view`, the user will see the 'license expired' message. No toastr messages or warnings will appear.

## Why?
Certain API calls would be made even though the license is invalid and the BE would return an error. The resulting message would appear in the UI alongside the 'license expired' message.
The `Plugins view` contained an info message, which remained also alongside the 'license expired' message.

## How?
I added license validity checks before the API call and in the template.


DesiBorisova:
Jenkins retry build

src/js/angular/core/templates/core-errors.html
```diff
@@ -32,7 +32,7 @@
                 </div>
                 <ul class="list-group limit-height clearfix two-columns repos" ng-mouseleave="hidePopoverForRepo($event)">
                     <li ng-repeat="repository in getAccessibleRepositories() | orderBy: ['type === \'system\'', 'location', 'id']"
-                        ng-if="repository.id !== getActiveRepository() || repository.location !== getActiveRepositoryObject().location"
+                        ng-if="isLicenseValid() && repository.id !== getActiveRepository() || repository.location !== getActiveRepositoryObject().location"
```
svilenvelikov, line 35: test this again

test/namespaces/controllers.spec.js
```diff
@@ -23,9 +23,10 @@ describe('=> NamespacesCtrl tests', function () {
         toastr,
         httpGetNamespaces,
         modalInstance,
-        $translate;
+        $translate,
+        $licenseService;
 
-    beforeEach(angular.mock.inject(function (_$httpBackend_, _$repositories_, _RDF4JRepositoriesRestService_, _toastr_, _$location_, _$controller_, _$window_, _$timeout_, $rootScope, $q, _$translate_) {
+    beforeEach(angular.mock.inject(function (_$httpBackend_, _$repositories_, _RDF4JRepositoriesRestService_, _toastr_, _$location_, _$controller_, _$window_, _$timeout_, $rootScope, $q, _$translate_, _$licenseService_) {
```
svilenvelikov, line 29: The unit tests are not run on CI anymore and will be removed

yordanalexandrov:
There are some other places where the toastr notification is shown or an action which requires a valid license leads to an error.
* Import screen would allow to choose files and after that show an error that the license is not valid.
* Graph overview (Explore) would allow to initialte Clear repository and return an error.
* TTYG would show an error toaster if the licence is bad.
* Also although not tested the Cluster view might throw an error.

src/js/angular/core/services/license.service.js
```diff
@@ -23,6 +23,8 @@ function licenseService($rootScope, LicenseRestService, $translate) {
                 that.loadingLicense = false;
                 updateProductType(that.license);
             });
+        }).finally(function () {
+            $rootScope.$broadcast('license.set');
```
svilenvelikov, line 27: Create a file /core/global-events.js and put an exported constant LICENSE_SET there. Then reuse it. See this for an example https://github.com/Ontotext-AD/graphdb-workbench/blob/master/src/js/angular/clustermanagement/events.js.

EDIT: 
Above should be done only if actually this broadcast is proper here, which I think is not the case. 
Given the code in the method before this line, it's obsious that the license might actually not be set when the finally clause is hit.
Another issue with this code is that the service is bound to the $rootScope which is bad, because services should be unaware of the UI. In other words, if this broadcast should be executed, then is should be done in the license controller most likely.

svilenvelikov (changes requested):

test/namespaces/controllers.spec.js
DesiBorisova: Reverted

src/js/angular/explore/controllers.js
```diff
@@ -579,7 +586,7 @@ function EditResourceCtrl($scope, $http, $location, toastr, $repositories, $uibM
     $scope.$watch(function () {
         return $repositories.getActiveRepository();
     }, function () {
-        if ($scope.activeRepository()) {
+        if ($licenseService.isLicenseValid() && $scope.activeRepository()) {
```
svilenvelikov, line 589: Why? Aren't you doing the check in the beginning of the getClassInstancesDetails method already?

src/js/angular/graphexplore/controllers/graphs-visualizations.controller.js
```diff
@@ -1226,7 +1226,7 @@ function GraphsVisualizationsCtrl(
     }
 
     function checkAutocompleteStatus() {
-        if ($licenseService.isLicenseValid()) {
+        if ($scope.isLicenseValid()) {
```
svilenvelikov, line 1229: use the service directly and not this local proxy method

src/js/angular/graphexplore/controllers/rdf-class-hierarchy.controller.js
```diff
@@ -57,16 +57,16 @@ function RdfClassHierarchyCtlr($scope, $rootScope, $location, $repositories, $li
     let selectedGraph = allGraphs;
 
     const initView = function () {
-        if (!$scope.getActiveRepository()) {
+        if (!$scope.getActiveRepository() && !$licenseService.isLicenseValid()) {
```
svilenvelikov, line 60: Are you sure that the conjunction should be AND and not OR here?

src/js/angular/graphexplore/controllers/rdf-class-hierarchy.controller.js
```diff
@@ -448,10 +452,12 @@ function RdfClassHierarchyCtlr($scope, $rootScope, $location, $repositories, $li
     }
 
     function getClassHierarchyData() {
-
+        if (!$licenseService.isLicenseValid()) {
```
svilenvelikov, line 455: I wonder why this was moved from the its previous place below? What is the difference? As far as I can see, the refreshDiagramExternalElements method doesn't do any requests.

src/js/angular/jdbc/controllers.js
```diff
@@ -32,7 +32,8 @@ function JdbcListCtrl($scope, $repositories, JdbcRestService, toastr, ModalServi
     $scope.getSqlConfigurations = function () {
         // Only do this if there is an active repo that isn't an Ontop or FedX repo.
         // Ontop and FedX repos don't support JDBC.
-        if ($repositories.getActiveRepository()
+        if ($scope.isLicenseValid()
```
svilenvelikov, line 35: Inject the license service and use the method from it. Don't rely on the inherited method from the parent scope. Also replace the other similar usages in this file.

src/pages/jdbc-create.html
```diff
@@ -9,205 +9,211 @@ <h1>
...
     </h1>
-    <div core-errors></div>
+    <div core-errors license></div>
     <div system-repo-warning></div>
     <div class="alert alert-danger" ng-show="repositoryError">
         <p>{{'not.usable.active.repo.error' | translate}}</p>
 
         <p>{{repositoryError}}</p>
     </div>
+    <div ng-if="isLicenseValid()">
```
svilenvelikov, line 19: As I can see, this in fact calls the isLicenseValid from the main controller. This is bad. Create a method in the local scope which delegates to the license service instead.

src/pages/monitor/queries.html
```diff
@@ -14,10 +14,10 @@ <h1>
 		   popover-append-to-body="true"><span class="icon-info text-tertiary"></span></span>
 	</h1>
 
-    <div core-errors ontop></div>
+    <div core-errors ontop license></div>
 
 	<div id="wb-monitoringQueries" class="ot-owlim-monitoring-queries" ng-if="!isActiveRepoOntopType()">
-		<div ng-show="getActiveRepository()">
+		<div ng-show="getActiveRepository() && isLicenseValid()">
```
svilenvelikov, line 20: Same as in the JDBC view

src/pages/namespaces.html
```diff
@@ -15,8 +15,8 @@ <h1>
 
 <div class="ot-loader ot-main-loader" onto-loader size="50" ng-show="loader"></div>
 <div id="wb-namespaces" ng-hide="loader">
-    <div core-errors></div>
-    <div ng-show="getActiveRepository()">
+    <div core-errors license></div>
+    <div ng-show="getActiveRepository() && isLicenseValid()">
```
svilenvelikov, line 19: Same as in the JDBC view

src/pages/plugins.html
```diff
@@ -53,7 +53,7 @@ <h1>
         </tr>
         </tbody>
     </table>
-    <div class="alert alert-info no-plugins-match-alert" ng-show="displayedPlugins.length === 0">
+    <div class="alert alert-info no-plugins-match-alert" ng-show="isLicenseValid() && displayedPlugins.length === 0">
```
svilenvelikov, line 56: Same as in the JDBC view

src/pages/sparql-editor.html
```diff
@@ -10,9 +10,9 @@ <h1 id="sparql-query-update-title-label">
               popover-append-to-body="true"><span class="icon-info text-tertiary"></span></span>
     </h1>
 
-    <div core-errors></div>
+    <div core-errors license></div>
 
     <yasgui-component id="query-editor" yasgui-config="yasguiConfig"
-                      ng-if="yasguiConfig && getActiveRepositoryNoError()"
+                      ng-if="yasguiConfig && getActiveRepositoryNoError() && isLicenseValid()"
```
svilenvelikov, line 16: Same as in JDBC

svilenvelikov:
> There are some other places where the toastr notification is shown or an action which requires a valid license leads to an error.
> 
> * Import screen would allow to choose files and after that show an error that the license is not valid.
> * Graph overview (Explore) would allow to initialte Clear repository and return an error.
> * TTYG would show an error toaster if the licence is bad.
> * Also although not tested the Cluster view might throw an error.

These are valid notes also and should be fixed!

src/js/angular/core/directives.js
```diff
@@ -333,7 +333,7 @@ function searchResourceInput($location, toastr, ClassInstanceDetailsService, Aut
             };
 
             $scope.$watch('namespacespromise', function () {
-                if (angular.isDefined($scope.namespacespromise)) {
+                if (angular.isDefined($scope.namespacespromise) && $licenseService.isLicenseValid()) {
```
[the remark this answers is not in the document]
DesiBorisova: Fair point. I will re-test and see if needed

src/js/angular/explore/controllers.js
```diff
@@ -530,7 +534,7 @@ function EditResourceCtrl($scope, $http, $location, toastr, $repositories, $uibM
     };
     $scope.newResource = false;
     $scope.datatypeOptions = StatementsService.getDatatypeOptions();
-
+    $scope.isLicenseValid = $licenseService.isLicenseValid();
```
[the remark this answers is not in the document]
DesiBorisova: Should be removed, yes

src/js/angular/explore/controllers.js
DesiBorisova: Yes, I'll fix the duplication.

src/js/angular/graphexplore/controllers/rdf-class-hierarchy.controller.js
DesiBorisova: Should be OR. Will fix.

src/js/angular/graphexplore/controllers/rdf-class-hierarchy.controller.js
DesiBorisova: I agree. I'll move it back.

src/js/angular/jdbc/controllers.js
DesiBorisova: OK, will fix

src/js/angular/graphexplore/directives/rdf-class-hierarchy.directive.js
```diff
@@ -124,6 +124,9 @@ function classHierarchyDirective($rootScope, $location, GraphDataRestService, $w
 
                 }).error(function (response) {
                 $rootScope.loader = false;
+                if (!$licenseService.isLicenseValid()) {
```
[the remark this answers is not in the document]
DesiBorisova: I'll move it back to the `if` above

src/js/angular/chatgpt/controllers.js
```diff
@@ -8,11 +8,14 @@ angular
...
     .controller('ChatGptSettingsCtrl', ChatGptSettingsCtrl);
 
-ChatGptCtrl.$inject = ['$scope', '$http', '$timeout', '$translate', '$uibModal', '$repositories', 'toastr', 'ModalService', 'LocalStorageAdapter'];
+ChatGptCtrl.$inject = ['$scope', '$http', '$timeout', '$translate', '$uibModal', '$repositories', 'toastr', 'ModalService', 'LocalStorageAdapter', '$licenseService'];
 
 const CHATGPTRETRIEVAL_ENDPOINT = 'rest/chat/retrieval';
 
-function ChatGptCtrl($scope, $http, $timeout, $translate, $uibModal, $repositories, toastr, ModalService, LocalStorageAdapter) {
+function ChatGptCtrl($scope, $http, $timeout, $translate, $uibModal, $repositories, toastr, ModalService, LocalStorageAdapter, $licenseService) {
+    $scope.isLicenseValid = function () {
```
svilenvelikov, line 16: What's the reason to put this public function on top mixing it with the private once?

DesiBorisova: Will move next to public

PR #1381, opened 2024-04-23
GDB-10170: Issue with long names in the import-resource-tree component

boyan-tonchev:
## What
 In the User Data tab, the Import RDF files option allows files with extremely long file names, up to 255 characters, to be imported. As a result, the overview elements overlap each other, and each checkbox cannot be clicked.

## Why
 When the name is long and there is no space for it to break into a new line, the import-resource-tree component becomes larger than the workbench page, causing a horizontal scrollbar to appear. Moving the scrollbar to the right causes the checkboxes to move under the right panel menu. While they are visible, they are not clickable.

## How
 Added a property to the cell where the name is displayed that allows the browser to break the word anywhere.

### DIFF
--- a/src/css/import-resource-tree.css
+++ b/src/css/import-resource-tree.css
@@ -62,6 +62,7 @@
     display: flex;
     align-items: center;
     line-height: 18px;
+    word-break: break-all;
 }
 
 .import-resource-tree .import-resource-table .cell-title em {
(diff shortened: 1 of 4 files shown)


boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

src/css/import-resource-tree.css
```diff
@@ -62,6 +62,7 @@
     display: flex;
     align-items: center;
     line-height: 18px;
+    word-break: break-all;
```
svilenvelikov, line 65: we should apply ellipsis in the middle, same as for the context

boyan-tonchev: Done

boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1385, opened 2024-04-26
GDB-10159: Redesign the cluster view legend

boyan-tonchev:
## What
Redesigned the cluster view legend.

## Why
Current legend implementation doesn’t have enough surface area so that all the information represented inside it can be clearly visible and perceived by the user without any additional actions needed after opening it.

## How
- A new "ClusterLegend" component has been implemented;
- Old legend has been replaced with the new component.

### DIFF
--- a/src/css/clustermanagement.css
+++ b/src/css/clustermanagement.css
@@ -256,3 +256,7 @@ div.nodetooltip {
 .advanced-options-toggle {
     margin-right: 6px;
 }
+
+.toggle-legend-btn .btn {
+    text-decoration: none;
+}
--- a/src/js/angular/clustermanagement/directives/cluster-legend.directive.js
+++ b/src/js/angular/clustermanagement/directives/cluster-legend.directive.js
@@ -1,0 +1,280 @@
+import {CLUSTER_MANAGEMENT_CONSTANTS} from "../../models/clustermanagement/cluster-management-constants";
+import * as CDS from "../services/cluster-drawing.service";
+import {LinkState} from "../../models/clustermanagement/states";
+import {ARROW_CONFIG} from "../services/cluster-drawing.service";
+
+const modules = [];
+angular.module('graphdb.framework.clustermanagement.directives.cluster-legend', modules)
+    .directive('clusterLegend', clusterLegend);
+
+clusterLegend.$inject = ['$rootScope', 'ClusterViewContextService', '$translate', '$document'];
+
+function clusterLegend($rootScope, ClusterViewContextService, $translate, $document) {
+    return {
+        restrict: 'E',
+        templateUrl: 'js/angular/clustermanagement/templates/cluster-legend.html',
+        scope: {},
+        link: ($scope) => {
+
+            // =========================
+            // Public variables
+            // =========================
+
+            $scope.hideLegend = true;
+
+            // =========================
+            // Private variables
+            // =========================
+
+            const subscriptions = [];
+            let legendGroup;
+            let legendHeight;
+            let legendWidth;
+
+            // =========================
+            // Public functions
+            // =========================
+
+            $scope.toggleLegend = () => {
+                $scope.hideLegend = !$scope.hideLegend;
+                legendGroup.classed('hidden', $scope.hideLegend);
+            };
+
+            // =========================
+            // Private functions
+            // =========================
+
+            const init = () => {
+                const clusterViewD3Container = ClusterViewContextService.getClusterViewD3Container();
+                createLegendGroup(clusterViewD3Container);
+                createLegendBackground();
+                addNodeStates(legendGroup);
+                addSyncStatus();
+                addLinks();
+                updateLegendDimensions();
+                updateLegendBackground();
+            };
+
+            const addNodeStates = (legendGroup) => {
+                addTitle(legendGroup, 'cluster_management.cluster_graphical_view.legend_title_node_state');
+                const legendNodeData = legendGroup
+                    .selectAll('.node-group')
+                    .data(CLUSTER_MANAGEMENT_CONSTANTS.getLegendNodes());
+                const legendNodesElements = CDS.createNodes(legendNodeData, CLUSTER_MANAGEMENT_CONSTANTS.SVG_NODE_WIDTH, true);
+                CDS.updateNodes(legendNodesElements);
+                const LABEL_Y_POSITION = 3;
+                legendNodesElements
+                    .attr('transform', function (d, i) {
+                        updateLegendDimensions();
+                        return `translate(${CLUSTER_MANAGEMENT_CONSTANTS.LEGEND_ITEM_PADDING_LEFT}, ${legendHeight + CLUSTER_MANAGEMENT_CONSTANTS.LEGEND_ITEM_PADDING_TOP})`;
+                    })
+                    .append('text')
+                    .style('font-size', CLUSTER_MANAGEMENT_CONSTANTS.LEGEND_ITEM_FONT_SIZE)
+                    .style('line-height', CLUSTER_MANAGEMENT_CONSTANTS.LEGEND_ITEM_LINE_HEIGHT)
+                    .style('font-weight', CLUSTER_MANAGEMENT_CONSTANTS.LEGEND_ITEM_FONT_WEIGHT)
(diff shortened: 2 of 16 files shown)


src/js/angular/clustermanagement/directives/cluster-legend.directive.js
```diff
@@ -0,0 +1,153 @@
+import {ClusterSvgService, HexagonConfig, LineConfig} from "../services/cluster-svg.service";
+
+const modules = [];
+angular.module('graphdb.framework.clustermanagement.directives.cluster-legend', modules)
+    .directive('clusterLegend', clusterLegend);
+
+clusterLegend.$inject = ['$rootScope', '$sce'];
+
+function clusterLegend($rootScope, $sce) {
```
svilenvelikov, line 9: Why do you need this rootScope here?

src/js/angular/clustermanagement/directives/cluster-legend.directive.js
```diff
@@ -0,0 +1,153 @@
...
+    return {
+        restrict: 'E',
+        templateUrl: 'js/angular/clustermanagement/templates/cluster-legend.html',
+        scope: {
+            clusterViewSelector: '='
```
svilenvelikov, line 14: You don't use this clusterViewSelector

src/js/angular/clustermanagement/directives/cluster-legend.directive.js
```diff
@@ -0,0 +1,153 @@
...
+        },
+        link: ($scope, element, attrs) => {
```
svilenvelikov, line 16: attrs is not used

src/js/angular/clustermanagement/directives/cluster-legend.directive.js
```diff
@@ -0,0 +1,153 @@
...
+            // =========================
+            const getNodeStates = () => {
+                return [
+                    {
+                        labelKey: 'cluster_management.cluster_graphical_view.node_state_leader',
+                        svg: $sce.trustAsHtml(ClusterSvgService.getLeaderSvg(new HexagonConfig().setSize(30).setClasses('node leader')))
+                    },
+                    {
+                        labelKey: 'cluster_management.cluster_graphical_view.node_state_follower',
+                        svg: $sce.trustAsHtml(ClusterSvgService.getLeaderSvg(new HexagonConfig().setSize(30).setClasses('node follower')))
```
yordanalexandrov, line 51: Why is this called `getLeaderSvg`, are svgs differernt for leader, follower or out of sync nodes?

src/js/angular/clustermanagement/services/cluster-svg.service.js
```diff
@@ -0,0 +1,100 @@
...
+
+    setColor(color) {
+        this.color = color;
+        return this;
+    }
+}
+
+export class ClusterSvgService {
+    static getLeaderSvg(config = new HexagonConfig()) {
+        return `<svg width="${config.size}" height="${config.size}" viewBox="-25 -25 50 50">
```
svilenvelikov, line 67: And now we have two dofferent ways to draw things in the cluster view. How is this better than using just d3.js?

src/js/angular/clustermanagement/services/cluster-svg.service.js
```diff
@@ -0,0 +1,100 @@
+export class LineConfig {
```
svilenvelikov, line 1: Move each class in separate file

src/js/angular/clustermanagement/templates/cluster-legend.html
```diff
@@ -0,0 +1,47 @@
+<span class="cluster-legend-wrapper">
```
svilenvelikov, line 1: It's incorrect to have block level elements inside inline one

src/js/angular/clustermanagement/directives/cluster-legend.directive.js
boyan-tonchev: I renamed it to getHexagonSvg, but it seems to have been lost. I'll rename it again.

src/js/angular/clustermanagement/directives/cluster-legend.directive.js
```diff
@@ -0,0 +1,153 @@
...
+                    }
+                ];
+            };
+
+            const getSyncStatuses = () => {
+                return [
+                    {
+                        labelKey: 'cluster_management.cluster_graphical_view.node_state_leader',
+                        classes: 'icon-any',
+                        icon: '\ue935'
```
yordanalexandrov, line 65: Would it not be better to have all those node state based `class` and `icon` combinations extracted somewhere as I think they would be reused in the cluster visualization as well?

yordanalexandrov: And also perhaps the labels can be composed with an enum concatenation at the end, so that not everything is hardcoded

src/js/angular/clustermanagement/templates/cluster-legend.html
```diff
@@ -0,0 +1,47 @@
...
+    <div class="cluster-legend-container">
```
svilenvelikov, line 2: What's the difference between cluster-legend-wrapper and cluster-legend-container?! The former can be just cluster-legend and the latter cluster-nodes or whatever

src/js/angular/clustermanagement/templates/cluster-legend.html
```diff
@@ -0,0 +1,47 @@
...
+          <div class="node-state" ng-repeat="nodeState in nodeStates">
+                <div class="row">
+                    <div class="row-icon" ng-bind-html="nodeState.svg"></div>
+                    <div class="row-value">{{ nodeState.labelKey | translate }}</div>
+                </div>
+            </div>
+        </div>
+
+
+        <div class="sync-statuses-container">
```
svilenvelikov, line 17: These are also node states I think

src/js/angular/clustermanagement/directives/cluster-legend.directive.js
yordanalexandrov: It might be `getNodeSvg` to be more readable. After all it is the Cluster SVG Service

src/css/clustermanagement.css
```diff
@@ -253,3 +253,73 @@ div.nodetooltip {
 .advanced-options-toggle {
   margin-right: 6px;
 }
+
+/**----------------------------------------------*/
```
svilenvelikov, line 257: I'll put these here because I don't want to search for the appropriate place for them:
* Nodes in the legend are smaller than in the mockup
* Headings of the sections are too big
* Font color is different
* We don't have leader and follower icons anymore
* Color of the lines representing the links should be the secondary color and not the tertiary (the green one) which you've used
* The thikness of the lines should be 4px probaly
* The recovering link type is missing
* The sync state rows are quite bumped into each other
* The legend panel is positioned too low and covers the button that opens it a bit.

src/js/angular/clustermanagement/directives/cluster-legend.directive.js
```diff
@@ -0,0 +1,153 @@
...
+            const legendElement = element.find('.cluster-legend-container');
+
+            // =========================
+            // Public functions
+            // =========================
+            $scope.toggleLegend = () => {
+                legendElement.toggleClass('hidden');
+            };
+
+            $scope.closeDialog = () => {
```
svilenvelikov, line 36: It'd be better if there is a way to close the leged with the [esc] key.

svilenvelikov (changes requested):

src/css/clustermanagement.css
svilenvelikov: And one more thing here. Cluster node hexagons now should have rounded corners. Look mockups and the notes there carefully.

src/js/angular/clustermanagement/directives/cluster-legend.directive.js
boyan-tonchev: Done

src/js/angular/clustermanagement/directives/cluster-legend.directive.js
boyan-tonchev: Done

PR #1386, opened 2024-04-26
GDB-9971: fixes check box selection

boyan-tonchev:
## What
 The checkbox of resources is frequently not selected/unselected.

## Why
 The view is updated every few seconds, and sometimes when the user clicks on a checkbox at the same time as the model is updated, the user's click is skipped. The import resource tree directive uses ImportResourceTreeElement for the model, which holds all the necessary information for visualization and a reference to the ImportResource returned from the backend. The problem occurred because in the tree, when the user modified a select element, the entire tree was replaced due to the server response, and the selection was ignored.

## How
 When the backend returns a response in the import controller, this response is transformed into a list of ImportResources. Then, an ImportResourceTreeElement tree is created and passed to the import resource tree directive. However, this is not the correct place for tree creation. Therefore, the creation of the tree was moved to the import resource tree directive. The controller now only maps the server response to a list of import resources, while the tree is created and updated in the directive.


src/js/angular/import/controllers/import-view.controller.js
```diff
@@ -86,12 +85,14 @@ importViewModule.controller('ImportViewCtrl', ['$scope', 'toastr', '$interval',
         $scope.toTitleCase = (str) => stringUtils.toTitleCase(str);
 
         $scope.pollList = () => {
-            listPollingHandler = $interval(() => {
-                // Skip iteration if we are updating something
-                if (!$scope.updating) {
-                    $scope.updateList(false);
-                }
-            }, LIST_POLLING_INTERVAL);
+            if (!listPollingHandler) {
```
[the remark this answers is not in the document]
boyan-tonchev: I will check it again.

svilenvelikov:
I've tested this and unfortunately it doesn't seam to fix the issue with the checkboxes. In addition this seems to break the the behavior for disabling the resource operation buttons when a checkbox is selected. Probably we don't have a test for this or it's in the flaky suite.

boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

src/js/angular/import/services/import-resource-tree.service.js
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /**
+     * Creates directory path (if no created yet) and returns the inner parent .
+     * @param {ImportResourceTreeElement} importServerResource - the root parent directory.
+     * @param {string[]} directoryPath - the list with directory names that have to be created.
+     *
+     * @return {ImportResourceTreeElement} the inner parent.
+     */
+    static getOrCreateParent(importServerResource, directoryPath) {
```
svilenvelikov, line 86: Why `importServerResource`? Isn't this supposed to be generic component?

src/js/angular/models/import/import-resource-tree-element.js
```diff
@@ -100,6 +100,20 @@ export class ImportResourceTreeElement {
         throw new Error('Unsupported resource type!');
     }
 
+    removeResource(importServerResource) {
```
svilenvelikov, line 103: Same here. Rename argument to importResource or just resource!

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -305,8 +305,37 @@ function importResourceTreeDirective($timeout, ImportContextService) {
             // Subscription handlers
             // =========================
 
-            const onResourcesUpdatedHandler = (resources) => {
-                $scope.resources = resources;
+            const onResourcesUpdatedHandler = (newResources = []) => {
+                const isUserImport = TABS.USER === ImportContextService.getActiveTabId();
+                if ($scope.resources.isEmpty()) {
+                    $scope.resources = ImportResourceTreeService.toImportResourceTree(_.cloneDeep(newResources), isUserImport);
```
svilenvelikov, line 311: I don't get it why the cloneDeep was needed here? This onResourcesUpdatedHandler is invoked when resources are loaded from the backend and on the initial bootstrap where an empty array is passed. So, what's the reason to clone a backend response before converting it to tree model? Do you expect someone to work with the raw response instead with the tree?

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -305,8 +305,37 @@ function importResourceTreeDirective($timeout, ImportContextService) {
...
+                    if (isUserImport) {
+                        $scope.sortedBy = SortingType.MODIFIED;
+                        sortResources();
+                    }
+                } else {
+                    // Removes missing files.
+                    $scope.resources.toList().forEach((resource) => {
```
svilenvelikov, line 318: This can be moved in the resources model

src/js/angular/import/services/import-resource-tree.service.js
boyan-tonchev: Renamed to `importResourceElement`

src/js/angular/import/directives/import-resource-tree.directive.js
boyan-tonchev: Because newResource is a parameter, we need to ensure the model is not changed. Actually, after these changes, we do not touch these models I will remove it.

src/js/angular/import/directives/import-resource-tree.directive.js
boyan-tonchev: Nice. I will create an update method in the model and move all the update logic there.

boyan-tonchev:
Jenkins retry build

test-cypress/integration/import/import-server-files.spec.js
```diff
@@ -64,15 +64,43 @@ describe('Import server files', () => {
...
-        ImportServerFilesSteps.getResources().should('have.length', 1);
+        ImportServerFilesSteps.getResources().should('have.length', 2);
         // When I select the files only filter
         ImportServerFilesSteps.selectFilesOnlyFilter();
         // Then I should see only the files
-        ImportServerFilesSteps.getResources().should('have.length', 13);
+        ImportServerFilesSteps.getResources().should('have.length', 15);
+    });
+
+    it('should be able to import hole directory', () => {
```
svilenvelikov, line 79: it's `whole` :)

svilenvelikov (dismissed):

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -190,28 +189,29 @@ function importResourceTreeDirective($timeout, ImportContextService) {
              * and can have their states reset.
              */
             const setCanResetResourcesFlag = () => {
-                $scope.canResetSelectedResources = $scope.selectedResources
-                    .some((treeResource) => treeResource.importResource.status === ImportResourceStatus.DONE);
+                $scope.canResetSelectedResources = $scope.resources.getAllSelected()
+                    .some((treeResource) => {
```
svilenvelikov, line 193: This is better also to be moved in the resources model, e.g. `resources.getSelectedImportedResources()`

boyan-tonchev: Done

src/js/angular/import/directives/import-resource-tree.directive.js
boyan-tonchev: Done

svilenvelikov (approved):

test-cypress/integration/import/import-server-files.spec.js
```diff
@@ -100,7 +100,5 @@ describe('Import server files', () => {
 
         // Then I expect all sub-resource statuses to be reset.
         ImportServerFilesSteps.checkImportedStatusIsEmpty('more-files');
-        ImportServerFilesSteps.checkImportedStatusIsEmpty('jsonld-file.jsonld');
```
svilenvelikov: So. These didn't work as expected? Can you leave them but comment them out and put a TODO to check them later, because these are valid assertions. Right?

boyan-tonchev: Yes, I will do it.

svilenvelikov (approved):

PR #1389, opened 2024-05-09
GDB-10211: Cluster configuration menu irregular behavior on zoom-out

boyan-tonchev:
## What
In Cluster management - the Cluster configuration menu pops-out  unexpectedly, when a zoom-out ( ctrl+mouse scroll ) action is performed in the browser.

## Why
It seems like the issue is related to the way the page slide panel (<pageslide>) is being sized using viewport units (vw). When the window is resizing by pressing control and scrolling, this might cause the page slide panel to momentarily appear or reposition due to the change in its size.

## How
Added a fixed size for the panel width, which is calculated before the panel is opened.


teodossidossev:
lgtm

PR #1393, opened 2024-05-13
GDB-10156 fix buttons height in yasr toolbar

svilenvelikov:
## What
It appears that some of the buttons rendered in the yasr toolbar have greater height than expected.

## Why
For consistency with the other large buttons in the WB.

## How
Applied some explicit height for the buttons in the yasr toolbar.


svilenvelikov:
Jenkins retry build

PR #1400, opened 2024-05-16
GDB-7884 optimize cluster view requests when no cluster is configured

svilenvelikov:
## What
Optimize cluster view requests when no cluster is configured.

## Why
Currently there is a flaw in the request promise chain where when the first request for the cluster status fails with 404 and is caught in the catch clause there is no return from there which makes the next request in the promise to advance and to actually be executed although there is no need from that.

## How
Added a return with rejection from the cluster status promise catch in order to allow the error to be propagated to the parent promise catch and next requests to be skipped.


boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

PR #1405, opened 2024-05-17
GDB-9569 - Show error when ACL role length less than 2 symbols

DesiBorisova:
## What?
When a new ACL rule is created, the CUSTOM ROLE cannot be 1 symbol long. The user will see a notification that the length is not allowed.

## Why?
Previously a CUSTOM ROLE of 1 symbol would be allowed. However, when trying to add it as a tag to a user, one symbol tags are not allowed.

## How?
I added a validation for the length of the ROLE, along with a notification.

## Testing?
I added a new test for this validation.

### DIFF
--- a/src/js/angular/aclmanagement/controllers.js
+++ b/src/js/angular/aclmanagement/controllers.js
@@ -193,6 +193,10 @@ function AclManagementCtrl($scope, $location, toastr, AclManagementRestService, 
      * @param {string} scope
      */
     $scope.saveRule = (scope) => {
+        if (!$scope.rulesModel.isRoleValidLength($scope.editedRuleScope, $scope.editedRuleIndex)) {
+            notifyInvalidRoleLength();
+            return;
+        }
         if ($scope.rulesModel.isRuleDuplicated($scope.editedRuleScope, $scope.editedRuleIndex)) {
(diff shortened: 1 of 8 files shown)


src/js/angular/aclmanagement/controllers.js
```diff
@@ -193,6 +193,10 @@ function AclManagementCtrl($scope, $location, toastr, AclManagementRestService,
      * @param {string} scope
      */
     $scope.saveRule = (scope) => {
+        if (!$scope.rulesModel.isRoleValidLength($scope.editedRuleScope, $scope.editedRuleIndex)) {
+            notifyInvalidRoleLength();
```
svilenvelikov, line 197: I don't like these toastrs but as I can see, the duplicated rule validation is implemented the same way, although this one is different because it validates single field and the other one validates the whole rule. Maybe the validation in this case should be done differently. Probably some error status should be added on the role field and the save button should be disabled until valid role is provided by the user.

DesiBorisova: Alright, I'll do it as suggested

svilenvelikov (approved):

svilenvelikov (approved):

PR #1412, opened 2024-05-21
GDB-10263 Update file status after each file update

svilenvelikov:
## WHAT
Update file status after each file update.

## WHY
There was an issue where when resources were rendered in the resource tree component, there was a `track by resource.name` expression in the template for optimization purposes which used to cause the issue where the template wasn't updated after resource was recreated even though it was passed via two-way binding.

## HOW
Added a new hash property in the resource model which is generated each time based on the server response. Changed the resource-tree directive to track resources by this hash property instead of name.

### DIFF
--- a/src/js/angular/utils/hash-utils.js
+++ b/src/js/angular/utils/hash-utils.js
@@ -1,6 +1,16 @@
 import {KJUR} from "jsrsasign";
 
-export const generateMD5Hash = (str) => {
+/**
+ * Returns a function that generates a MD5 hash for a given string.
+ * @return {function(*): *}
+ */
+export const md5HashGenerator = () => {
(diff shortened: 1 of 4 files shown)


src/js/angular/utils/hash-utils.js
```diff
@@ -0,0 +1,6 @@
+import {KJUR} from "jsrsasign";
+
+export const generateMD5Hash = (str) => {
+    const md = new KJUR.crypto.MessageDigest({alg: 'md5', prov: 'cryptojs'});
```
boyan-tonchev, line 4: Can this be created once, instead of creating it for each resource?

svilenvelikov: yeah, sure. I thought it might not be needed in this case, but I'll change it.

boyan-tonchev (approved):

PR #1414, opened 2024-05-22
GDB-10263: Fixes Issue with Loader

boyan-tonchev:
GDB-10263: Fixes Issue with Loader

What
Sometimes the loaders are not displayed when the import view is opened.

Why
The loader is included in the main import template, and its visibility is controlled by two controllers: one for user data and one for server files. The main controller uses three controllers with a shared variable that determines whether the loader should be displayed. There seems to be an issue with this variable, as both controllers extend the main import controller, leading to inconsistent behavior.

How
Moved the loader into the import-resource-tree directive. The variable that controls the visibility of the loader has been moved to the import context service, allowing it to be changed and used by all components. Additional Work
There is a possible issue when a REST service is called to fetch user data, and before it returns, the tab is changed. In this case, the resource tree might display incorrect results until updated again. To prevent this, a check has been added to ensure the response is processed only if the tab remains the same as it was before the request. If the tab has changed, the response is skipped.

### DIFF
--- a/src/js/angular/import/controllers/import-view.controller.js
+++ b/src/js/angular/import/controllers/import-view.controller.js
@@ -350,8 +350,11 @@ importViewModule.controller('ImportViewCtrl', ['$scope', 'toastr', '$interval', 
          */
         $scope.updateListHttp = (force) => {
             const filesLoader = $scope.activeTabId === TABS.USER ? ImportRestService.getUploadedFiles : ImportRestService.getServerFiles;
+            const executedInTabId = $scope.activeTabId;
             filesLoader($repositories.getActiveRepository()).success(function (data) {
-
+                if (executedInTabId !== ImportContextService.getActiveTabId()) {
+                    return;
+                }
                 if (TABS.SERVER === $scope.activeTabId) {
                     ImportContextService.updateResources(toImportServerResource(toImportResource(data)));
                 } else if (TABS.USER === $scope.activeTabId) {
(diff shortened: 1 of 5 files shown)


boyan-tonchev:
Jenkins retry build

src/js/angular/import/controllers/import-view.controller.js
```diff
@@ -390,7 +393,7 @@ importViewModule.controller('ImportViewCtrl', ['$scope', 'toastr', '$interval',
                 $scope.savedSettings = _.mapKeys(_.filter($scope.files, 'parserSettings'), 'name');
             }).error(function (data) {
                 toastr.warning($translate.instant('import.error.could.not.get.files', {data: getError(data)}));
-            });
+            }).finally(() => ImportContextService.updateShowLoader(false));
```
yordanalexandrov, line 396: This is a bit odd. I don't see you setting the loader to true in this functinon, but you set it to false. I think it is better to have only one method responsible fo the loader states change. You can return the promise and put a finally at the other end.

yordanalexandrov: Or set it true here before the promise. Whichever makes more sense.

boyan-tonchev: This method is called at regular intervals. We want the loader to be displayed when the view is loaded or the tab is changed. When this happens, the showLoader variable is set to true.

yordanalexandrov (approved):

src/js/angular/import/controllers/import-view.controller.js
svilenvelikov: I agree with @yordanalexandrov that this is not the best solution and the first thing he proposed would be a better approach, a.k.a. return the promise form updateListHttp method and hide the loader in the caller

boyan-tonchev: done

boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1415, opened 2024-05-22
GDB-10263 Help box on inport page should be visible on empty state or otherwise when explicitly opened

svilenvelikov:
## WHAT
Help box on inport page should be visible on empty state or otherwise when explicitly opened

## WHY
It was decided that some changes are needed in order to further improve the UX in the import view.


svilenvelikov:
This improvement was implemented with https://github.com/Ontotext-AD/graphdb-workbench/pull/1416

PR #1419, opened 2024-06-03
GDB-10327: Implement error dialog on import failure

boyan-tonchev:
What
When an import fails and the error message is lengthy, the view does not look good.

Why
In the WB, the server message returned for every import is displayed to the user as it is, without any modifications. When the message is too lengthy, it is displayed as simple text without formatting. This makes the error report unusable.

How
When the error message is too lengthy, we will show a small piece to the user as a button. When the user clicks it, a dialog with the full error message will open.

### DIFF
--- a/src/css/import-resource-tree.css
+++ b/src/css/import-resource-tree.css
@@ -42,6 +42,17 @@
     padding-top: 8px;
 }
 
+.import-resource-table .even-row-start .cell,
+.import-resource-table .odd-row-start .cell {
+    padding-top: 12px;
+}
+
+.import-resource-table .even-row-end .cell,
+.import-resource-table .odd-row-end .cell {
+    padding-top: 4px;
+    padding-bottom: 12px;
+}
+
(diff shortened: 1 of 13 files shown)


src/css/import-resource-tree.css
```diff
@@ -42,6 +42,17 @@
...
 
+.import-resource-table .even-row-start .cell,
+.import-resource-table .odd-row-start .cell {
+    padding-top: 12px;
+}
+
+.import-resource-table .even-row-end .cell,
+.import-resource-table .odd-row-end .cell {
+    padding-top: 4px;
+    padding-bottom: 12px;
```
svilenvelikov, line 53: Why these additional paddings were needed? Is by request by someone?

svilenvelikov (approved):

src/css/import-resource-tree.css
boyan-tonchev: Yes. After conversation with Donka.

PR #1421, opened 2024-06-04
Fixed three time calling init function in import view

boyan-tonchev:
# Fixes three-time initialization of main import service

## What
- When the import view is opened, the init method of the main controller is called three times;
- Both the import and server data controllers are initialized regardless of the selected tab

## Why
On Bootstrap, all import controllers are initialized because they are defined in one module.
- When the main controller initializes all fields and functions, the init method is called. Inside the controller declaration, it calls its own init() function.
- When the server import controller is initialized, it extends the main controller. After initializing the fields and functions:
  - The main controller's init() function is called for the second time (triggered because the server import controller extends the main controller).
  - When the server controller is initialized, it calls its init method regardless of which tab is chosen.
- When the user import controller is initialized, it extends the main controller. After initializing the fields and functions:
  - The main controller's init() function is called for the third time (triggered because the user import controller extends the main controller);
  - When the user controller is initialized, it calls its init method regardless of which tab is chosen.
### Summary
When the page is loaded:
- The init method of the main controller is called due to self-initialization (This is not OK because the main controller is an 'abstract' class and should not be self-initialized);
-The server import controller is called regardless of whether the user import tab is selected:
   - The main class init method is called (at this point, it is OK for the superclass logic to be initialized);
   - The init method of the server import controller is called (after the superclass is ready, it is OK to initialize the server import controller).
- The user import controller is called regardless of whether the server import tab is selected:
   - The main class init method is called (at this point, it is OK for the superclass logic to be initialized);
   - The init method of the user import controller is called (after the superclass is ready, it is OK to initialize the user import controller).

## How
- Remove the self-initialization of the main controller;
- Add a condition to prevent the server import controller from self-initializing if the user import tab is selected;
- Add a condition to prevent the user import controller from self-initializing if the server import tab is selected.

There is a random NPE in Cypress tests

What
Sometimes Cypress fails when trying to import RDF data from a URL or text snippet.

Why
This can happen when the setSettingsFor function is called before the default settings are loaded.

How
A getDefaultSettings function is implemented that returns a Promise with the default settings of the chosen repository. The function returns immediately if the settings are already loaded; otherwise, it loads them from the backend.

### DIFF
--- a/src/js/angular/import/controllers/import-view.controller.js
+++ b/src/js/angular/import/controllers/import-view.controller.js
@@ -83,15 +83,6 @@ importViewModule.controller('ImportViewCtrl', ['$scope', 'toastr', '$interval', 
 
         $scope.toTitleCase = (str) => stringUtils.toTitleCase(str);
 
-        $scope.pollList = () => {
-            listPollingHandler = $interval(() => {
-                // Skip iteration if we are updating something
-                if (!$scope.updating) {
-                    $scope.updateList(false);
-                }
-            }, LIST_POLLING_INTERVAL);
-        };
-
         /**
          * Sets the settings for the given file and triggers import process.
          * TODO: this function should be refactored because it does too much.
@@ -302,6 +293,15 @@ importViewModule.controller('ImportViewCtrl', ['$scope', 'toastr', '$interval', 
         // Private functions
         // =========================
 
+        const pollList = () => {
+            listPollingHandler = $interval(() => {
+                // Skip iteration if we are updating something
+                if (!$scope.updating) {
+                    $scope.updateList(false);
+                }
+            }, LIST_POLLING_INTERVAL);
+        };
+
         const getSettings = () => {
             if (!$scope.canWriteActiveRepo()) {
                 return;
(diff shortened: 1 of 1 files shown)


src/js/angular/import/controllers/import-view.controller.js
```diff
@@ -425,28 +453,30 @@ importViewModule.controller('ImportViewCtrl', ['$scope', 'toastr', '$interval',
...
 
         // =========================
         // Initialization
         // =========================
-
-        // TODO: Beware that this init is called tree times due to the child controllers which extends this one. We should refactor this.
-        const init = () => {
+        $scope.importViewControllerInit = () => {
+            loadDefaultSettings();
+            pollList();
```
svilenvelikov, line 475: I wonder if these two calls above: the `pollList()` and the `loadDefaultSettings()` shouldn't be moved below subscriptions and loading the app data, or even chanined to getAppData maybe?

boyan-tonchev: I will try

svilenvelikov (approved):

src/js/angular/import/controllers/import-view.controller.js
boyan-tonchev: Done

PR #1429, opened 2024-06-17
GDB-10354 - Move tabs in Resource view together when window resized

DesiBorisova:
## What?
When resizing the window, the tabs in the Resource view will move down together, to make space for the button group on the right.

## Why?
The previous behavior was odd, because each tab would move down separately.

## How?
I changed the styling of the tabs.

## Screenshots?
On resize after the suggested improvement to the styling:
Screenshot from 2024-06-17 10-13-02

### DIFF
--- a/src/css/bootstrap-graphdb-theme.css
+++ b/src/css/bootstrap-graphdb-theme.css
@@ -1412,6 +1412,7 @@ footer.footer .container-fluid.main-container p {
 -------------------------------------------------------------*/
 .nav-tabs {
     font-weight: 400;
+    display: inline-block;
 }
 
 .nav-tabs .nav-item {


src/css/bootstrap-graphdb-theme.css
```diff
@@ -1412,6 +1412,7 @@ footer.footer .container-fluid.main-container p {
 -------------------------------------------------------------*/
 .nav-tabs {
     font-weight: 400;
+    display: inline-block;
```
svilenvelikov, line 1415: I don't quite like setting this here in the theme because it would affect all nav-tabs, but I think it's OK for now

svilenvelikov (approved):

DesiBorisova:
Jenkins retry build

PR #1432, opened 2024-06-18
GDB-10366 import view batch operations improvements

svilenvelikov:
## What
- Don't filter out selected folders when doing batch reset;
- The messages describing the added/removed statements are not displayed after the import finishes. They are only displayed after a page refresh.

## Why
- As we recently started sending the folders when doing batch import as well and they can now have an import status. This is in contrast with the old behavior where folders didn't have an import status;
- There are flags (hasReplacedGraphs, hasAddedStatements and hasRemovedStatements) that are calculated when directives are created. These flags remain unchanged even if the import resource is modified.

## How
- Removed the folders filtering when collecting resources for batch reset;
- Remove the flags from the import-resource-message and import-resource-status-info directives. The flags in import-resource-status-info were unused after the decision that the message should always be displayed.


boyan-tonchev (dismissed):

svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

PR #1435, opened 2024-06-18
GDB-10369 JSONLD and NDJSONLD export improvements

sava-savov-ontotext:
What
Current WB implementation of the Link header value in JSON-LD requests sent to GDB is missing the ; rel=http://www.w3.org/ns/json-ld#frame and ; rel=http://www.w3.org/ns/json-ld#context attributes. 
Why
These rel attributes are essential for indicating the relationship type of the linked resources as per the JSON-LD specifications.
How
Added rel attribute in Link header based on selected mode.


svilenvelikov (approved):

sava-savov-ontotext:
Jenkins retry build

PR #1452, opened 2024-06-28
GDB-10391 - Render menu labels faster

DesiBorisova:
## What?
The "Import" and "SPARQL" main menu labels will render faster on window resize.

## Why?
The two labels would load slower than the rest during resize.

## How?
I changed the ng-if to ng-show and added a debounce function to the function called.


src/js/angular/controllers.js
```diff
@@ -716,6 +719,40 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repos
...
+     * Creates a debounced function that delays invoking the provided function until after a specified wait time has
+     * elapsed since the last time the debounced function was invoked. Optionally, the function can be invoked immediately
+     * and then wait for the specified time before allowing the next invocation.
+     * @param {Function} func - The function to debounce.
+     * @param {number} wait - The number of milliseconds to delay.
+     * @param {boolean} [callImmediately=false] - If true, the function will be invoked immediately, then wait for the specified time
+     *                                           before allowing the next invocation.
+     * @return {Function} - Returns the new debounced function.
+     */
+    function debounce(func, wait, callImmediately) {
```
svilenvelikov, line 732: Two notes here. First, why this is here and not in some utility and second, why not just using the lodash's debounce function?

DesiBorisova: The lodash debounce doesn't always work so well in this case. The labels still sometimes load slower. I will move this to a utility class instead.

svilenvelikov: Well, how that can be?! :) Did you try tweaking the debounce parameters? Try passing `{trailing: false}`.

src/js/angular/controllers.js
```diff
@@ -23,6 +23,7 @@ import './guides/directives';
 import {GUIDE_PAUSE} from './guides/tour-lib-services/shepherd.service';
 import 'angular-pageslide-directive/dist/angular-pageslide-directive';
 import 'angularjs-slider/dist/rzslider.min';
+import _ from "lodash";
```
svilenvelikov, line 26: import only the debounce function and not the entire thing please

DesiBorisova: Done :smile:

svilenvelikov (approved):

PR #1455, opened 2024-07-02
GDB-10382: Default message info not displayed when the reinitialized function is called

boyan-tonchev:
What
The YASR results and YASQE buttons are not cleared when the repository is changed from the repository view.

Why
The YASGUI is a third-party library and does not know about repository changes. Managing repository changes is a function of the GraphDB Workbench.

How
Added functionality to persist the repository used by the YASGUI instance. When the YASGUI is initialized, a clearState flag is passed to the configuration. The value of the flag is based on whether the current repository is different from the last one used by YASGUI. If the repositories differ, the flag is set to true, which triggers the clearing of the YASR results and the YASQE buttons.

### DIFF
--- a/src/js/angular/sparql-editor/controllers.js
+++ b/src/js/angular/sparql-editor/controllers.js
@@ -15,13 +15,13 @@ import {VIEW_SPARQL_EDITOR} from "../models/sparql/constants";
 import {CancelAbortingQuery} from "../models/sparql/cancel-aborting-query";
 import {QueryMode} from "../models/ontotext-yasgui/query-mode";
 import 'angular/core/services/event-emitter-service';
-import {YasguiResetFlags} from "../models/ontotext-yasgui/yasgui-reset-flags";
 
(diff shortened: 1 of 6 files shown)


boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

src/js/angular/sparql-editor/controllers.js
```diff
@@ -70,6 +70,7 @@ function SparqlEditorCtrl($rootScope,
 
     const QUERY_EDITOR_ID = '#query-editor';
     let activeRepository = $repositories.getActiveRepository();
+    let isOntopRepo;
```
svilenvelikov, line 73: Should we have a default value here?

boyan-tonchev: Done

svilenvelikov (approved):

PR #1458, opened 2024-07-05
GDB-10463 workbench lyaout and navigation

svilenvelikov:
## What
Implemented navbar application and registered in the single-spa layout.

## Why
Our old workbench and new workbench applications expose only views and we need a new custom layout and navigation.

## How
Created a new navbar application that is integrated in the single-spa layout.


boyan-tonchev:
The install, start, and build scripts need to be updated to install, start, and build the new app navbar.

boyan-tonchev (approved):

PR #1476, opened 2024-07-18
GDB-10512: run the host and legacy workbench on a single server

boyan-tonchev:
## What
Updates projects to build/start the root-config and legacy-workbench projects on one destination/server.

## Why
The root-config is the main project that orchestrates all other ones. The legacy-workbench uses relative paths for endpoints. If both projects run on one server, no changes will be needed for the legacy-workbench to work.

## How
Added webpack to the root container of all projects. Added configurations for development and production builds.

# Known issues that will be fixed in next requests
- Version replacements for js files
- License report
- When a resource is changed, the rebuild is triggered but the browser page is not refreshed.


packages/legacy-workbench/package.json
```diff
@@ -38,37 +35,17 @@
...
-        "extract-loader": "^5.1.0",
-        "file-loader": "^4.3.0",
-        "html-loader": "^0.5.5",
-        "html-webpack-plugin": "^5.6.0",
-        "ignore-loader": "^0.1.2",
         "is-obj": "^2.0.0",
         "less": "^3.10.3",
         "less-loader": "^5.0.0",
         "license-checker": "^25.0.1",
         "mini-css-extract-plugin": "^0.8.0",
```
svilenvelikov, line 45: Check if this webpack plugin is still needed here

packages/legacy-workbench/package.json
```diff
@@ -38,37 +35,17 @@
...
         "eslint-config-google": "^0.13.0",
-        "expose-loader": "^0.7.5",
-        "extract-loader": "^5.1.0",
-        "file-loader": "^4.3.0",
-        "html-loader": "^0.5.5",
-        "html-webpack-plugin": "^5.6.0",
-        "ignore-loader": "^0.1.2",
         "is-obj": "^2.0.0",
         "less": "^3.10.3",
         "less-loader": "^5.0.0",
```
svilenvelikov, line 43: This is also a webpack plugin.

webpack.config.common.js
```diff
@@ -0,0 +1,308 @@
...
+        resolve: {
+            modules: [
+                './packages/legacy-workbench/src/js/',
+                './packages/legacy-workbench/node_modules',
+                './packages/root-config/src/js/',
+                './packages/root-config/node_modules'
+            ],
+            extensions: ['.js']
+        },
+        externals: ["@ontotext/workbench-api", "@ontotext/components"],
```
svilenvelikov, line 56: `@ontotext/components` check this out as well! We don't have such module

packages/legacy-workbench/package.json
boyan-tonchev: Done

packages/workbench/extra-webpack.config.js
```diff
@@ -8,7 +8,7 @@ module.exports = (config, options) => {
   // delete singleSpaWebpackConfig.externals['zone.js'];
   // Feel free to modify this webpack config however you'd like to
 
-  singleSpaWebpackConfig.externals = ["@ontotext/workbench-api"];
+  singleSpaWebpackConfig.externals = ["@ontotext/workbench-api", "@ontotext/components"];
```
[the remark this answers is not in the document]
boyan-tonchev: Done

webpack.config.common.js
boyan-tonchev: Done

packages/legacy-workbench/package.json
boyan-tonchev: Done

packages/root-config/src/index.ejs
```diff
@@ -37,29 +37,20 @@
...
     <link rel="preload" href="https://cdn.jsdelivr.net/npm/single-spa@5.9.0/lib/system/single-spa.min.js" as="script">
-    <script src="http://localhost:9001/plugin-registry.js?v=2.7.0-RC3"></script>
-    <script src="http://localhost:9001/plugins.js?v=2.7.0-RC3"></script>
+    <script src="plugin-registry.js?v=<%= buildVersion %>"></script>
+    <script src="plugins.js?v=<%= buildVersion %>"></script>
 
     <!-- Add your organization's prod import map URL to this script's src  -->
     <!-- <script type="systemjs-importmap" src="/importmap.json"></script> -->
 
-    <% if (isLocal) { %>
```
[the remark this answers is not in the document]
boyan-tonchev: We don't need it at the moment. It is used when we want to load different files depending on the mode ("development" or "production"). In our case, we have different webpack configurations and we pass the names of bundles during the build process, so we always have to load these files.

svilenvelikov (approved):

PR #1485, opened 2024-07-25
GDB-10443 Implement i18n in the new workbench

boyan-tonchev:
Implement i18n in the new workbench

## What
 - implemented translation for legacy-workbench and shared-component when click on language label;
 - Implemented core functionality for emitting events across all microfrontends;
 - Implemented a service provider.
 - Implemented language.service.ts.

## Why
 -The language selector is moved in shared-component module and when the language is changed all microfrontends have to be notified;
 - Microfrontends are separate modules and should not be aware of each other. The API is a shared module across all of them. The event emitting functionality enables communication between these modules
 - The new service provider is responsible for managing shared service instances across microfrontends. Frameworks like Angular and React provide dependency injection, allowing a single instance to be injected into all components. In a single-spa application, such functionality does not exist by default. The service provider addresses this need.
 - The service will manage language functionality.
 

## How
 - Added a listeners for the language change event. When the event is triggered, the new language value is set in the translation service, which triggers the re-translating the entire workbench;
 - Utilized the browser's built-in custom event propagation. Events are emitted from the document body, ensuring that all microfrontends can listen to them.
 - The service provider includes lazy loading functionality that returns an instance of an already instantiated service. If the service is not present, a new one is created and stored in a static map. The static modifier ensures that only one instance of each service class is used across the microfrontends.
 - The language service manages language-related functionality. All microfrontends can use it to fetch and update the language.


packages/api/src/services/workbench-language.service.ts
```diff
@@ -0,0 +1,26 @@
...
+import {LanguageChangeEvent} from '../models/events/language/language-chage.event';
+import {WorkbenchService} from './workbenchService';
+import {WorkbenchGlobalContextService} from './workbench-global-context.service';
+import {Subject} from '@reactivex/rxjs/dist/package';
+
+export class WorkbenchLanguageService implements WorkbenchService {
+    private readonly eventService: WorkbenchEventService;
+    private readonly globalContextService: WorkbenchGlobalContextService;
+
+    static readonly DEFAULT_LANGUAGE = 'en';
```
svilenvelikov, line 12: There is something odd in these services the global context one and this. At first they seem a bit overcomplicated and second, it seems that there is some mashup in their purposes.

packages/api/src/services/workbenchService.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Interface for marking a service object as a workbench service.
+ * This interface is used to designate services that are specific
+ * to the workbench environment.
```
svilenvelikov, line 4: Will there be any other environments?

packages/legacy-workbench/src/app.js
```diff
@@ -218,6 +220,21 @@ const moduleDefinition = function (productInfo) {
         ThemeService.applyDarkThemeMode();
 
         GuidesService.init();
+
+        // =========================
+        // Functions and configurations for integration with the shared-components module.
+        // =========================
+        let globalContextService = WorkbenchServiceProvider.get(WorkbenchGlobalContextService);
+
+        let languageChangeSubscriptions = globalContextService.onLanguageChanged()
```
svilenvelikov, line 229: Again. Why we have a language service if we don't even use it?

packages/legacy-workbench/src/app.js
```diff
@@ -218,6 +220,21 @@ const moduleDefinition = function (productInfo) {
...
+            .subscribe((language) => {
+            $translate.use(language);
+        });
+
+        $rootScope.$on('destroy', () => {
+            console.log("Destroy")
```
svilenvelikov, line 235: logger

packages/shared-components/src/components/onto-header/onto-header.tsx
```diff
@@ -1,18 +1,39 @@
...
 @Component({
   tag: 'onto-header',
   styleUrl: 'onto-header.scss',
   shadow: false,
 })
 export class OntoHeader {
+
+  private eventService: WorkbenchEventService;
+
+  private locale = 'en'
```
svilenvelikov, line 13: So, in how many places we need this default to be declared?!

packages/shared-components/src/components/onto-header/onto-header.tsx
```diff
@@ -1,18 +1,39 @@
...
+  }
+
+  private onClick(): void {
+    // TODO remove this when implement language selector
+    if (this.locale === 'en') {
+      this.locale = 'fr';
+    } else {
+      this.locale = 'en';
+    }
+    this.eventService.emit({NAME: WorkbenchEventType.LANGUAGE_CHANGED, payload: {locale: this.locale}})
```
svilenvelikov, line 26: Well. This is another way to emit language chage event. The logic clearly needs to be reconsidered and cleaned up. Right now you have a glovbal context service, a language service and yet here the event is emitted directly via the event service. Why?

packages/shared-components/src/components/onto-header/onto-header.tsx
```diff
@@ -1,18 +1,39 @@
...
+
+  private eventService: WorkbenchEventService;
+
+  private locale = 'en'
+
+  constructor() {
+    this.eventService = WorkbenchServiceProvider.get(WorkbenchEventService);
+  }
+
+  private onClick(): void {
```
svilenvelikov, line 19: There is better way to name these handlers. The event which triggers the handler is `click` but the method does the language change logic. So it should be named changeLanguage or something.

svilenvelikov (changes requested):

packages/api/src/models/events/workbench.event.ts
```diff
@@ -0,0 +1,12 @@
+export const WorkbenchEventType = {
+    'LANGUAGE_CHANGED': 'languageChangeEvent'
```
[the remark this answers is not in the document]
boyan-tonchev: Here will be defined the types of all global events.

packages/api/src/services/workbench-event.service.ts
```diff
@@ -0,0 +1,32 @@
+import {WorkbenchEvent} from '../models/events/workbench.event';
+import {WorkbenchService} from './workbenchService';
+
+/**
+ * Service used for global communication within the workbench ecosystem.
+ */
+export class WorkbenchEventService implements WorkbenchService {
+    private hostElement;
```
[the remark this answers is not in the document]
boyan-tonchev: This is just a reference to the document body. It doesn't change once the page is loaded, which is why I hold the reference here. The alternative would be to search for the body element every time an event is fired.

packages/shared-components/src/components/onto-header/onto-header.tsx
boyan-tonchev: This is code just for testing. It will be removed in next request when language selector is implemented.

packages/shared-components/src/components/onto-header/onto-header.tsx
boyan-tonchev: This is code just for testing. It will be removed in next request when language selector is implemented.

packages/shared-components/src/components/onto-header/onto-header.tsx
boyan-tonchev: This is what I wanted to discuss with you and the team. There are two ways for other modules to interact with the global store:

The first is through the global context store using events and subscriptions.
The second is via services. For example, if we need to read or write the selected language, we can interact with the LanguageService.
Depending on the decision, I will rewrite the global and language services.

packages/api/src/services/workbench-event.service.ts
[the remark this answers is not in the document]
svilenvelikov: `The alternative would be to search for the body element every time an event is fired.` this is the way it should be. It's not just the body, but almost the entire page. Caching it in a service is not a good idea as I already said.

packages/api/src/models/events/workbench.event.ts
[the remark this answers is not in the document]
svilenvelikov: I know that we sometimes used to define them in a single place, but I think that its probably better to keep related stuff closer.

packages/shared-components/src/components/onto-header/onto-header.tsx
svilenvelikov: So, put a TODO maybe, because I didn't think of this :)

packages/shared-components/src/components/onto-header/onto-header.tsx
svilenvelikov: I prefer the service approach. I dont want to drag the entire global store to places where it is not needed, not to mention that it would have a more generic API which might lead to confusion and errors. If I want to subscribe to language change event or emit the same, I'd use the language service and not the generic one. So, in short, stick with the service approach.

svilenvelikov:
Did you plan to write some tests?

boyan-tonchev:
> Did you plan to write some tests?

Yes, but with other task GDB-10570

packages/api/src/models/events/workbench.event.ts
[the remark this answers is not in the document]
boyan-tonchev: Removed

packages/api/src/services/workbench-event.service.ts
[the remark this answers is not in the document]
boyan-tonchev: Done

packages/api/src/services/workbench-event.service.ts
```diff
@@ -0,0 +1,32 @@
...
+    constructor() {
+        this.hostElement = document.querySelector('body');
+    }
+
+    emit(workbenchEvent: WorkbenchEvent): CustomEvent {
```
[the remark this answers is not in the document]
boyan-tonchev: documented

packages/api/src/services/workbench-global-context.service.ts
```diff
@@ -0,0 +1,29 @@
+import {ReplaySubject, Subject} from '@reactivex/rxjs/dist/package';
+import {WorkbenchServiceProvider} from '../workbench-service.provider';
+import {WorkbenchEventService} from './workbench-event.service';
+import {WorkbenchEventType} from '../models/events/workbench.event';
+
+/**
+ * The purpose of this service is to hold the application's global state.
```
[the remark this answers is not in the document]
boyan-tonchev: The class is removed

packages/api/src/services/workbench-global-context.service.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export class WorkbenchGlobalContextService {
```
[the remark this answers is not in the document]
boyan-tonchev: Removed

packages/api/src/services/workbench-language.service.ts
boyan-tonchev: The global one is removed.

packages/api/src/services/workbenchService.ts
boyan-tonchev: Updated

packages/legacy-workbench/src/app.js
boyan-tonchev: Refactored

packages/shared-components/src/components/onto-header/onto-header.tsx
boyan-tonchev: Added comment

packages/shared-components/src/components/onto-header/onto-header.tsx
boyan-tonchev: Refactored.

packages/shared-components/src/components/onto-header/onto-header.tsx
boyan-tonchev: Commented

svilenvelikov (approved):

PR #1489, opened 2024-07-29
GDB-10573 - Add hash to repositories list for "track by" in view

DesiBorisova:
## What?

- When hovering on a button in the repositories table, the popovers will remain visible until the mouse is moved away from the button or some of the data in the repository Map is changed.

- Added a missing French label in the Guides view.

## Why?
The popovers in the repositories table would hide each time the list was refreshed from a call to the API (every 5 seconds).

## How?
I add hashes to the repositories, before they are stored in the Locations Map. The template now will track the changes by each hash.

### DIFF
--- a/test-cypress/integration/import/import-view.spec.js
+++ b/test-cypress/integration/import/import-view.spec.js
@@ -98,6 +98,8 @@ describe('Import view', () => {
 
         // When I go to user tab and upload a file
         ImportSteps.openUserDataTab();
+        // And I wait for tab data to load
+        cy.wait(100);
         ImportUserDataSteps.selectFile(ImportUserDataSteps.createFile(testFiles[0], bnodes));
         ImportSettingsDialogSteps.import();
         ImportUserDataSteps.getResources().should('have.length', 1);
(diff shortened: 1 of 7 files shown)


DesiBorisova:
Jenkins retry build

DesiBorisova:
Jenkins retry build

test-cypress/integration/import/import-view.spec.js
```diff
@@ -98,6 +98,8 @@ describe('Import view', () => {
 
         // When I go to user tab and upload a file
         ImportSteps.openUserDataTab();
+        // And I wait for tab data to load
+        cy.wait(100);
```
svilenvelikov, line 102: Does this improve the things?

svilenvelikov (approved):

test-cypress/integration/import/import-view.spec.js
DesiBorisova: Locally, the test keeps passing. Seems like it gives Cypress just enough time to load what it needs on the page and find the modal on file upload. I tried checking for file creation and size and contents instead of waiting, but it was ineffective (the file was created, but the test still failed).

PR #1491, opened 2024-07-30
feat(GDB-10629) Add GA4 tracking code in GDB Workbench

teodossidossev:
## WHAT:
- Introduced tracking logic based on the type of license in use (Free, Evaluation, or Enterprise).
- Refactored the $licenseService to manage license statuses, types, and tracking permissions more effectively.
- Updated Cypress tests to cover different license types and ensure correct behavior of Google Tag Manager (GTM) injection.
- Modified the app’s initialization to check and update the license status on load.
- Added stubs for different license types in tests.

## WHY:
To enable Google Tag Manager (GTM) on the application for enhanced tracking and analytics capabilities.

## HOW:
- Refactored the $licenseService to handle the license information, product type, and tracking permissions.
- Added a method to dynamically inject or remove GTM scripts based on the license type and tracking allowance.
- Implemented test cases in Cypress to verify GTM behavior for different license scenarios.
- Updated the app's run block to automatically check the license status on app initialization and update tracking accordingly.

### DIFF
--- a/src/js/angular/controllers.js
+++ b/src/js/angular/controllers.js
@@ -125,11 +125,11 @@ function homeCtrl($scope, $rootScope, $http, $repositories, $jwtAuth, $licenseSe
 
 mainCtrl.$inject = ['$scope', '$menuItems', '$jwtAuth', '$http', 'toastr', '$location', '$repositories', '$licenseService', '$rootScope',
     'productInfo', '$timeout', 'ModalService', '$interval', '$filter', 'LicenseRestService', 'RepositoriesRestService',
-    'MonitoringRestService', 'SparqlRestService', '$sce', 'LocalStorageAdapter', 'LSKeys', '$translate', 'UriUtils', '$q', 'GuidesService', '$route', '$window', 'AuthTokenService'];
+    'MonitoringRestService', 'SparqlRestService', '$sce', 'LocalStorageAdapter', 'LSKeys', '$translate', 'UriUtils', '$q', 'GuidesService', '$route', '$window', 'AuthTokenService', '$document'];
 
 function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repositories, $licenseService, $rootScope,
                   productInfo, $timeout, ModalService, $interval, $filter, LicenseRestService, RepositoriesRestService,
-                  MonitoringRestService, SparqlRestService, $sce, LocalStorageAdapter, LSKeys, $translate, UriUtils, $q, GuidesService, $route, $window, AuthTokenService) {
+                  MonitoringRestService, SparqlRestService, $sce, LocalStorageAdapter, LSKeys, $translate, UriUtils, $q, GuidesService, $route, $window, AuthTokenService, $document) {
     $scope.descr = $translate.instant('main.gdb.description');
     $scope.documentation = '';
     $scope.menu = $menuItems;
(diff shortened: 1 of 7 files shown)


src/js/angular/controllers.js
```diff
@@ -125,11 +125,11 @@ function homeCtrl($scope, $rootScope, $http, $repositories, $jwtAuth, $licenseSe
...
     'productInfo', '$timeout', 'ModalService', '$interval', '$filter', 'LicenseRestService', 'RepositoriesRestService',
-    'MonitoringRestService', 'SparqlRestService', '$sce', 'LocalStorageAdapter', 'LSKeys', '$translate', 'UriUtils', '$q', 'GuidesService', '$route', '$window', 'AuthTokenService'];
+    'MonitoringRestService', 'SparqlRestService', '$sce', 'LocalStorageAdapter', 'LSKeys', '$translate', 'UriUtils', '$q', 'GuidesService', '$route', '$window', 'AuthTokenService', '$document'];
 
 function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repositories, $licenseService, $rootScope,
                   productInfo, $timeout, ModalService, $interval, $filter, LicenseRestService, RepositoriesRestService,
-                  MonitoringRestService, SparqlRestService, $sce, LocalStorageAdapter, LSKeys, $translate, UriUtils, $q, GuidesService, $route, $window, AuthTokenService) {
+                  MonitoringRestService, SparqlRestService, $sce, LocalStorageAdapter, LSKeys, $translate, UriUtils, $q, GuidesService, $route, $window, AuthTokenService, $document) {
```
yordanalexandrov, line 132: Why is `$document` added? Am I missing something

teodossidossev: no, my bad

yordanalexandrov (dismissed):

src/js/angular/core/services/license.service.js
```diff
@@ -1,47 +1,198 @@
...
+     *
+     * This function searches through all `<script>` elements in the document and returns `true` if any script
+     * has a `src` attribute that contains 'googletagmanager'. It returns `false` otherwise.
+     *
+     * @return {boolean} - Returns `true` if a GTM script is found, otherwise `false`.
+     * @private
+     */
+    const gtmScriptExists = () => {
+        const scripts = $document[0].getElementsByTagName('script');
+        return Array.from(scripts).some((script) => script.src && script.src.includes('googletagmanager'));
```
svilenvelikov, line 121: Two things here:
* You add the gtm script in the head, but here you are searching for it in the entire document which a bit resource waisting because all app scripts are appended in the body and they are huge. So, you can get the scripts only from the head tag where there just a few of them and they are small relative to bundle scripts.
* Why don't you add an attribute (id for example) to the gtm script tag and just select it instead of searching in scripts content as you do currently.

test-cypress/integration/home/workbench.home.spec.js
```diff
@@ -191,6 +192,37 @@ describe('Home screen validation', () => {
         });
     });
 
+    context.only('GA', () => {
```
svilenvelikov, line 195: only

svilenvelikov (dismissed):

test-cypress/integration/home/workbench.home.spec.js
teodossidossev: fixed

src/js/angular/core/services/license.service.js
teodossidossev: When the GTM script is injected, it dynamically adds additional scripts depending on how many tags
are configured within your GTM container. These scripts are usually inserted into the `<head>` section,
but it’s better to search through the entire document to ensure all related GTM scripts are found.
Added to js docs also

svilenvelikov (approved):

PR #1501, opened 2024-08-08
GDB-10621 Remove the Authentication options in add GDB modal in cluster

svilenvelikov:
What
When adding cluster nodes we have three options for node-to-node authentication - None, Basic auth, and Signature.

Why
In context of the cluster only the last Signature option is selected and valid and it is applied implicit.

How
Added new boolean field 'clusterMode' in newLocation object, which is set only in the cluster controller and based on it's value the radio buttons are hidden and the message modified
Implemented additional tests.

(cherry picked from commit cc0ac273c94b7917602c453eb1d73d648efad0c6)


svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

PR #1511, opened 2024-08-19
GDB-10450 - Add warnings when user uses CUSTOM_ prefix

DesiBorisova:
## What?
The user will be warned when using "CUSTOM_" as a prefix for `ACL` rules and when adding roles to a `User` in `Setup`. 

## Why?
The back-end adds the prefix and the user is not notified, which can lead to errors when adding roles.

## How?
I created new templates for the error messages and added a check during the rest of the rule validation steps.

## Testing?
Tests added.

## Screenshots?
The message in ACL view:
Screenshot from 2024-08-21 17-48-38
Screenshot from 2024-08-21 17-48-51

The message in User setup:
image


DesiBorisova:
Jenkins retry build

boyan-tonchev (approved):

src/js/angular/security/controllers.js
```diff
@@ -510,6 +511,7 @@ securityCtrl.controller('CommonUserCtrl', ['$rootScope', '$scope', '$http', 'toa
          * @return {boolean} true if valid, otherwise false
          */
         $scope.isCustomRoleValid = function (fieldValue) {
+            $scope.showCustomPrefixMessage = RoleNamePrefixUtils.isCustomPrefixUsed(fieldValue.text);
```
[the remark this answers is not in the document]
svilenvelikov: Also, how boolean type variables are usually called? Better name would be `hasCustomPrefix` or something.

src/pages/aclmanagement.html
```diff
@@ -164,7 +166,11 @@ <h1 id="acl-management-view-title">
                                         <div class="small" ng-if="ruleData.role.$touched && !ruleData.role.$valid">
                                             <small>{{'acl_management.errors.role_length_too_short' | translate}}</small>
                                         </div>
+                                        <div class="small float-lg-left prefix-warning-text" ng-if="rule.role.startsWith(doublePrefix) && !showPrefixWarningIcon">
+                                            <small>{{'acl_management.rulestable.prefix_warning.text' | translate}}</small>
+                                        </div>
                                         <em></em>
+                                        <em ng-if="rule.role.startsWith(doublePrefix) && showPrefixWarningIcon" class="icon-warning" gdb-tooltip="{{'acl_management.rulestable.prefix_warning.text' | translate}}" tooltip-placement="right"></em>
```
svilenvelikov, line 173: `ng-if="rule.role.startsWith(doublePrefix) && showPrefixWarningIcon"` I don't get this. Can the role name have a double `custom` prefix without having a single one before that?

src/pages/aclmanagement.html
```diff
@@ -654,7 +678,11 @@ <h1 id="acl-management-view-title">
                                         <div class="small" ng-if="ruleData.role.$touched && !ruleData.role.$valid">
                                             <small>{{'acl_management.errors.role_length_too_short' | translate}}</small>
                                         </div>
+                                        <div class="small float-lg-left prefix-warning-text" ng-if="rule.role.startsWith(doublePrefix) && !showPrefixWarningIcon">
```
svilenvelikov, line 681: I think these checks needs to be reconsidered, because they look quite strange at the moment. How can a role starts with a CUSTOM_CUSTOM_ and not with a CUSTOM_ before that which is the meaning of the showPrefixWarningIcon if I read it properly?

test-cypress/steps/setup/acl-management-steps.js
```diff
@@ -165,6 +169,22 @@ export class AclManagementSteps {
         return cy.get('div.small');
     }
 
+    static getPrefixWarning() {
+        return cy.get('.prefix-warning-text');
```
svilenvelikov, line 173: Why you get stuff without context but directly from the page?! Don't do that! Try to narrow down your selectors as much as possible. Look at how the other methods in the steps are structured!

test-cypress/steps/setup/acl-management-steps.js
```diff
@@ -165,6 +169,22 @@ export class AclManagementSteps {
...
+    static getWarningIcon() {
+        return cy.get('.icon-warning');
+    }
+
+    static getWarningIconTooltipText() {
+        return cy.get('.angular-tooltip');
+    }
+
+    static mouseoverWarningIcon() {
+        cy.get('.icon-warning').first().trigger('mouseover');
```
svilenvelikov, line 185: This is what I meant with the above comment. You wouldn't needed the `first()` call if you scoped the selector in the first place. Not to mention that you duplicated the `.icon-warning` selector violating the purpose of the step files...

PR #1518, opened 2024-08-23
(cherry pick) GDB-10450 - Add warnings when user uses CUSTOM_ prefix (#1511)

DesiBorisova:
## What?
The user will be warned when using "CUSTOM_" as a prefix for ACL rules and when adding roles to a User in Setup.

## Why?
The back-end adds the prefix and the user is not notified, which can lead to errors when adding roles.

## How?
I created new templates for the error messages and added a check during the rest of the rule validation steps.

## Testing?
Tests added.

## Screenshots?
The message in ACL view:
image
image

The message in User setup:
image

* GDB-10450 - add warnings when user uses CUSTOM_ prefix on ACL rule creation and during user custom role addition. Tests added

* GDB-10450 - add new labels to cypress locale file

* GDB-10450 - remove prefix warning dialogs and replace with inline text. Alter tests and remove unused templates and controllers.

* GDB-10450 - resize columns in ACL table

* GDB-10450 - refactor and create util class

(cherry picked from commit 229d163af4d954d5c7f671d43110fb3166b09a32)

### DIFF
--- a/src/css/aclmanagement.css
+++ b/src/css/aclmanagement.css
@@ -104,6 +104,11 @@ html, body {
     margin: 4px 0 0 3px;
 }
 
+.acl-management-view .acl-rules .role-cell .data .icon-warning {
+    float: right;
+    vertical-align: middle;
+}
+
 .acl-management-view .acl-rules .operation-column {
     width: 10%;
 }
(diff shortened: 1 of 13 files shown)


DesiBorisova:
Jenkins retry build

PR #1519, opened 2024-08-27
GDB-10785 - Rework and refactor existing solution for ACL role warnings on CUSTOM_ prefix use

DesiBorisova:
## What?
The functionality of the ACL role warnings will remain mostly unchanged. Only the input field border will be highlighted when CUSTOM_ prefix detected.

## Why?
The code needed to be reworked to reflect the review notes in MR https://github.com/Ontotext-AD/graphdb-workbench/pull/1511.

## How?
After discussing with team members, the changes proposed are in the existing custom-role-prefix.directive, which I extended to act as a validator in the cases when the user types "CUSTOM_". When the field is no longer present, I use the model to signal if a warning icon should be visible or not.
I made minor changes to function names and tests.
JSdoc for the directive updated.

### DIFF
--- a/src/css/aclmanagement.css
+++ b/src/css/aclmanagement.css
@@ -104,6 +104,12 @@ html, body {
     margin: 4px 0 0 3px;
 }
 
+#graphRoleCell, #pluginRoleCell, #systemRoleCell, #statementRoleCell {
+    overflow-wrap: anywhere;
+    display: flex;
+    align-items: center;
+}
+
 .acl-management-view .acl-rules .input-container {
     display: flex;
     align-items: center;
--- a/src/css/aclmanagement.css
+++ b/src/css/aclmanagement.css
@@ -104,10 +104,13 @@ html, body {
     margin: 4px 0 0 3px;
 }
 
-#graphRoleCell, #pluginRoleCell, #systemRoleCell, #statementRoleCell {
-    overflow-wrap: anywhere;
+.acl-management-view .acl-rules .preview-rule-row .role-icon-container {
     display: flex;
     align-items: center;
+}
+
+.acl-management-view .acl-rules .preview-rule-row .role-icon-container .acl-tag-role {
+    overflow-wrap: anywhere;
 }
 
 .acl-management-view .acl-rules .input-container {
--- a/src/css/aclmanagement.css
+++ b/src/css/aclmanagement.css
@@ -104,9 +104,13 @@ html, body {
     margin: 4px 0 0 3px;
 }
 
-.acl-management-view .acl-rules .role-cell .data .icon-warning {
-    float: right;
-    vertical-align: middle;
+.acl-management-view .acl-rules .input-container {
+    display: flex;
+    align-items: center;
+}
+
+.acl-management-view .acl-rules .role-cell .icon-warning {
+    margin-left: 3px;
 }
 
 .acl-management-view .acl-rules .operation-column {
@@ -252,6 +256,7 @@ input::placeholder {
 }
 
 .textarea-edit {
+    flex: 1;
     padding: 4px 8px;
     min-height: 29px;
 }
--- a/src/js/angular/aclmanagement/controllers.js
+++ b/src/js/angular/aclmanagement/controllers.js
@@ -4,7 +4,6 @@ import {mapAclRulesResponse} from "../rest/mappers/aclmanagement-mapper";
 import {isEqual} from 'lodash';
 import {mapNamespacesResponse} from "../rest/mappers/namespaces-mapper";
 import {ACL_SCOPE, DEFAULT_CONTEXT_VALUES, DEFAULT_URI_VALUES, DEFAULT_CLEAR_GRAPH_CONTEXT_VALUES} from "./model";
-import {RoleNamePrefixUtils} from "../utils/role-name-prefix-utils";
 
 const modules = [
     'graphdb.framework.rest.plugins.service',
@@ -138,7 +137,7 @@ function AclManagementCtrl($scope, $location, toastr, AclManagementRestService, 
 
     $scope.rowHeights = {};
 
-    $scope.doublePrefix = "CUSTOM_CUSTOM_";
+    $scope.hasCustomPrefix = false;
 
     //
     // Public functions
@@ -155,6 +154,7 @@ function AclManagementCtrl($scope, $location, toastr, AclManagementRestService, 
      */
     $scope.addRule = (scope, index) => {
         $scope.rulesModel.addRule(scope, index);
+        $scope.showPrefixWarningIcon(false);
         $scope.editedRuleIndex = index;
         $scope.editedRuleScope = scope;
         $scope.isNewRule = true;
@@ -311,8 +311,8 @@ function AclManagementCtrl($scope, $location, toastr, AclManagementRestService, 
         setModelDirty(scope);
     };
 
-    $scope.showWarning = function () {
-        $scope.showPrefixWarningIcon = true;
+    $scope.showPrefixWarningIcon = function (showWarning) {
+        $scope.hasCustomPrefix = showWarning;
     };
 
     /**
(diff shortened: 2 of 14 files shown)


src/js/angular/security/templates/user.html
```diff
@@ -219,14 +219,14 @@ <h3>{{'security.user.custom_role' | translate}}</h3>
...
-                                        ng-keydown="checkForBackspace($event)"
+                                        ng-keydown="checkUserInput($event)"
                                         ng-cut="removeErrorOnCut()"
                                         placeholder="{{'security.user.add.custom_role.msg' | translate}}"></tags-input>
                             <div class="small" ng-hide="isRoleValid">
                                 <small>{{'security.user.role.too.short' | translate}}</small>
                             </div>
                             <div class="small prefix-warning" ng-if="showCustomPrefixMessage">
-                                <small>{{'acl_management.rulestable.prefix_warning.text' | translate}}</small>
+                                <small>{{'acl_management.rulestable.custom_prefix_warning.text' | translate}}</small>
```
teodossidossev, line 229: use the error to display the message here

teodossidossev (changes requested):

DesiBorisova:
Jenkins retry build

src/js/angular/aclmanagement/controllers.js
```diff
@@ -311,8 +312,8 @@ function AclManagementCtrl($scope, $location, toastr, AclManagementRestService,
         setModelDirty(scope);
     };
 
-    $scope.showWarning = function () {
-        $scope.showPrefixWarningIcon = true;
+    $scope.showPrefixWarningIcon = function (showWarning) {
```
svilenvelikov, line 315: If this is not used in the template you should make it private function.

src/js/angular/aclmanagement/controllers.js
```diff
@@ -328,14 +329,11 @@ function AclManagementCtrl($scope, $location, toastr, AclManagementRestService,
...
     $scope.performSearchActionOnEnter = function (event, scope, form) {
-        if (!RoleNamePrefixUtils.isCustomPrefixUsed(event.target.value)) {
-            $scope.showPrefixWarningIcon = false;
-        }
         if (event.keyCode === 13) {
             event.stopPropagation();
             event.preventDefault();
             $scope.triggerValidation(form);
-            if (form.$valid) {
+            if (!form.$error.required || !form.$error.minLength) {
```
svilenvelikov, line 336: On a second thought, the idea to have the prefix validation logic in the directive doesn't look like a good idea. The problem now is that because of the new validation rule you make the model dirty and invalid, which enforces you to do the stuff like the one above. And I  don't like it. I someone adds a new validator someday, and that validator needs to make the model invalid in certain cases, then he/she must know that the code should also be changed somewhere else in order for the feature to work as expected. In other words, the specific new validator needs to be added here in this statement. This is bad because it couples the directive with the controller.

svilenvelikov: In my oppinion, if this new validation logic related with the custom_ prefix stays in the directive, then it should not change the model validity, but rather put some additional property like $warning instead and the UI logic (in the template) to use it instead. Right now, the field is highlighted as invalid, although according to the requirement, the presence of the custom_ prefix should trigger just a warning to appear.

src/js/angular/aclmanagement/directives/custom-role-prefix.directive.js
```diff
@@ -54,18 +57,47 @@ function customRolePrefixDirective() {
...
 
-                // Set up ngModel formatters and parsers for input element
-                ngModelCtrl.$formatters.push(function(modelValue) {
-                    return handleCustomPrefix(modelValue, false);
-                });
+                if (isNgTagsInput) {
+                    ngModelCtrl.$parsers.push(function (viewValue) {
+                        if (Array.isArray(viewValue)) {
+                            const isValid = viewValue.every((tag) => !tag.startsWith('CUSTOM_') && !tag.startsWith('!CUSTOM_'));
+                            ngModelCtrl.$setValidity('customPrefix', isValid);
```
svilenvelikov, line 70: Here, instead of altering ngModel, better add additional flag or object, like
```
warning = { 'customPrefix': { valid: isValid}}
```

src/js/angular/aclmanagement/directives/custom-role-prefix.directive.js
```diff
@@ -54,18 +57,47 @@ function customRolePrefixDirective() {
...
+                        }
+                        return viewValue;
+                    });
+
+                    scope.$watch(attrs.ngModel, function (newVal) {
```
svilenvelikov, line 75: Are you unsubscribing this watcher somewhere?

src/js/angular/aclmanagement/directives/custom-role-prefix.directive.js
```diff
@@ -54,18 +57,47 @@ function customRolePrefixDirective() {
...
                 const ngModelCtrl = element.controller('ngModel');
 
-                // Set up ngModel formatters and parsers for input element
-                ngModelCtrl.$formatters.push(function(modelValue) {
-                    return handleCustomPrefix(modelValue, false);
-                });
+                if (isNgTagsInput) {
+                    ngModelCtrl.$parsers.push(function (viewValue) {
+                        if (Array.isArray(viewValue)) {
+                            const isValid = viewValue.every((tag) => !tag.startsWith('CUSTOM_') && !tag.startsWith('!CUSTOM_'));
```
svilenvelikov, line 69: Can you please extract these as constants and reuse them

src/pages/aclmanagement.html
```diff
@@ -251,7 +260,7 @@ <h1 id="acl-management-view-title">
                                             </button>
                                             <button ng-click="saveRule(activeTabScope)"
                                                     class="btn btn-link save-rule-btn"
-                                                    ng-disabled="!ruleData.$valid"
+                                                    ng-disabled="ruleData.role.$error.required || ruleData.role.$error.minlength"
```
svilenvelikov, line 263: Wll, this is anot point for my doubts about altering the ngModel with this new validation. Here you are forced to do this in order to maintain the behavior in the template although it has nothing to do with the new feature

src/css/aclmanagement.css
```diff
@@ -104,6 +104,12 @@ html, body {
     margin: 4px 0 0 3px;
 }
 
+#graphRoleCell, #pluginRoleCell, #systemRoleCell, #statementRoleCell {
```
svilenvelikov, line 107: Put this in scope!

src/js/angular/aclmanagement/directives/custom-role-prefix.directive.js
```diff
@@ -44,19 +44,27 @@ function customRolePrefixDirective() {
     return {
         restrict: 'A',
         link: function(scope, element, attrs) {
+            const customPrefix = "CUSTOM_";
+            const negatedPrefix = "!CUSTOM_";
```
svilenvelikov, line 48: negatedCustomPrefix ?

src/pages/aclmanagement.html
```diff
@@ -107,15 +107,16 @@ <h1 id="acl-management-view-title">
                                             {{rule.policy}}
                                         </span>
                                     </td>
-                                    <td class="role-cell data">
+                                    <td id="statementRoleCell" class="role-cell data">
```
svilenvelikov, line 110: No. It's incorrect to have multiple elements with the same id on a page. Turn this to a css class instead!

svilenvelikov: apply this to others as well

src/css/aclmanagement.css
```diff
@@ -104,6 +104,12 @@ html, body {
...
+    overflow-wrap: anywhere;
```
svilenvelikov, line 108: Not sure why do you need this? You should be cautious with these properties because many of them are inheritted by the children and the things can get messy

DesiBorisova: My idea here was to override the `break-word` set for all the `td.data`, because I noticed that long role names with underscores (no hyphens or spaces) don't break and overflow into the next cell. But I don't know how realistic it is to expect that users will use only underscores.

svilenvelikov (approved):

DesiBorisova:
Jenkins retry build

teodossidossev (approved):

PR #1522, opened 2024-08-29
GDB-937 - Fix ruleset upload logic on invalid file selection

DesiBorisova:
## What?
When creating a repository and adding a custom ruleset, an invalid ruleset file followed by a valid one, will not stop the repository from being created. The valid ruleset file will be set.

## Why?
Uploading an invalid ruleset file caused an error and even though a valid file was later selected, the error persisted. This was due to the API response having status 200 on successful and failed uploads. Hence, the correct error handler was not triggered and the flag for invalid file remained set.

## How?
I set the flag to the correct value when the file is uploaded successfully. This allowed the repository to be created and then deleted without errors.

## Testing?
Test added.


teodossidossev (approved):

DesiBorisova:
Jenkins retry build

PR #1525, opened 2024-09-03
GDB-10709 agent list component

svilenvelikov:
## What
Introduce an agent list component.

## Why
By requirement

## What
Introduced a new component which renders a list with agents in the right sidebar.


svilenvelikov:
Jenkins retry build

PR #1529, opened 2024-09-05
GDB-10709 introduce a no agents view with create agent button and help message

svilenvelikov:
## What
Introduce a no agents view with create agent button and help message.
Agent list is filtered by the current repository by default.

## Why
This no agent view is rendered when there are no agents created yet and displays help messages to the user and provides a create agent button.

## How
Implemented a new component for the view and integrated with the page.
Changed the way how the chats and agents are initially loaded because when there are no agents, then the no agents view must be rendered and chats request may not be executed.
Implemented tests.
Add more tests.
Improved styling by adding more padding to list items and made the agent list scrollable only so that the filter menu is always visible.

### DIFF
--- a/src/i18n/locale-en.json
+++ b/src/i18n/locale-en.json
@@ -219,6 +219,7 @@
     "ttyg": {
         "helpInfo": "Talk to Your Graph is chat bot based on OpenAI's GPT-4 model that can answer questions about data stored in GraphDB, leveraging both your data and the general knowledge of the GPT-4 model. The bot works in connection with the ChatGPT Retrieval connector.",
         "hint": "Talk to your graph – simply ask a question!",
+        "loading": "Loading...",
         "help": {
             "btn": {
                 "show_help": {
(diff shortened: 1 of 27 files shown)


src/js/angular/ttyg/controllers/ttyg-view.controller.js
```diff
@@ -259,13 +262,30 @@ function TTYGViewCtrl($rootScope, $scope, $http, $timeout, $translate, $uibModal
     const loadChats = () => {
         $scope.loadingChats = true;
```
boyan-tonchev, line 263: I think it would be better if the "loadingChats" and "loadingAgents" flags are not set here. The methods "loadChats" and "loadAgents" return promises, so it would be more appropriate to set the flags where these methods are actually called, if needed. For example, "loadChats" is used in "onExportChat," where showing a loader is unnecessary.

src/js/angular/ttyg/controllers/ttyg-view.controller.js
```diff
@@ -329,7 +356,7 @@ function TTYGViewCtrl($rootScope, $scope, $http, $timeout, $translate, $uibModal
                 TTYGContextService.selectChat($scope.chats.getFirstChat());
             }
         }
-    };
```
boyan-tonchev, line 332: Revert

src/i18n/locale-en.json
```diff
@@ -219,6 +219,7 @@
     "ttyg": {
         "helpInfo": "Talk to Your Graph is chat bot based on OpenAI's GPT-4 model that can answer questions about data stored in GraphDB, leveraging both your data and the general knowledge of the GPT-4 model. The bot works in connection with the ChatGPT Retrieval connector.",
         "hint": "Talk to your graph – simply ask a question!",
```
boyan-tonchev, line 221: The translations also need to be added to "/test-cypress/fixtures/locale-en.json".

boyan-tonchev (approved):

PR #1536, opened 2024-09-12
feat(GDB-10850) Implement node remove functionality

teodossidossev:
## WHAT:
- Added a modal confirmation for deleting nodes using ModalService in the deleteNode method.
- Adjusted ClusterContextService methods to handle both Node and Location objects for cluster modifications.
- Refactored the saveNode method to optimize the logic for adding, restoring, or creating nodes in the cluster.
- Introduced handleNodesUpdate to ensure sequential processing of adding and removing nodes.
- Updated JSDocs across functions to improve clarity.

## WHY:
To support node deletion from the cluster, providing confirmation dialogs and optimized logic for cluster node handling.

## HOW:
- Introduced handleNodesUpdate, a new utility function that handles adding/removing nodes in sequence using Promises.
- Added ModalService for confirmation dialogs when deleting nodes and linked it with deleteFromCluster.
- Updated ClusterContextService to use Location.fromJSON for node location creation and enhanced its handling of addition and deletion logic.
- Updated template cluster-list.html to disable node deletion when there are fewer than 3 nodes and added node parameter in deleteNode calls.
- Refactored saveNode method to use a helper function findNodeByEndpoint for cleaner node lookup.

### DIFF
--- a/src/js/angular/clustermanagement/directives/cluster-list.directive.js
+++ b/src/js/angular/clustermanagement/directives/cluster-list.directive.js
@@ -4,9 +4,9 @@ const modules = [];
 angular.module('graphdb.framework.clustermanagement.directives.cluster-list', modules)
     .directive('clusterList', ClusterListComponent);
 
-ClusterListComponent.$inject = ['$translate', '$timeout', '$repositories', 'productInfo', 'toastr', 'RemoteLocationsService', 'ClusterContextService'];
+ClusterListComponent.$inject = ['$translate', '$timeout', '$repositories', 'productInfo', 'toastr', 'RemoteLocationsService', 'ClusterContextService', 'ModalService'];
 
-function ClusterListComponent($translate, $timeout, $repositories, productInfo, toastr, RemoteLocationsService, ClusterContextService) {
+function ClusterListComponent($translate, $timeout, $repositories, productInfo, toastr, RemoteLocationsService, ClusterContextService, ModalService) {
     return {
         restrict: 'E',
         templateUrl: 'js/angular/clustermanagement/templates/cluster-list.html',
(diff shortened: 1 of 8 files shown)


src/js/angular/clustermanagement/directives/cluster-list.directive.js
```diff
@@ -73,33 +73,45 @@ function ClusterListComponent($translate, $timeout, $repositories, productInfo,
...
-             * If the node is available in the cluster, it is added, otherwise, a new location is created and added.
+             * If the node is found in the cluster deletion list, it is restored.
+             * If the node is available in the cluster, it is added.
+             * Otherwise, a new location is created and added.
+             *
              * @param {string} endpoint - The endpoint of the node to save.
              */
             $scope.saveNode = (endpoint) => {
-                const node = $scope.cluster.getAvailable().filter((available) => available.endpoint === endpoint)[0];
+                const findNodeByEndpoint = (nodes) => nodes.find((node) => node.endpoint === endpoint)[0];
```
svilenvelikov, line 83: Why this logic is not in the cluster model?!

src/js/angular/clustermanagement/services/cluster-context.service.js
```diff
@@ -64,6 +63,36 @@ function ClusterContextService(EventEmitterService) {
...
+     * @param {Node} node - The node to restore.
+     * @throws {Error} Throws an error if the node is not provided or invalid.
+     * @return {void}
+     */
+    const restoreNode = (node) => {
+        if (!node) {
+            throw new Error("Invalid node");
+        }
+
+        const index = _clusterView.getDeleteFromCluster().findIndex((nodeToDelete) => nodeToDelete.endpoint === node.endpoint);
```
svilenvelikov, line 91: Here again. Move this `findIndex...` in the model to prevent abstraction to leak here and there and to make the code harder to be extended in the future

src/js/angular/clustermanagement/templates/cluster-list.html
```diff
@@ -57,7 +57,7 @@
                                 gdb-tooltip="{{'cluster_management.update_cluster_group_dialog.actions.replace_node' | translate}}">
                             <i class="fa-regular fa-arrow-right-arrow-left"></i>
                         </button>
-                        <button ng-click="deleteNode($index)" class="btn btn-link delete-node-btn secondary"
+                        <button ng-click="deleteNode($index, node)" ng-disabled="clusterView.length < 3" class="btn btn-link delete-node-btn secondary"
```
svilenvelikov, line 60: Is it possible this to be some property in the model and to be re/calculated when the model is changed?

src/js/angular/models/clustermanagement/cluster.js
```diff
@@ -290,23 +291,24 @@ export class ClusterViewModel {
...
+     * If the item is already marked for addition, it will be removed from the addition list.
+     * Otherwise, it will be added to the deletion list.
+     *
+     * @param {Node|Location} itemToDelete - The item to delete. This can either be a Node or Location object.
+     * @return {void}
      */
-    deleteFromCluster(node) {
-        this._deleteFromCluster.push(node);
+    deleteFromCluster(itemToDelete) {
+        const locationIndex = this._addToCluster.findIndex((location) => location.endpoint === itemToDelete.endpoint);
```
svilenvelikov, line 306: You can create some class which you can use instead of the plain js array for these addToCluster and deleteFromCluster and hide all these operations with these node lists which you do here and there

src/js/angular/clustermanagement/directives/cluster-list.directive.js
teodossidossev: done

src/js/angular/clustermanagement/services/cluster-context.service.js
teodossidossev: done

src/js/angular/clustermanagement/templates/cluster-list.html
teodossidossev: done

src/js/angular/models/clustermanagement/cluster.js
teodossidossev: yup

svilenvelikov (approved):

PR #1551, opened 2024-09-20
GDB-10714 - introduce agent edit action

svilenvelikov:
## What
Introduce agent edit action.

## Why
Users should be able to edit existing agents.

## How
Implemented mapper from the agent model to agent form model using the default values as a base. Wired the edit agent and open the agent settings modal with the needed data.

### DIFF
--- a/src/i18n/locale-en.json
+++ b/src/i18n/locale-en.json
@@ -287,7 +287,10 @@
             "deleting_agent": "Deleting agent...",
             "deleted_repository": "Deleted repository",
             "create_agent_modal": {
-                "title": "Create Agent",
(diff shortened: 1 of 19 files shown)


boyan-tonchev (approved):

src/i18n/locale-en.json
```diff
@@ -287,7 +287,10 @@
             "deleting_agent": "Deleting agent...",
             "deleted_repository": "Deleted repository",
             "create_agent_modal": {
-                "title": "Create Agent",
+                "title": {
```
teodossidossev, line 290: update test locales also

PR #1554, opened 2024-09-24
GDB-10772 - Show error message on resource edit when user not admin

DesiBorisova:
## What?
When editing a resource as an Admin and switching to a read only user, the page will show a message, instead of allowing the user to continue editing.

## Why?
Editing a resource as a non-admin user was allowed when the URL was accessed manually. The behavior should be that only admin users are allowed to edit resources.

## How?
I show an error message if the user does not have an admin role.

## Screeenshots?
Screenshot from 2024-09-24 15-01-20


boyan-tonchev (approved):

DesiBorisova:
Jenkins retry build

PR #1557, opened 2024-09-26
feat(GDB-10922): Apply visual changes

teodossidossev:
## WHAT:
- Fixed UI and content-related requirements in the cluster configuration module.
- Updated css, labels, tooltips,
- improved cluster node management with a restore node action

## WHY:
Requirements

## HOW:
- Updated HTML for "Edit" and "Create" modes.
- Added logic to track unsaved changes.
- Implemented restore node functionality.
- Adjusted CSS

### DIFF
--- a/src/css/cluster-list.css
+++ b/src/css/cluster-list.css
@@ -1,0 +1,94 @@
+.add-node-wrapper {
+    position: sticky;
+    top: 0;
+    z-index: 1000;
+    background-color: white;
+    width: 100%;
+    margin-top: 0;
+    text-align: right;
+    padding: 16px 12px 0 12px;
+}
+
+.cluster-list-modal {
+    padding: 0;
+}
+
+.modal-header {
+    padding: 15px 15px 0 15px;
+}
+
+.table-responsive {
+    padding: 0 12px 12px 12px;
+}
+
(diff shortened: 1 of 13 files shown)


boyan-tonchev (approved):

src/css/cluster-list.css
```diff
@@ -0,0 +1,94 @@
...
+    padding-left: 12px;
+    border-left: 1px solid #eceeef;
+
+}
+
+.location-cell {
+    width: 50%;
+    border-left: 1px solid #eceeef;
+
+    &.adding {
```
svilenvelikov, line 57: You know that, this is vanilla css and not sass, right?

src/css/cluster-list.css
```diff
@@ -0,0 +1,94 @@
...
+        width: 1400px;
+        max-width: 1400px;
+    }
+}
+
+.errors {
+    min-height: 36px;
+}
+
+.table td, .table th {
```
svilenvelikov, line 41: Please, try to always scope selectors under the directive's root element selector. Directives in angularjs does not use shadow dom, neither a virtual one, so in theory, its possible to get some styles overlapping

PR #1561, opened 2024-09-27
GDB-10717 Update the export chat call and functionality

yordanalexandrov:
## What
Updated the export chat endpoint and added logic for downloading the file

## Why
The endpoint for the export chat has been completed

## How
Changed the rest service call to GET
Set filename to default `chat-export` || extract it from `content-disposition` Added logic to `ttyg-view.controller` to use `saveAs` and save the blob

## Also in this MR
Fixed and enabled `chat-list` cypress spec

### DIFF
--- a/src/js/angular/rest/ttyg.rest.service.js
+++ b/src/js/angular/rest/ttyg.rest.service.js
@@ -53,23 +53,13 @@ function TTYGRestService($http) {
     /**
      * Exports a conversation by its ID.
      * @param {string} id
-     * @return {Promise<{Blob, string}>} Returns the conversation as a Blob and the filename wrapped in a Promise
+     * @return {Promise<{Blob}>} Returns the conversation as a Blob
      */
     const exportConversation = (id) => {
         if (DEVELOPMENT) {
             return _fakeBackend.exportConversation(id);
         }
-        return $http.get(`${CONVERSATIONS_ENDPOINT}/export/${id}`, {responseType: 'blob'}).then(function (res) {
-            const data = res.data;
-            const headers = res.headers();
-            const contentDispositionHeader = headers['content-disposition'];
-            let filename = 'chat-export';
-            if (contentDispositionHeader) {
-                filename = contentDispositionHeader.split('filename=')[1];
-                filename = filename.substring(0, filename.length);
-            }
-            return {data, filename};
-        });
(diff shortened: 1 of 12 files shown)


src/js/angular/rest/ttyg.rest.service.js
```diff
@@ -52,17 +52,24 @@ function TTYGRestService($http) {
...
      * @param {string} id
-     * @return {*}
+     * @return {Promise<{Blob, string}>} Returns the conversation as a Blob and the filename wrapped in a Promise
      */
     const exportConversation = (id) => {
         if (DEVELOPMENT) {
             return _fakeBackend.exportConversation(id);
         }
-        return $http.post(`${CONVERSATIONS_ENDPOINT}/export/${id}`);
+        return $http.get(`${CONVERSATIONS_ENDPOINT}/export/${id}`, {responseType: 'blob'}).then(function (res) {
```
boyan-tonchev, line 62: Could you move this response transformation into ttyg.service.js? We usually perform response conversions to models in that file.

yordanalexandrov: I completely missed that there is a service in the middle :(
Moved it to the service.

boyan-tonchev (dismissed):

yordanalexandrov:
Jenkins retry build

test-cypress/integration/ttyg/chat-list.spec.js
```diff
@@ -129,9 +141,40 @@ describe.skip('TTYG chat list', () => {
         TTYGViewSteps.getChatByDayGroups().should('have.length', 1);
     });
 
+    // FIXME: Skipped because of exception thrown by GDB
```
svilenvelikov, line 144: All TTYG UI functionality is tested with mocked backend because it is using an external service (the OpenAI API) which we can't rely on to be present always for our tests. That's why you should mock the export API for these tests as well.

svilenvelikov, line 144: What is the exception which GDB throws? For the TTYG tests we must mock all GDB API calls because it relies on an external service (the OpenAI API).

yordanalexandrov: Actually I got a bit confused.
I will remove the todos and enable the tests.

I have mocked them and they pass with the mocks.

The exception is due to a problem in how the response is sent by GraphDB, which is reported and will be fixed.

svilenvelikov (approved):

PR #1564, opened 2024-09-27
feat(GDB-10853) Remove old cluster management code

teodossidossev:
## WHAT
- Removed old cluster management code and added new tests.
- Deleted legacy controllers, templates, services, and associated Cypress tests related to the previous cluster management system.
- Added new tests to cover the updated functionality.

## WHY
The old cluster management system has been deprecated and replaced with a new implementation.

## HOW
- Deleted old Angular controllers, services, templates, and Cypress test files related to the previous cluster management.
- Updated i18n files to remove unnecessary translations.
- Added new Cypress tests and steps to validate the updated cluster management operations.


src/js/angular/clustermanagement/templates/cluster-list.html
```diff
@@ -209,7 +209,7 @@
                 <div class="indented-div">
                         <span class="padding-label wrapper-div"
                               gdb-tooltip="{{'cluster_management.cluster_page.advanced_options_tooltip' | translate}}">
-                            <button type="button" class="text-btn" data-toggle="collapse" style="text-align: left"
+                            <button type="button" class="text-btn advanced-options-btn" data-toggle="collapse" style="text-align: left"
```
svilenvelikov, line 212: it's minor, but this inline style can be moved in the stylesheet or replace it with the utility class `text-left`

src/js/angular/clustermanagement/templates/cluster-list.html
svilenvelikov, line 1: I'll add a note here because it's related with a code outside of the current chages scope.
Now that I've tested this a bit more, I see that in this `cluster-list` (which is also a weird name - it should've been named `node-list` or even better `cluster-configuration`) has some issues with the validation of the advanced option fields. If I try to create or edit a cluster, but just change these fields without changing the nodes, I can't save because the save button remains disabled.
Can you please verify that?

svilenvelikov (approved):

src/js/angular/clustermanagement/templates/cluster-list.html
teodossidossev: I'll handle the renaming in the 10733 branch

PR #1567, opened 2024-09-30
feat(GDB-10852) Add local node handling and tooltip for delete restrctions

teodossidossev:
## WHAT
- Implemented logic to identify the local node within the cluster.
- Enhanced the canDeleteNode function to restrict deletion if more than half of the healthy nodes are selected.
- Updated the UI to display a custom tooltip explaining why the delete button is disabled when applicable.
- Added translation keys and modified the cluster-list.html template to dynamically change tooltips.
- Updated CSS to style the tooltip when deletion is restricted.

## WHY
Ensuring that the system restricts users from deleting more than half of the healthy nodes to maintain quorum-based replication and prevent split-brain scenarios.

## HOW
- Modified the cluster model to set the local node's endpoint during initialization.
- Updated the canDeleteNode logic to enforce restrictions based on the total number of nodes in the cluster.
- Used dynamic tooltip messages in the HTML to show different messages depending on the button's enabled/disabled state.
- Added a CSS class to style the tooltip appropriately when deletion is restricted.
- Updated locale files

### DIFF
--- a/src/js/angular/clustermanagement/templates/cluster-list.html
+++ b/src/js/angular/clustermanagement/templates/cluster-list.html
@@ -48,14 +48,14 @@
                     </div>
                 </td>
                 <td class="info-cell data">
-                    <div ng-if="node.item.isLocal">
-                        {{'cluster_management.update_cluster_group_dialog.current_node' | translate}}
-                    </div>
+                    <span ng-if="node.isLocal">
+                        {{'cluster_management.update_cluster_group_dialog.current_node' | translate}}<span ng-if="node.item.nodeState">, </span>
+                    </span>
 
-                    <div ng-if="node.item.nodeState">
+                    <span ng-if="node.item.nodeState">
                         {{ ('cluster_management.update_cluster_group_dialog.' + node.item.nodeState.toLowerCase()) |
                         translate }}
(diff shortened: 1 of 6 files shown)


src/js/angular/clustermanagement/templates/cluster-list.html
```diff
@@ -125,9 +126,14 @@
                     </div>
                 </td>
                 <td class="info-cell data">
-                    <div ng-if="node.item.isLocal">
-                        {{'cluster_management.update_cluster_group_dialog.current_node' | translate}}
-                    </div>
+                    <span ng-if="node.isLocal">
+                        {{'cluster_management.update_cluster_group_dialog.current_node' | translate}}<span ng-if="node.item.nodeState">, </span>
```
svilenvelikov, line 130: umm, isn't it possible the comma to be added in the span below without ng-if?

svilenvelikov (approved):

src/js/angular/clustermanagement/templates/cluster-list.html
teodossidossev: No, because the span below can exist without this one

PR #1573, opened 2024-10-01
Gdb 10937 fixes links to create similarity and chatgpt pages

boyan-tonchev:
## What
Added confirmation dialog for repository change when the user clicks on a link redirecting to the create similarity page or create ChatGPT retrieval connector.

## Why
The both pages depends on the selected repository. To allow the user to create an index or connector in the repository chosen in the create agent dialog, the repository must be changed before navigating to the create similarity or create ChatGPT page if the repository selected in the workbench differs from the one in the dialog.

## How
Added a check to control when the confirmation dialog is displayed and when the user is redirected without prompting.

### DIFF
--- a/src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
+++ b/src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
@@ -132,7 +132,7 @@ function AgentSettingsModalController($scope, $uibModalInstance, SimilarityServi
         const message = decodeHTML(
             $translate.instant(
                 'ttyg.agent.create_agent_modal.form.fts_search.fts_disabled_message',
-                {repositoryEditPage: '#/repository/edit/' + $scope.activeRepositoryInfo.id}
+                {repositoryEditPage: '#/repository/edit/' + $scope.agentFormModel.repositoryId}
             )
         );
         return $sce.trustAsHtml(message);
--- a/src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
+++ b/src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
@@ -169,19 +169,10 @@ function AgentSettingsModalController(
     };
 
     /**
-     * Resolves the hint for the missing retrieval connector message. This is needed because the hint contains a html
-     * link that should be properly rendered.
-     * @return {*}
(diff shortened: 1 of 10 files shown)


src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
```diff
@@ -132,42 +155,24 @@ function AgentSettingsModalController($scope, $uibModalInstance, SimilarityServi
...
-        // The hint contains a html link which should be properly rendered.
-        const message = decodeHTML(
-            $translate.instant(
-                'ttyg.agent.create_agent_modal.form.retrieval_search.no_retrieval_connectors_message',
-                {retrievalConnectorPage: '#/connectors'}
-            )
-        );
-        return $sce.trustAsHtml(message);
+    $scope.goToConnectorsView = () => {
+        TTYGContextService.emit(TTYGEventName.GO_TO_CONNECTORS_VIEW, {repositoryId: $scope.agentFormModel.repositoryId});
```
svilenvelikov, line 175: but why?? Didn't we discussed that this controller is more like a smart component. You shoud've just did these confirmations here!

src/js/angular/ttyg/templates/modal/agent-settings-modal.html
```diff
@@ -134,8 +134,12 @@ <h4 class="modal-title">{{(isEdit ? 'ttyg.agent.create_agent_modal.title.edit '
 
                         <div ng-if="extractionMethod.method === extractionMethods.SIMILARITY"
                              class="extraction-method-options">
-                            <div ng-if="!similarityIndexes.length" class="no-similarity-index-message"
-                                 ng-bind-html="getNoSimilarityIndexHelpMessage()">
+                            <div ng-if="!similarityIndexes.length" class="no-similarity-index-message">
+                                {{'ttyg.agent.create_agent_modal.form.similarity_index.no_similarity_index_message.message_prefix' | translate}}
```
svilenvelikov, line 138: I don't like splitting these messages, but I guess that the link with the angular directive doesn't work if you put it in the label. Right?

svilenvelikov (approved):

svilenvelikov (approved):

PR #1574, opened 2024-10-01
GDB-10714 - introduce agent clone action

svilenvelikov:
## What
Introduce agent clone action.

## Why
Users should be able to clone existing agents.

## How
Implemented the handler for the clone action in the main controller. Extended the dialog to show appropriate labels according agent operation: edit, create or clone. Implemented tests.
Additionally fixed the agent model to agent form model mapper to properly set the sparqlOption property for the extraction methods.


src/js/angular/ttyg/controllers/ttyg-view.controller.js
```diff
@@ -276,6 +279,49 @@ function TTYGViewCtrl($rootScope, $scope, $http, $timeout, $translate, $uibModal
             });
     };
 
+    $scope.onCloneAgent = (agent) => {
+        let agentToClone = agent;
```
yordanalexandrov, line 283: What is the purpose of this assignment?
I see `onEditAgent` does it also.

src/js/angular/ttyg/controllers/ttyg-view.controller.js
```diff
@@ -276,6 +279,49 @@ function TTYGViewCtrl($rootScope, $scope, $http, $timeout, $translate, $uibModal
...
+        if (!agentToClone) {
+            agentToClone = TTYGContextService.getSelectedAgent();
+        }
+        const agentFormModel = agentFormModelMapper(agentToClone);
+        agentFormModel.name = `clone-${agentFormModel.name}`;
+        const activeRepositoryInfo = repositoryInfoMapper($repositories.getActiveRepositoryObject());
+        console.log(`clone agent`, agentFormModel, agent);
```
yordanalexandrov, line 290: Remove logging

src/js/angular/ttyg/controllers/ttyg-view.controller.js
```diff
@@ -276,6 +279,49 @@ function TTYGViewCtrl($rootScope, $scope, $http, $timeout, $translate, $uibModal
...
+            size: 'lg'
+        };
+        $uibModal.open(options).result.then(
+            // confirmed handler
+            (data) => {
+                TTYGService.createAgent(data)
+                    .then((updatedAgent) => {
+                        toastr.success($translate.instant("ttyg.agent.messages.agent_save_successfully", {agentName: updatedAgent.name}));
+                        const hasSelectedAgent = TTYGContextService.getSelectedAgent();
+                        if (hasSelectedAgent && data.id === hasSelectedAgent.id) {
```
yordanalexandrov, line 314: Could this actually happen?
Is the id not unique?

yordanalexandrov (approved):

src/js/angular/ttyg/controllers/ttyg-view.controller.js
svilenvelikov: My bad. For the edit operation it isneeded because you can trigger the edit operation via the edit button for an agent in the right sidebar, then you have the `agent` parameter, or via the edit button in the toolbar above the chat panel in which case you are editing the currently selected agent and in this case the `agent` parameter is undefined.
In the case for the clone operation though, this is not needed, because you can clone only through the menu on each agent onthe right sidebar.

svilenvelikov: This is documented in the jsdoc for the edit action handler also

src/js/angular/ttyg/controllers/ttyg-view.controller.js
svilenvelikov: It is. This is also not needed for the clone action.

svilenvelikov: removed

src/js/angular/ttyg/controllers/ttyg-view.controller.js
svilenvelikov: removed

PR #1575, opened 2024-10-02
Gdb 10733 rework cluster management modal view

teodossidossev:


teodossidossev:
used to fix commit lost issues in GDB-10733 branch

PR #1576, opened 2024-10-02
GDB-10950 Use GDB defaults when creating

yordanalexandrov:
## What
GDB API provides default values for agents which should be populated in the create/edit modals

## Why
This will move the responsibility for the default values to GDB where they matter

## How
- added a call to the `defaults` endpoint
- added a prop in the `TTYGContextService` to cache and hold the defaults value
- added a getter in `TTYGContextService` to get the cached default agent or to fetch it from GDB
- refactored the `agentFormModelMapper` to set default values where needed.
- removed hardcoded `AGENT_MODEL_DEFAULT_VALUES` along with the `newAgentFormModelProvider`
- moved default form model values inside the model and set them on initialization
- refactored `onCreateAgent`, `onEditAgent` to first fetch the default values
- added validation errors to `agentSettingsForm` in `agent-settings-modal.controller`, when there are no indexes or connectors


src/js/angular/ttyg/services/ttyg-context.service.js
```diff
@@ -36,6 +36,27 @@ function TTYGContextService(EventEmitterService) {
...
+    let _defaultAgent = undefined;
+
+    /**
+     * @return {Promise<AgentModel>}
+     */
+    const getDefaultAgent = () => {
+        if (_defaultAgent) {
+            return Promise.resolve(cloneDeep(_defaultAgent));
+        }
+        return TTYGService.getDefaultAgent()
```
svilenvelikov, line 53: I prefer this to be somewhere else, but not here. Perhaps the ttyg-view.controller is better place. The context service should not deal with loading data

src/js/angular/ttyg/controllers/ttyg-view.controller.js
```diff
@@ -272,38 +279,44 @@ function TTYGViewCtrl(
...
-            (data) => {
-                TTYGService.editAgent(data)
-                    .then((updatedAgent) => {
-                        toastr.success($translate.instant("ttyg.agent.messages.agent_save_successfully", {agentName: updatedAgent.name}));
-                        const hasSelectedAgent = TTYGContextService.getSelectedAgent();
-                        if (hasSelectedAgent && data.id === hasSelectedAgent.id) {
-                            TTYGContextService.selectAgent(updatedAgent);
+        TTYGContextService.getDefaultAgent()
+            .then((agentDefaultValues) => {
+                const agentFormModel = agentFormModelMapper(agentToEdit, agentDefaultValues);
```
svilenvelikov, line 284: I think it'd be better to call the ttyg.service directly and move the transformation there

src/js/angular/ttyg/services/ttyg-context.service.js
yordanalexandrov: Moved it to the `ttyg-view.controller`

src/js/angular/ttyg/services/ttyg-context.service.js
```diff
@@ -47,14 +47,11 @@ function TTYGContextService(EventEmitterService, TTYGService) {
...
     const getDefaultAgent = () => {
-        if (_defaultAgent) {
-            return Promise.resolve(cloneDeep(_defaultAgent));
-        }
-        return TTYGService.getDefaultAgent()
-            .then((response) => {
-                _defaultAgent = response;
-                return cloneDeep(_defaultAgent);
-            });
+        return _defaultAgent;
```
svilenvelikov, line 50: You forgot to clone it.

svilenvelikov (approved):

yordanalexandrov:
Jenkins retry build

svilenvelikov (dismissed):

svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

svilenvelikov (approved):

PR #1582, opened 2024-10-03
fix(GDB-10854) Remove unused locations

teodossidossev:
### WHAT
 Remove unused locations after cluster update. Locations with errors during creation are also removed.

 ## WHY
 To sanitize unused locations as they are unnecessary and slow down cluster configuration processes.

 ## HOW
 - Modified the `cluster-list.directive.js` and `cluster-management.controller.js` to delete locations after an error during the cluster update process and after cluster update.
 - Removed the unused methods.
 - Refactored `cluster-list.css` by removing nested rules left

### DIFF
--- a/src/js/angular/clustermanagement/controllers/cluster-management.controller.js
+++ b/src/js/angular/clustermanagement/controllers/cluster-management.controller.js
@@ -86,7 +86,9 @@ function ClusterManagementCtrl($scope, $http, $q, toastr, $repositories, $uibMod
     };
 
     $scope.showUpdateClusterGroupDialog = () => {
+        $scope.setLoader(true);
         getLocationsWithRpcAddresses().then(() => {
+            $scope.setLoader(false);
             return $uibModal.open({
                 templateUrl: 'js/angular/clustermanagement/templates/modal/update-cluster-group-dialog.html',
                 controller: 'UpdateClusterGroupDialogCtrl',
--- a/src/js/angular/clustermanagement/controllers/cluster-management.controller.js
+++ b/src/js/angular/clustermanagement/controllers/cluster-management.controller.js
@@ -108,6 +108,9 @@ function ClusterManagementCtrl($scope, $http, $q, toastr, $repositories, $uibMod
                 const nodes = cluster.getUpdateActions();
                 editCluster(nodes);
             }
+        }).catch((error) => {
+            $scope.setLoader(false);
+            console.error(error); // eslint-disable-line no-console
         }).finally(() => {
             getLocationsWithRpcAddresses();
         });
--- a/src/js/angular/clustermanagement/controllers/cluster-management.controller.js
+++ b/src/js/angular/clustermanagement/controllers/cluster-management.controller.js
@@ -203,7 +203,7 @@ function ClusterManagementCtrl($scope, $http, $q, toastr, $repositories, $uibMod
             })
             .finally(() => {
                 $scope.setLoader(false);
-                updateCluster(true);
+                updateCluster(true, true);
             });
     };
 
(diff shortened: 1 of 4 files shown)


src/js/angular/clustermanagement/directives/cluster-list.directive.js
```diff
@@ -103,6 +103,7 @@ function ClusterListComponent($translate, $timeout, $repositories, productInfo,
                         })
                         .catch((error) => {
                             handleErrors(error.data, error.status);
+                            $repositories.deleteLocation(endpoint);
```
yordanalexandrov, line 106: There is something odd here.
You add a location, which might fail, but this would still create a location, that you have to delete.

On line 237, the logic states that a location adding could be successful and still have errors.
This would not trigger the `catch` though.

yordanalexandrov: Also using the `deleteLocation` in the `$repositories` service, has some side effects.

If this is a new location, I guess you can safely delete it using the `LocationsRestService` through `RemoteLocationsService` (there is not delete fn yet there)

svilenvelikov: I also wanted to add something about the method on the line 237 because it's out of scope of this change, but still It probably should be fixed
The method is like this:
```
const addNewLocation = (newLocation) => {
                return RemoteLocationsService.addLocationHttp(newLocation)
                    .then((location) => {
                        if (location.error) {
                            handleErrors(location.error);
                            return;
                        }
                        if (location) {
                            return Location.fromJSON(location);
                        }
                    })
                    .catch((error) => {
                        throw error;
                    });
            };
```
the issue is that you throw a synchronous error from a promise chain which might be problematic and cause unhandled promise rejections. You better do something like this instead
```
catch((error) => {
  return Promise.reject(new Error(error));
})
```

teodossidossev: added

src/js/angular/clustermanagement/controllers/cluster-management.controller.js
```diff
@@ -86,7 +86,9 @@ function ClusterManagementCtrl($scope, $http, $q, toastr, $repositories, $uibMod
     };
 
     $scope.showUpdateClusterGroupDialog = () => {
+        $scope.setLoader(true);
         getLocationsWithRpcAddresses().then(() => {
+            $scope.setLoader(false);
```
svilenvelikov, line 91: If the request fails who will remove the loader?

src/js/angular/clustermanagement/directives/cluster-list.directive.js
teodossidossev: removed the checks on line 237, added the promise rejection

src/js/angular/clustermanagement/controllers/cluster-management.controller.js
teodossidossev: added catch to handle scenario

src/js/angular/clustermanagement/controllers/cluster-management.controller.js
```diff
@@ -108,6 +108,9 @@ function ClusterManagementCtrl($scope, $http, $q, toastr, $repositories, $uibMod
                 const nodes = cluster.getUpdateActions();
                 editCluster(nodes);
             }
+        }).catch((error) => {
+            $scope.setLoader(false);
```
svilenvelikov, line 112: Well. This will do the job, just we usually set these flags down in finally blocks. Don't change it anymore here.

PR #1584, opened 2024-10-04
GDB-10937 don't close create agent dialog on error

svilenvelikov:
## What
Changed the behavior on agent create/edit/clone operations so that the agent settings modal dialog is not closed in case of errors returned by the server.

## Why
This ensures better user experience by allowing the user just to correct some parameters thet might be incorrect or missing in case of error instead of beginning the entire process from the beginning.

## How
* Moved the agent operations handlers in the agent settings modal controller. The controller now is responsible to invoke the backend and to handle the case if there is error or not and to prevent automatic closing of the dialog in case of errors.
* Update the loader directive to allow the message to not be rendered.
* Updated the tests to handle the new behavior.
* Also updated all ttyg stubs by providing the request method as an explicit parameter to cy.intercept to prevent interceptors overriding.
* Additionally, added explicit waits for some requests after page visit to make the tests more robust.


svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

src/js/angular/rest/ttyg.rest.service.fake.backend.js
```diff
@@ -98,22 +102,27 @@ export class TtygRestServiceFakeBackend {
     }
 
     getAgents() {
-        return Promise.resolve({data: [...agentsList]});
+        return new Promise((resolve) => setTimeout(() => resolve({data: []}), LOAD_AGENTS_DELAY));
```
boyan-tonchev, line 105: We should return agentsList, not an empty list.

src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
```diff
@@ -224,6 +234,76 @@ function AgentSettingsModalController(
...
+     */
+    const createAgent = (newAgentPayload) => {
+        $scope.savingAgent = true;
+        return TTYGService.createAgent(newAgentPayload)
+            .then((agentModel) => {
+                $uibModalInstance.close(agentModel);
+                toastr.success($translate.instant('ttyg.agent.messages.agent_save_successfully', {agentName: agentModel.name}));
+            })
+            .catch((error) => {
+                toastr.error(getError(error, 0, 100));
```
boyan-tonchev, line 260: Maybe we should add an error message like in editAgent, instead of displaying what the backend returned.

src/js/angular/rest/ttyg.rest.service.fake.backend.js
svilenvelikov: i've forgotten to revert this here but I did it in my next MR

src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
svilenvelikov: I will add one in my next MR

PR #1586, opened 2024-10-05
Gdb 10937 remove tha chat panel when selected chat is deleted

boyan-tonchev:
10937: Clear chat panel when the selected chat is deleted

## What
The chat history in the chat panel was not cleared when the selected chat was deleted.

## Why
This behavior had not been implemented.

## How
Added a listener for chat deletion in the chat panel, and when the event occurs, the chat is removed from the panel.

## Additional work
Add human-readable label for query methods


test-cypress/integration/ttyg/chat-list.spec.js
svilenvelikov, line 1: part of the test extension below is redundant

src/js/angular/ttyg/directives/chat-panel.directive.js
```diff
@@ -193,8 +200,11 @@ function ChatPanelComponent(toastr, $translate, TTYGContextService) {
                 });
             };
 
-            const init = () => {
+            const init = (chat) => {
```
svilenvelikov, line 203: why you added these changes here? Better create some reset method or something, because this is confusing now

src/js/angular/ttyg/controllers/ttyg-view.controller.js
```diff
@@ -553,12 +553,14 @@ function TTYGViewCtrl(
      * @param {ChatModel} chat - the chat to be deleted.
      */
     const onDeleteChat = (chat) => {
+        TTYGContextService.deleteChat(chat);
```
svilenvelikov, line 556: Why? Shouldn't you remove the chat from the context when its actually deleted, and not removing it before the request and the adding it back if operation fails??

test-cypress/integration/ttyg/chat-list.spec.js
boyan-tonchev: Removed the redundant steps

svilenvelikov (approved):

svilenvelikov:
Jenkins retry build

PR #1589, opened 2024-10-07
fix(GDB-10375) Refactor language loading

teodossidossev:
## WHAT:
- Refactored the process for preloading and applying translations in the app.
- Updated the moduleDefinition function to accept preloaded translations.
- Improved language preloading logic to load multiple languages before initializing the workbench.
- Applied the font-display: swap property to all font-face definitions in CSS.

## WHY:
- To ensure translations are loaded before the app starts
- The font-display: swap property was added to optimize web font rendering particularly on slower networks.

## HOW:
- Introduced a new loadTranslationsBeforeWorkbenchStart function that handles the preloading of translation files and stores them in a translations object.
- Updated the startWorkbench function to accept the preloaded translation data and apply them through $translateProvider.
- Added fallback logic in case translation files are not preloaded.
- Updated all font-face rules in the CSS to use the font-display: swap property for better font rendering performance.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,10 +49,6 @@ pipeline {
       steps {
         withSonarQubeEnv('SonarCloud') {
           script {
-                    // Fix coverage permissions
-                    sh "sudo chown -R \$(id -u):\$(id -g) coverage/"
-                    sh "sudo chown -R \$(id -u):\$(id -g) cypress/"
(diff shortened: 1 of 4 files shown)


svilenvelikov (approved):

teodossidossev:
Jenkins retry build

PR #1590, opened 2024-10-07
Various TTYG fixes

avataar:
## What
- Changed date in chat list to ISO format (yyyy-mm-dd)
- Process and display multiple consecutive agent messages in chats
- Adjusted spacing between messages to better reflect flows with multiple consecutive agent messages and a look that looks more like a dialog instead of a question/answer list
- Text wrap markdown pre to prevent content from spilling out of the box
- Changed color of date/today/yesterday in left sidebar to gray-color-dark
- Changed color of repo in right sidebar to gray-color-dark
- Fixed flaky test
- Improvements to run-cypress-tests.sh so it can be run directly and provide more flexibility

## Why
- Improve usability
- Fix some agent messages not being shown
- Reliable tests needed

## How
- Some CSS/HTML changes
- Minor change in how chat messages are converted
- Minor test fix

### DIFF
--- a/src/js/angular/ttyg/templates/chat-item-detail.html
+++ b/src/js/angular/ttyg/templates/chat-item-detail.html
@@ -1,16 +1,16 @@
 <link href="css/ttyg/chat-item-details.css?v=[AIV]{version}[/AIV]" rel="stylesheet"/>
 
 <div class="chat-detail">
-    <div class="user-message">
+    <div class="user-message" ng-if="chatItemDetail.question">
         <div class="question alert-help">
             {{ chatItemDetail.question.message }}
         </div>
     </div>
-    <div class="assistant">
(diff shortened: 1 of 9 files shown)


src/js/angular/ttyg/templates/chat-item-detail.html
```diff
@@ -1,16 +1,16 @@
 <link href="css/ttyg/chat-item-details.css?v=[AIV]{version}[/AIV]" rel="stylesheet"/>
 
 <div class="chat-detail">
-    <div class="user-message">
+    <div class="user-message" ng-if="chatItemDetail.question">
```
svilenvelikov, line 4: For now this will do the job, but at some point we have to extend the model and avoid these checks here.

PR #1593, opened 2024-10-08
GDB-10937 - reset context and fix sparql method config

svilenvelikov:
## What
* Reset the ttyg context on destroy.
* Change the sparqlConstruct parameter for the sparql search method to sparqlQuery.
* Hide the edit agent settings action when there is no selected agent.

## Why
* The context needs to be cleared because it is implemented as a service and the services in angular are singletons and the properties in it remains when the page controller is destroyed which is unexpected.
* The edit agent settings button needs to be hid because when there is no selected agent, it doesn't behave properly.
* The sparqlQuery property was changed again due to changes in the backend.

## How
Applied the needed changes.

### DIFF
--- a/test-cypress/integration/ttyg/ttyg-view.spec.js
+++ b/test-cypress/integration/ttyg/ttyg-view.spec.js
@@ -38,7 +38,8 @@ describe('TTYG view', () => {
         // Verify the chat panel
         TTYGViewSteps.getChat().should('be.visible');
         TTYGViewSteps.getChatPanelToolbar().should('be.visible');
-        TTYGViewSteps.getEditCurrentAgentButton().should('be.visible');
+        // the edit agent settings button should not be visible when there is not a selected agent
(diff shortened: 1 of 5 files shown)


test-cypress/integration/ttyg/ttyg-view.spec.js
```diff
@@ -14,7 +14,7 @@ describe('TTYG view', () => {
         NamespaceStubs.stubNameSpaceResponse(repositoryId, '/namespaces/get-repository-starwars-namespaces.json');
     });
 
-    it('Should load ttyg page and render main components', () => {
+    it.only('Should load ttyg page and render main components', () => {
```
teodossidossev, line 17: is `only` intentionally?

svilenvelikov: no, will be removed

PR #1595, opened 2024-10-08
GDB-11001: The Explain Example button does not work

boyan-tonchev:
## What
- The "Explain Example" button is non-functional;
- A new chat field is not created when attempting to start a chat for the first time. ## Why
- The onAskHowDeliveredAnswer function binding is not passed to the chatItemDetail component;
- There was an issue where an action reset the question during the chat creation flow.

## How
- Added the missing binding for onAskHowDeliveredAnswer;
- Updated the chat creation flow. When the "Start a New Chat" button is clicked, or the "Ask" button is clicked when no chats exist, the ttygController now creates a dummy chat model instance and updates the context with it. Once the chat is created by the backend, the dummy chat is replaced with the newly created instance.

### DIFF
--- a/src/js/angular/models/ttyg/chats.js
+++ b/src/js/angular/models/ttyg/chats.js
@@ -165,6 +165,17 @@ export class ChatsListModel {
         this.updateChatsByDay();
     }
 
+    replaceChat(newChat, oldChat) {
+        this._chats = this._chats.map((chat) => {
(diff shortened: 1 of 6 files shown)


svilenvelikov (approved):

boyan-tonchev:
Jenkins retry build

PR #1597, opened 2024-10-08
GDB-11001: Filtering Repositories in TTYG Functionality

boyan-tonchev:
## What
Added filtering of repositories for:
 - The dropdown in the "Create Agent" dialog;
 - The repository filter in the agent list panel.

## Why
TTYG functionality is designed to work only with GraphDB repositories.

## How
Implemented filtering by repository type when loading repositories for dropdowns.

### DIFF
--- a/src/js/angular/ttyg/controllers/ttyg-view.controller.js
+++ b/src/js/angular/ttyg/controllers/ttyg-view.controller.js
@@ -640,8 +640,7 @@ function TTYGViewCtrl(
     const buildAgentsFilterModel = () => {
         const currentRepository = $repositories.getActiveRepository();
         // TODO: this should be refreshed automatically when the repositories change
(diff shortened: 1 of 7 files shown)


src/js/angular/ttyg/controllers/ttyg-view.controller.js
```diff
@@ -801,6 +803,7 @@ function TTYGViewCtrl(
 
     const buildRepositoryList = () => {
         $scope.activeRepositoryList = $repositories.getReadableRepositories()
+            .filter((repo) => repo.type === 'graphdb')
```
svilenvelikov, line 806: You can create another method in the repositories service like getReadableGraphdbRepositories or something and reuse it

boyan-tonchev: Done

svilenvelikov (approved):

PR #1598, opened 2024-10-09
GDB-10716: Add Visualization When Selected Agent is Changed

boyan-tonchev:
## What
Added an info message in the chat panel when the agent is changed, displaying the chosen agent for answering.

## Why
To improve the user experience.

## How
Inserted an info message between chat items that are handled by different agents.


boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1604, opened 2024-10-11
feat(GDB-11024) Add CookieService to manage installation ID tracking

teodossidossev:
## WHAT:
- Introduced a new CookieService for managing cookies
- Modified the license check logic to include setting or deleting cookies based on tracking permissions.
- Updated the run block in app.js to inject the new CookieService.
- Added handling of devMode to distinguish between development and production environments.

## WHY:
- To ensure tracking by using cookies
- The devMode flag was introduced to prevent tracking and cookie management in development environments, ensuring that tracking is only enabled in production.

## HOW:
- Created a CookieService that provides functions to get, set, and delete cookies.
- The cookie structure includes the version, installationId, and timestamp.
- Integrated the CookieService with the license checking logic, allowing cookies to be set or deleted based on whether tracking is permitted.
- The devMode flag was added to the application configuration, and its value (true for development and false for production) is injected dynamically in the HTML template.
- If devMode is true, tracking is disabled, and no cookies are set.
- Added Cypress tests to verify the correct behavior of cookies based on different license types, environments, and the devMode flag.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -42,6 +42,7 @@ const modules = [
     'graphdb.framework.core.directives.prop-indeterminate',
     'graphdb.framework.guides.services',
     'graphdb.framework.core.services.licenseService',
+    'graphdb.framework.core.services.cookieService',
     'graphdb.framework.core.directives.operationsstatusesmonitor',
     'graphdb.framework.core.directives.autocomplete',
     'ngCustomElement'
@@ -194,8 +195,8 @@ const moduleDefinition = function (productInfo, translations) {
     workbench.constant('productInfo', productInfo);
 
     // we need to inject $jwtAuth here in order to init the service before everything else
-    workbench.run(['$rootScope', '$route', 'toastr', '$sce', '$translate', 'ThemeService', 'WorkbenchSettingsStorageService', 'LSKeys', 'GuidesService', '$licenseService',
-        function ($rootScope, $route, toastr, $sce, $translate, ThemeService, WorkbenchSettingsStorageService, LSKeys, GuidesService, $licenseService) {
+    workbench.run(['$rootScope', '$route', 'toastr', '$sce', '$translate', 'ThemeService', 'WorkbenchSettingsStorageService', 'LSKeys', 'GuidesService', '$licenseService', 'CookieService',
+        function ($rootScope, $route, toastr, $sce, $translate, ThemeService, WorkbenchSettingsStorageService, LSKeys, GuidesService, $licenseService, CookieService) {
             $rootScope.$on('$routeChangeSuccess', function () {
                 updateTitleAndHelpInfo();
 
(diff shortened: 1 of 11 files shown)


src/js/angular/core/services/cookie.service.js
```diff
@@ -0,0 +1,76 @@
...
+    .service('CookieService', ['$document', CookieService]);
+
+/**
+ * Service to manage cookies for the application. It checks if a specific cookie exists,
+ * sets it if necessary, retrieves it, or deletes it when required.
+ *
+ * @param {Object} $document - AngularJS service to interact with the document object.
+ * @return {Object} - Returns an object with methods to get, set, and delete cookies.
+ */
+function CookieService($document) {
```
svilenvelikov, line 18: Why don't you use the angular's $cookie service? You can still wrap this around it

webpack.config.dev.js
```diff
@@ -34,7 +34,8 @@ module.exports = merge(commonConfig, {
             template: './src/template.html',
             favicon: 'src/img/icon.png',
             templateParameters: {
-                version: PACKAGE.version
+                version: PACKAGE.version,
+                devMode: true
```
svilenvelikov, line 38: you can add `devMode: false` in the production webpack config as well

src/js/angular/core/services/cookie.service.js
```diff
@@ -0,0 +1,76 @@
+const COOKIE_NAME = '_wb=';
+const COOKIE_VERSION = 'WB1';
+const PATH = "; path=/";
+const EXPIRATION = "; expires=Fri, 31 Dec 2099 23:59:59 GMT";
+const EXPIRED = "; Max-Age=-99999999";
```
svilenvelikov, line 5: Write documentation for these constants above because they are specific and related to this concrete cookie.

src/app.js
```diff
@@ -224,8 +225,15 @@ const moduleDefinition = function (productInfo, translations) {
...
 
-            // Checks license status and adds tracking code when free/evaluation license
-            $licenseService.checkLicenseStatus();
+            // Checks license status and adds tracking code and cookies when free/evaluation license
+            $licenseService.checkLicenseStatus().then(() => {
+                if ($licenseService.isTrackingAllowed) {
+                    const installationId = $licenseService.license().installationId || '';
+                    CookieService.setCookieIfAbsent(installationId);
+                } else {
+                    $licenseService.deleteCookie();
```
svilenvelikov, line 234: ummm, not sure about this one

src/js/angular/core/services/cookie.service.js
teodossidossev: in version 1.3.8 it is quite limited from to functionality. That's why I wrote our generic cookie service.

webpack.config.dev.js
teodossidossev: done

src/app.js
teodossidossev: fixed

src/js/angular/core/services/cookie.service.js
teodossidossev: done

svilenvelikov (approved):

PR #1605, opened 2024-10-11
Gdb 10937 update explain functionality with the new model

boyan-tonchev:
## What
The explain response models have been updated.

## Why
The backend exposes new fields in the response, providing an opportunity for further UX improvements.

## How
The UI models have been updated to reflect the changes.

## Additional work
- The actions remain visible until the explain response is expanded;
- The raw query is hidden if it is identical to the query;
- The raw query text ellipsis when the query is too long;
- Hide the "Copy SPARQL Query" button if the query in queryMethod is not SPARQL;
- Change the name of the query label depending on the query type: "JSON" or "SPARQL."

### DIFF
--- a/test-cypress/stubs/ttyg/ttyg-stubs.js
+++ b/test-cypress/stubs/ttyg/ttyg-stubs.js
@@ -139,117 +139,17 @@ export class TTYGStubs extends Stubs {
         }).as('get-agent-defaults');
     }
 
-    static stubAnswerQuestion() {
-        cy.intercept({
-            method: 'POST',
-            url: '/rest/chat/conversations'
-        }, (req) => {
-            const requestBody = req.body;
-
-            const answer = {
-                id: requestBody.conversationId,
-                name: "Han Solo is a character",
-                timestamp: Math.floor(Date.now() / 1000),
-                messages: [
-                    {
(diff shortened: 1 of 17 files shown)


test-cypress/stubs/ttyg/ttyg-stubs.js
```diff
@@ -178,4 +178,78 @@ export class TTYGStubs extends Stubs {
...
+        cy.intercept({
+            method: 'POST',
+            url: 'rest/chat/conversations/explain'
+        }, (req) => {
+            const requestBody = req.body;
+
+            const explainResponse = {
+                conversationId: requestBody.conversationId,
+                answerId: requestBody.answerId,
+                queryMethods: [
```
svilenvelikov, line 192: You are often using such smarter stubs but I don't like them. They are harder to be maintained and updated.

svilenvelikov: why don't you just use fixtures?

boyan-tonchev: but they are closer to how the real system works.

boyan-tonchev: Done

svilenvelikov (approved):

PR #1614, opened 2024-10-15
GDB-10937 create agent form validation fix and copy property btn

svilenvelikov:
## What
* Fixed validation of the similarity search method form in agent settings dialog when the method is selected.
* Fixed the placeholder label key for a field.
* Improved the warning message that appears in the no agents view to allow gpt token property to be copied via button.
* Fixed slider positioning for the similarity index threshold field.

## Why
These will improve the usability.

## How
Removed the incorrect dynamic css class added to the similarity method panel.
Fixed the label of a placeholder.
Extended the warning message that appears on the no agents view.

### DIFF
--- a/src/js/angular/core/directives/copy-to-clipboard/copy-to-clipboard.directive.js
+++ b/src/js/angular/core/directives/copy-to-clipboard/copy-to-clipboard.directive.js
@@ -41,7 +41,25 @@ copyToClipboard.$inject = ['$translate', 'toastr'];
  */
 function copyToClipboard($translate, toastr) {
     return {
-        template: '<button class="copy-btn" gdb-tooltip="{{tooltipText | translate}}" ng-click="copyToClipboard()"><i class="fa-regular fa-clone"></i></button>',
+        template: `
+            <style>
+                .copy-btn {
+                    border: none;
+                    background-color: transparent;
+                    cursor: pointer;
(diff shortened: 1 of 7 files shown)


src/js/angular/core/directives/copy-to-clipboard/copy-to-clipboard.directive.js
```diff
@@ -41,7 +41,25 @@ copyToClipboard.$inject = ['$translate', 'toastr'];
...
+                }
+                .copy-btn:hover {
+                    transform: scale(1.1);
+                    transition: all 0.1s ease-out;
+                }
+                .copy-btn:focus {
+                    outline: none;
+                }
+            </style>
+            <button class="copy-btn" gdb-tooltip="{{tooltipText | translate}}" ng-click="copyToClipboard()"><i class="fa-regular fa-clone"></i></button>
```
boyan-tonchev, line 61: We have similar styling applied in 'agent-settings-modal.css' and 'chat-item-details'. We could clean them up.

svilenvelikov: they are different, maybe the color and others - I haven't compared them

boyan-tonchev (approved):

boyan-tonchev (approved):

PR #1617, opened 2024-10-16
feat(GDB-11024) Add Cookie Consent and Policy Functionality

teodossidossev:
## WHAT:
- Implemented cookie consent modal and cookie policy page.
- Added cookie-related translations in English and French locales.
- Updated the jwt-auth.service.js to handle user consent for cookies.
- Added Cypress tests to validate cookie consent behavior.
- Introduced new CSS styles for cookie consent and policy modals.
- Added tracking for user cookie consent in user settings.

## WHY:
Requirements

## HOW:
- Created a new directive for cookie consent (cookie-consent.directive.js) and templates for cookie policy and cookie consent.
- Updated user settings to include a cookie consent flag.
- Refactored jwt-auth.service.js
- Added Cypress tests to validate the consent modal, cookie policy link, and proper settings update.
- Introduced new translations and styling for cookie consent and policy modals.


src/js/angular/core/directives/cookie-policy/cookie-consent.directive.js
```diff
@@ -0,0 +1,72 @@
...
+                }).result.then(() => $scope.acceptConsent());
+            };
+
+            // =========================
+            // Private functions
+            // =========================
+            const init = () => {
+                $licenseService.checkLicenseStatus().then(() => {
+                    if ($licenseService.isTrackingAllowed()) {
+                        $jwtAuth.getPrincipal().then((data) => {
```
svilenvelikov, line 54: Better avoid nesting promises. E.g.
```
$licenseService.checkLicenseStatus().then(() => {
    if ($licenseService.isTrackingAllowed()) {
        return $jwtAuth.getPrincipal();
    }
}).then((data) => {
    if (!data) {
        return;
    }
    appSettings = data.appSettings;
    username = data.username;

    if (!appSettings.COOKIE_CONSENT) {
        $scope.showCookieConsent = true;
    }
});
```

svilenvelikov: Also add a catch statement

src/js/angular/security/controllers.js
```diff
@@ -931,6 +931,24 @@ securityCtrl.controller('ChangeUserPasswordSettingsCtrl', ['$scope', 'toastr', '
             ThemeService.applyTheme(theme.name);
         };
 
+        $licenseService.checkLicenseStatus().then(() => {
```
svilenvelikov, line 934: Can you move this in the init or put it in a method and call it from the init

test-cypress/integration/home/workbench.home.spec.js
```diff
@@ -350,6 +352,64 @@ describe('Home screen validation', () => {
         });
     });
 
+    context.only('Cookie policy', () => {
```
svilenvelikov, line 355: only

src/js/angular/security/controllers.js
svilenvelikov: Also catch is missing

src/template.html
```diff
@@ -158,6 +158,7 @@
                 {{sesameVersion}} &bull; Connectors {{connectorsVersion}} &bull; Workbench {{workbenchVersion}} &bull; &copy; {{previousYear}}&ndash;{{currentYear}} <a href="http://ontotext.com" target="_blank" rel="noopener noreferrer">Ontotext AD</a>. {{'all.rights.reserved.label' | translate}}
             </p>
         </div>
+        <cookie-consent></cookie-consent>
```
svilenvelikov, line 161: Not sure if this should be here or in the user template where the actual link which opens the colicy dialog

src/js/angular/core/services/jwt-auth.service.js
svilenvelikov, line 1: I'm not sure if I can validate these changes TBH. IMO this is better to be extracted in a separate MR and to introduce it in next version and here to be introduced only the needed changes for this feature.

svilenvelikov (changes requested):

src/js/angular/core/services/jwt-auth.service.js
```diff
@@ -3,294 +3,302 @@ import 'angular/core/services/openid-auth.service.js';
...
+            let _hasOverrideAuth = false;
+            let _externalAuth = undefined;
+            let _authImplementation = undefined;
+            let _openIDEnabled = false;
+            let _passwordLoginEnabled = undefined;
+            let _openIDConfig = undefined;
+            let _gdbUrl = undefined;
+
+            // =========================
+            // RootScope functions
```
svilenvelikov, line 51: All these rootScope shits usually should be just public functions exposed by the service and used wherever needed in controllers. I'm not asking you to do do this, but only pointing out a really bad case of tight coupling .

src/js/angular/core/services/jwt-auth.service.js
```diff
@@ -3,294 +3,302 @@ import 'angular/core/services/openid-auth.service.js';
...
                 if (noaccess) {
-                    $location.search('noaccess');
+                    $location.search(NO_ACCESS);
                 } else if (expired) {
-                    $location.search('expired');
+                    $location.search(EXPIRED);
                 }
                 // Countering race condition. When the unauthorized interceptor catches error 401 or 409, then we must make
                 // sure that a request is made to access the login page before proceeding with the rejection of the
                 // original request. Otherwise the login page is not accessible in the context of spring security.
```
svilenvelikov, line 101: Nice. A simple thing like the logout tuned into a nightmare. There is a recent bug about this BTW.

src/js/angular/core/services/jwt-auth.service.js
```diff
@@ -3,294 +3,302 @@ import 'angular/core/services/openid-auth.service.js';
...
+            // =========================
+            // Public functions
+            // =========================
+            /**
+             * Checks if the current user is authenticated.
+             * @return {boolean} True if the user is authenticated.
+             */
+            const isAuthenticated = () => !_securityEnabled || _hasExplicitAuthentication();
+
+            const _hasExplicitAuthentication = () => {
```
svilenvelikov, line 124: What it the matter with these underscores BTW. In the context of classes and constructor functions it is a convention to denote private functions usually but here you've put them in the public section just because you then reexposed them via some other unprefixed functions. This clearly can be simplified.

src/js/angular/core/services/jwt-auth.service.js
```diff
@@ -3,294 +3,302 @@ import 'angular/core/services/openid-auth.service.js';
...
-                        }
-                    }
-                });
-            };
+            const hasExternalAuthUser = () => _externalAuthUser;
 
-            this.initSecurity = function () {
-                this.securityInitialized = false;
+            /**
+             * Initializes security by retrieving the security configuration and authenticating the user.
```
svilenvelikov, line 153: I wish it was so simple, but this function is just plain stupid and does a ton of mindbogling stuff and has a flow that is beyond my abilities to follow.

src/template.html
teodossidossev: should be able to display on every page - this is the consent popup

src/js/angular/core/directives/cookie-policy/cookie-consent.directive.js
teodossidossev: done

test-cypress/integration/home/workbench.home.spec.js
teodossidossev: removed

src/js/angular/core/services/jwt-auth.service.js
teodossidossev: the goal was in the initial MR the changes to be minimal - to separate the private properties from the public ones. In the next iteration to be simplified. But this is already the subject of the next MR.

src/js/angular/security/controllers.js
teodossidossev: done

svilenvelikov (dismissed):

svilenvelikov (approved):

PR #1622, opened 2024-10-17
GDB-10937 show warnings in agent settings modal

svilenvelikov:
## What
* Show warning when the user sets temperature higher than 1.
* Show warning and an alert to the user when he tries to override the base instructions.

## Why
* Because it would help the user to receive more meaningful results from the agent.
* Because this would prevent unexpected results if the user provides incorrect instructions.

## How
* Implemented icon with tooltip next to the temperature field label, and a handler for the property change to toggle the warning.
* Implemented icon with tooltip and an alert that is displayed when the field value becomes different from empty string.

### DIFF
--- a/src/css/ttyg/agent-settings-modal.css
+++ b/src/css/ttyg/agent-settings-modal.css
@@ -1,5 +1,7 @@
 :root {
     --field-label-weight: 500;
+    --error-color: #FA787E;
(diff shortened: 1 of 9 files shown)


boyan-tonchev:
Jenkins retry build

boyan-tonchev (approved):

PR #1623, opened 2024-10-18
GDB-11070 - improve logic for menu toggle and main content resize

DesiBorisova:
## What?
When the side main menu is collapsed, the main content will always remain at expanded width, even on page refresh.

## Why?
On page refresh/load, the main content wouldn't apply its expanded class and there would be a gap between the collapsed menu and the main content.

## How?
I reworked the logic and added a menu check on page load, while keeping the checks on menu toggle.

## Screenshots?
The gap remains like this when the page loads and the menu is collapsed.
Screenshot from 2024-10-18 09-37-20

### DIFF
--- a/src/js/angular/controllers.js
+++ b/src/js/angular/controllers.js
@@ -135,11 +135,30 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repos
     $scope.descr = $translate.instant('main.gdb.description');
     $scope.documentation = '';
     $scope.menu = $menuItems;
+    $scope.menuCollapsed = false;
     $scope.tutorialState = LocalStorageAdapter.get(LSKeys.TUTORIAL_STATE) !== 1;
     $scope.userLoggedIn = false;
     $scope.embedded = $location.search().embedded;
     $scope.productInfo = productInfo;
     $scope.guidePaused = 'true' === LocalStorageAdapter.get(GUIDE_PAUSE);
+
+    // Check on page load
+    $scope.isMenuCollapsedOnLoad = function () {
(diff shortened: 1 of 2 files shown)


src/js/angular/controllers.js
```diff
@@ -135,12 +135,31 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repos
...
+        return $('.main-menu').hasClass('collapsed');
+    };
+
+    $timeout(function () {
+        $scope.menuCollapsed = $scope.isMenuCollapsedOnLoad();
+    }, 0);
+
+    // Check after initial load
+    $scope.checkMenu = debounce(function () {
+        return $('.main-menu').hasClass('collapsed');
```
svilenvelikov, line 156: can't you use the isMenuCollapsedOnLoad function here?

src/js/angular/controllers.js
```diff
@@ -135,12 +135,37 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repos
...
     $scope.productInfo = productInfo;
     $scope.guidePaused = 'true' === LocalStorageAdapter.get(GUIDE_PAUSE);
 
+    $scope.isMenuCollapsedOnLoad = function () {
+        return $('.main-menu').hasClass('collapsed');
+    };
+
+    $scope.checkMenu = debounce(function () {
+        const collapsed = $scope.isMenuCollapsedOnLoad();
+        if ($scope.menuCollapsed !== collapsed) {
```
svilenvelikov, line 151: this if statement is redundant, this can be just `$scope.menuCollapsed = collapsed`

PR #1629, opened 2024-10-21
GDB-10937 handle error on missing chat and other visual issues

svilenvelikov:
## What
* Error in console when previously selected hat was deleted and page reloaded.
* Fixed issues with the chat panel vertical offset which caused the panel to overlap with the page footer when the page help is opened.

## Why
* When chat was selected but then deleted, it's id stays in the localstorage and when the page is loaded, then controller tries to load the chat which is no longer in the chat list.
* The chat panel offset did not account the case where the help message might occupy additional vertical space.

## How
* Added check in controller to prevent error when a chat that was persisted in the local storage but was in fact deleted from the chat list.
* Added dynamic vertical offset calculation based on if the help message is opened or not.


boyan-tonchev (dismissed):

svilenvelikov:
Jenkins retry build

boyan-tonchev (approved):

PR #1631, opened 2024-10-21
GDB-10925 Optimized the RDF resource search directive

boyan-tonchev:
## What
There are multiple backend calls being made to fetch namespaces and check if autocomplete is enabled.

## Why
The rdfResourceSearch directive is currently designed to call these endpoints whenever the "repositoryIsSet" or "autocompleteStatus" events are emitted. However, these results are not always needed because the directive is closed most of the time. In some views, this isn't an issue as the events are fired once when the page loads, but in others, like "Repository", the "repositoryIsSet" event is triggered every 5 seconds, causing unnecessary data to be loaded repeatedly.

## How
- Removed the "repositoryIsSet" and "autocompleteStatus" listeners;
- Refactored the code to manage the show and hide functionality for the RDF resource search input. The backend calls are now triggered only when the user clicks on the search icon, eliminating unnecessary requests to the backend.


boyan-tonchev:
Jenkins retry build

boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1633, opened 2024-10-22
GDB-10937 introduce agent instructions explain dialog

svilenvelikov:
## What
Introduce agent instructions explain dialog where the user can see a preview of each of the agent settings.

## Why
Requirement.

## How
* Implemented a new modal dialog component for the agent instructions preview and wired it to be opened through a button in the agent settings modal.
* Rest client, service, models and a mapper were implemented as well.


src/js/angular/ttyg/templates/modal/agent-instructions-explain-modal.html
```diff
@@ -0,0 +1,25 @@
...
+            </span>
+            <div class="instruction-content">
+                {{instruction.content}}
+            </div>
+        </div>
+    </div>
+</div>
+
+<div class="modal-footer">
+    <button type="button" class="btn btn-secondary" ng-click="onCloseAgentInstructionsExplainModal()">Close</button>
```
boyan-tonchev, line 24: translate the button label

PR #1635, opened 2024-10-23
GDB-10937: Added the time when the question was asked to the tooltips for the agent icons and the questions.

boyan-tonchev:
## What
Added the time when the question was asked to the tooltips for the agent icons and the questions.

## Why
To improve user experience (UX).

## How
- Added a GraphDB tooltip to the agent icons and the questions.
- Enhanced the GraphDB tooltip to support HTML.
- Removed the htmlTranslate filter (it had issues with translations).
- Introduced a new trustAsHtml filter.


boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1642, opened 2024-10-24
GDB-10937: Incorrect Agent Settings Validation

boyan-tonchev:
## What
There is a scenario where the agent settings dialog appears valid, but the "Create agent"/"Save agent" button remains inactive. The steps to reproduce the issue are:
 - Open the dialog, fill in the name, choose a repository, activate the "SPARQL query" method, and select any radio button. At this point, the "Create agent"/"Save agent" button becomes enabled;
 - Next, enable either the "Similarity search" or "ChatGPT Retrieval connector" query method. The button will become disabled;
 - Disable the query method from the previous step, but the button remains disabled.

## Why
When either the "Similarity search" or "ChatGPT Retrieval connector" method is enabled, a validation check is performed to ensure that indexes or connectors are selected. However, when these methods are disabled, the validation errors are not cleared, causing the button to remain disabled even though the validation should have been reset.

## How
Validation errors are now cleared when a method is disabled, ensuring the "Save" button's state updates correctly.


svilenvelikov (dismissed):

boyan-tonchev:
Jenkins retry build

PR #1653, opened 2024-10-29
GDB-11131: Fixes max width for explanation messages

boyan-tonchev:
## What
Addresses an issue where the chat panel sometimes displays a horizontal scroll.

## Why
Typically, the backend returns formatted assistant queries, which the markdown library displays correctly with line breaks and tabs. Occasionally, however, the backend returns a single, continuous string without line breaks or tabs, causing excessively long lines that expand the container width and introduce horizontal scrolling.

## How
Set a maximum width on the container displaying explanation messages, which forces the browser to wrap long query strings correctly.

### DIFF
--- a/src/css/ttyg/chat-item-details.css
+++ b/src/css/ttyg/chat-item-details.css
@@ -1,10 +1,12 @@
 :root {
     --border: 1px solid #C2C1C2;
+    --assistan-icon-weight: 2.5rem;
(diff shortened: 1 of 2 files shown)


boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

PR #1656, opened 2024-10-29
Fixed various issues with FontAwesome vs our icons

avataar:
- FontAwesome icons are scaled globally (also added a preview in ux-test2)
- FontAwesome icons use only the fa class (and not fa-regular or alternatives like fa-thin) and the font weight is determined by the --fa-style variable.
- Removed various hacks for individual icons
- Removed FontAwesome 4 from package list
  - There was one icon in export view using it
  - Note that this npm module brought the classes pull-right and pull-left, which now need to be defined explicitly now
- Copy button directive and Open in SPARQL directive are now styled like all similar icon buttons inside the directive
- Open in SPARQL directive uses a composite FontAwesome icon to recreate the {...} icon from our font and match the rest of the FontAwesome icons around it

Other fixes:
- Added max-width to .assistant-message in TTYG to prevent long native queries from spilling to the right in explains
- Added font-variant-ligatures: none to `<pre>` in TTYG explain to prevent queries showing fancy ligatures for some character combinations

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,6 @@
                 "autofill-event": "0.0.1",
                 "d3": "^7.8.5",
                 "echarts": "^5.4.3",
-                "font-awesome": "^4.7.0",
                 "jquery": "^3.4.1",
                 "jsonld": "^8.3.2",
                 "jsrsasign": "^11.0.0",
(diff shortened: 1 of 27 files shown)


teodossidossev (commented):
LGTM. Did not tested TTYG

PR #1660, opened 2024-10-29
GDB-10937: Automatically Update Validation in Create/Edit Agent Dialog

boyan-tonchev:
## What
The validation for the create/edit agent dialog is refreshed when the dialog’s tab becomes active.

## Why
When users open the create/edit agent dialog and select the "FTS," "Similarity," or "ChatGPT" execution method, a help message appears if the required settings are not yet configured in GraphDB. This message provides a link to the appropriate view in the workbench. After users navigate to that view and complete the necessary setup, they can return to the dialog tab, where the validation will update automatically to reflect these changes, eliminating the need for additional manual refresh actions.

## How
An event listener was added to detect when the user reactivates the tab containing the open create/edit agent dialog. When this event occurs, the dialog’s validation is automatically refreshed to incorporate any changes the user made.

## Additional work
Added validation to the Full-text Search (FTS) query method.

### DIFF
--- a/src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
+++ b/src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
@@ -241,6 +241,9 @@ function AgentSettingsModalController(
      */
     $scope.updateSimilaritySearchPanel = (clearIndexSelection = false) => {
         const similaritySearchExtractionMethod = $scope.agentFormModel.assistantExtractionMethods.getSimilarityExtractionMethod();
+        if (!similaritySearchExtractionMethod.selected) {
+            return;
+        }
         if (clearIndexSelection) {
             similaritySearchExtractionMethod.similarityIndex = null;
         }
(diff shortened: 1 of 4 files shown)


src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
```diff
@@ -262,8 +268,19 @@ function AgentSettingsModalController(
 
     $scope.checkIfFTSEnabled = () => {
         if (!$scope.agentFormModel.repositoryId) {
+            $scope.agentSettingsForm.$setValidity('FTSDisabled', false);
```
svilenvelikov, line 271: You must add this in the template for the FTS method as well `ng-class="{'has-error': agentSettingsForm.$error.FTSDisabled}"`

svilenvelikov: Now I see that this is missing also for the similarity search panel. Add it there too please

svilenvelikov (approved):

PR #1665, opened 2024-10-30
GDB-11114: License appears editable regardless than GraphDB configured through a file

boyan-tonchev:
GDB-11114: License appears editable regardless than GraphDB configured through a file

What
The buttons "Revert to Free License" and "Set New License" are visible, and the informational message, "License cannot be changed from the Workbench as it was set ..." does not appear, regardless than GraphDB has a license set via the configuration graphdb.license.file.

Why
This occurs due to an incorrect check following the server response.

How
The license configuration check has been corrected to address this issue.

(cherry picked from commit f298e2c4523d2b9fcecd96f032b996729e3ddf72)


boyan-tonchev:
Jenkins retry build

PR #1666, opened 2024-10-30
GDB-11134 fix error handling in similarity indexes view

svilenvelikov:
## What
Fix error handling in similarity indexes view.

## Why
Error handling in the similarity indexes view is broken and leads to unpleasant experience in the user. For example, when user tries to create or edit a namespace, but the server returns some error, the UI doesn't show a notification for the error, but just navigates to other view which leads to confusion.

## How
* Refactored the save and edit action handlers and added a flag that indicates whether the operation is in progress which is also reflected in the UI with a progress indicator in the respective action button.

### DIFF
--- a/src/js/angular/similarity/controllers/create-index.controller.js
+++ b/src/js/angular/similarity/controllers/create-index.controller.js
@@ -76,6 +76,11 @@ function CreateSimilarityIdxCtrl(
     $scope.saveOrUpdateExecuted = false;
     $scope.loadingControllerResources = false;
     $scope.helpHidden = LocalStorageAdapter.get(LSKeys.HIDE_SIMILARITY_HELP) === 1;
+    /**
+     * Flag that indicates if the index save operation is in progress.
+     * @type {boolean}
+     */
+    $scope.savingIndex = false;
 
     let isDirty = false;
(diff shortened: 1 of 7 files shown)


src/js/angular/similarity/controllers/create-index.controller.js
```diff
@@ -186,22 +196,30 @@ function CreateSimilarityIdxCtrl(
 
     $scope.saveSearchQuery = function () {
         $scope.saveOrUpdateExecuted = true;
-        if (!isDirty) {
```
boyan-tonchev, line 189: The purpose of this check is to avoid calling save if there are no changes in the index.

svilenvelikov:
Jenkins retry build

src/js/angular/similarity/controllers/create-index.controller.js
svilenvelikov: Changed the logic to really prevent the save operation if the query is pristine.

boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

boyan-tonchev (approved):

PR #1676, opened 2024-11-01
GDB-11134 fix error handling in my settings view

svilenvelikov:
## What
Fix the error handling in my settings view so that the UI state to be properly saved or retained when server error occurs during save or update operation.

## Why
This broken error handling harms the user experience in the view.

## How
* Moved switching the theme in the success handler after user's data have been updated to prevent the inconsistency in the UI state when a server error occurs and the user update operation fails.
* Extracted the ChangeUserPasswordSettingsCtrl as UserSettingsController in separate controller file. Extracted constants and utilities in separate files.
* Also reorganized and cleaned the code in the new UserSettingsController a bit for better readability and maintainability.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/src/js/angular/security/controllers/user-settings.controller.js
+++ b/src/js/angular/security/controllers/user-settings.controller.js
@@ -3,6 +3,7 @@ import {READ_REPO, WRITE_REPO} from "../services/constants";
 import {UserType} from 'angular/utils/user-utils';
 import {parseAuthorities} from "../services/authorities-util";
 import {UpdateUserPayload} from "../../models/security/security";
+import {CookiePolicyModalController} from "../../core/directives/cookie-policy/cookie-policy-modal-controller";
 
 angular
     .module('graphdb.framework.security.controllers.user-settings', [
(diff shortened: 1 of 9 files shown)


src/js/angular/security/controllers/user-settings.controller.js
```diff
@@ -0,0 +1,272 @@
...
+        const timer = $timeout(function () {
+            $window.history.back();
+        }, 100);
+        $scope.$on('$destroy', function () {
+            $timeout.cancel(timer);
+        });
+    };
+
+    // Wrapped in a scope function for ease of testing
+    $scope.getPrincipal = function () {
```
teodossidossev, line 103: you can simply combine this and next methods. Looks odd as is

src/js/angular/security/controllers/user-settings.controller.js
```diff
@@ -0,0 +1,272 @@
...
+        waitBeforeRedirectBack = $timeout(function () {
+            $scope.loader = false;
+            $window.history.back();
+        }, 2000);
+    };
+
+    const initUserData = function (scope) {
+        // Copy needed so that Cancel would work correctly, need to call updateCurrentUserData on OK
+        scope.userData = _.cloneDeep(scope.currentUserData);
+        scope.user = {username: scope.userData.username};
```
teodossidossev, line 231: we should introduce an UI user model

teodossidossev (changes requested):

src/js/angular/security/controllers/user-settings.controller.js
svilenvelikov: Sure. But not now.

teodossidossev (approved):

PR #1684, opened 2024-11-05
GDB-11179 issue with times in tooltip

boyan-tonchev:
## What
When the "Regenerate" or "How did you derive this answer?" buttons are clicked, the tooltip showing the time for questions and answers is incorrect.

## Why
Upon asking a question, the chat messages list in the UI is updated so the question becomes visible immediately after the user clicks the "Ask" button. Therefore, the question time needs to be set on the chatItem passed to the askQuestion function. This behavior already works for regular questions, but in the "Regenerate" and "How did you derive this answer?" scenarios, the time was not set before askQuestion was called.

## How
The question time is now set before calling the askQuestion function when the "Regenerate" or "How did you derive this answer?" buttons are clicked.

### Before
image
image

### After
image
image

## Checklist
- [ ] Branch name
- [ ] Target branch
- [ ] Commit messages
- [ ] MR name
- [ ] MR Description
- [ ] Tests

### DIFF
--- a/src/i18n/locale-en.json
+++ b/src/i18n/locale-en.json
@@ -436,7 +436,7 @@
                 "form": {
                     "agent_name": {
                         "label": "Agent name",
-                        "placeholder": "Enter user friendly name",
+                        "placeholder": "Enter а user friendly name",
                         "tooltip": "A descriptive name that helps you identify this agent."
                     },
                     "repository": {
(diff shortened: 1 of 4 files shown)


boyan-tonchev:
Jenkins retry build

svilenvelikov (approved):

boyan-tonchev:
Jenkins retry build

PR #1686, opened 2024-11-05
GDB-11171 - Show ellipsis on long chat names

DesiBorisova:
## What
Long chat names will be truncated with an ellipsis. The user can see the longer name on hover.

## Why
The names were not truncated, due to an incorrect selector.

## How
I added the correct selector and a directive, which works on hover to add the long name as a tooltip.

## Testing
N/A

## Screenshots
image

## Checklist
- [ ] Branch name
- [ ] Target branch
- [ ] Commit messages
- [ ] MR name
- [ ] MR Description
- [ ] Tests

### DIFF
--- a/src/css/ttyg/chat-list.css
+++ b/src/css/ttyg/chat-list.css
@@ -61,7 +61,7 @@
     width: 90%;
 }
 
-.chat-list-component .chat-item .chat-name .inline-editable-text {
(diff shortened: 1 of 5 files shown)


svilenvelikov:
There is too much left space on the right with this solution.

svilenvelikov (approved):

PR #1690, opened 2024-11-07
GDB-10692 - Change toggle behavior of import checkbox

DesiBorisova:
## What
The way the checkbox and drop-down in the Import page works will be as follows:

- When the user **selects** the checkbox next to the drop-down, all files will be **selected**.

- When the user **deselects** the checkbox next to the drop-down, all files will be **deselected**.

- When the user **clicks** the checkbox next to the drop-down while it is **indeterminate**, all files will be **deselected**.

- Selecting from the drop-down ("All", "None", "Imported", "Not imported") will not show the selected option next to the checkbox drop-down. Only the applicable files will be checked in the list.

- Manually selecting all the files will check the checkbox by the drop-down list.

- Manually deselecting all the files will uncheck the checkbox by the drop-down list.

## Why
This behavior was agreed with our UX specialist and matches the prototype in the task comment. The behavior is more intuitive for users, as it closely matches the Gmail email list checkbox drop-down.

## How
I added logic to switch the checkbox states, depending on the conditions. I edited the styles, so that the independent checkbox can appear as part of the button as before.

## Testing
N/A

## Screenshots
Screenshot from 2024-11-07 12-29-18

Screenshot from 2024-11-07 12-29-33

Screenshot from 2024-11-07 12-29-41

## Checklist
- [ ] Branch name
- [ ] Target branch
- [ ] Commit messages
- [ ] MR name
- [ ] MR Description
- [ ] Tests


src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -214,6 +232,25 @@ function importResourceTreeDirective($timeout, ImportContextService) {
...
+                }
+            };
+
+            const toggleCheckboxes = (checkboxElement, shouldSelect) => {
+                $scope.displayResources.forEach((resource) => resource.selected = shouldSelect);
+                $scope.selectedByStatus = shouldSelect ? STATUS_OPTIONS.ALL : STATUS_OPTIONS.NONE;
+                $scope.areAllDisplayedImportResourcesSelected = shouldSelect;
+                $scope.areAllDisplayedImportResourcesPartialSelected = false;
+
+                $timeout(() => {
```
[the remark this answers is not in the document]
DesiBorisova: Ah, the timeout isn't needed anymore. The checkbox state and variables are all updated before the state is changed. I'll remove it.

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -180,6 +180,24 @@ function importResourceTreeDirective($timeout, ImportContextService) {
                 $scope.onEditResource({resource});
             };
 
+            $scope.toggleSelectAll = () => {
+                const allSelected = $scope.displayResources.every((resource) => resource.selected);
+                const noneSelected = $scope.displayResources.every((resource) => !resource.selected);
+                const checkboxElement = document.getElementById('checkboxInput');
```
svilenvelikov, line 186: You shouldn't query the entire document. You have the directive's html element. Use it. But bare in mind that it's wrapped as a jquery object.

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -180,6 +180,24 @@ function importResourceTreeDirective($timeout, ImportContextService) {
                 $scope.onEditResource({resource});
             };
 
+            $scope.toggleSelectAll = () => {
+                const allSelected = $scope.displayResources.every((resource) => resource.selected);
```
svilenvelikov, line 184: I don't like these operations with the displayResources property. I know that you are just following the pattern, but it's time for you to start thinking about the code quality practices we are talking constantly. So, let's do this right, can we? 
First. Create another model which wraps the `ImportResourceTreeElement[]` list returned by the `resources.toList()`. Then implement methods for all such needs inside. Then update references where needed and reuse those methods, thus effectively abstracting the internal implementation of the tree elements list.

svilenvelikov:
Can you style the cursor when it's over checkboxexto be like a pointer?

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -214,6 +232,23 @@ function importResourceTreeDirective($timeout, ImportContextService) {
...
                 $scope.areAllDisplayedImportResourcesPartialSelected = hasSelectedDisplayedImportResource && hasUnselectedDisplayedImportResource;
+
+                if ($scope.areAllDisplayedImportResourcesSelected) {
+                    $scope.selectedByStatus = STATUS_OPTIONS.ALL;
+                } else if (!$scope.areAllDisplayedImportResourcesSelected && !$scope.areAllDisplayedImportResourcesPartialSelected) {
+                    $scope.selectedByStatus = STATUS_OPTIONS.NONE;
+                }
+            };
+
+            const toggleCheckboxes = (checkboxElement, shouldSelect) => {
```
svilenvelikov, line 243: `shouldSelect` sounds strange. Can you come up with some better name?

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -214,6 +232,23 @@ function importResourceTreeDirective($timeout, ImportContextService) {
...
+                $scope.displayResources.forEach((resource) => resource.selected = shouldSelect);
+                $scope.selectedByStatus = shouldSelect ? STATUS_OPTIONS.ALL : STATUS_OPTIONS.NONE;
+                $scope.areAllDisplayedImportResourcesSelected = shouldSelect;
+                $scope.areAllDisplayedImportResourcesPartialSelected = false;
+
+                if (checkboxElement) {
+                    checkboxElement.checked = shouldSelect;
```
svilenvelikov, line 250: I don't like this. Why are you working with the html element here when it's state is bound to some property in the directive?! See `areAllDisplayedImportResourcesSelected`!

src/js/angular/import/templates/import-resource-tree.html
```diff
@@ -2,13 +2,14 @@
 <div class="import-resource-tree mt-2" ng-if="!showLoader && resources && !resources.isEmpty()">
     <div class="import-resource-tree-header">
         <div class="import-resources-actions">
-            <div class="import-resource-status-dropdown btn-group" uib-dropdown>
+            <div class="import-resource-status-dropdown btn-group btn-secondary" uib-dropdown>
+                <input type="checkbox" id="checkboxInput" ng-class="{'selected': selectedByStatus}"
+                       prop-indeterminate="areAllDisplayedImportResourcesPartialSelected"
+                       ng-checked="selectedByStatus === STATUS_OPTIONS.ALL"
```
svilenvelikov, line 8: Why? See the other other note about the state of this input!

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -214,6 +232,23 @@ function importResourceTreeDirective($timeout, ImportContextService) {
                 const hasSelectedDisplayedImportResource = $scope.displayResources.some((resource) => resource.selected);
                 $scope.areAllDisplayedImportResourcesSelected = hasSelectedDisplayedImportResource && !hasUnselectedDisplayedImportResource;
                 $scope.areAllDisplayedImportResourcesPartialSelected = hasSelectedDisplayedImportResource && hasUnselectedDisplayedImportResource;
+
+                if ($scope.areAllDisplayedImportResourcesSelected) {
+                    $scope.selectedByStatus = STATUS_OPTIONS.ALL;
+                } else if (!$scope.areAllDisplayedImportResourcesSelected && !$scope.areAllDisplayedImportResourcesPartialSelected) {
+                    $scope.selectedByStatus = STATUS_OPTIONS.NONE;
+                }
```
svilenvelikov, line 240: I'm not sure if the changes in this method are actually needed

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -195,7 +222,8 @@ function importResourceTreeDirective($timeout, ImportContextService) {
             const updateListedImportResources = () => {
                 $scope.resources.getRoot().updateSelectionState();
                 sortResources();
-                $scope.displayResources = $scope.resources.toList()
+                $scope.checkboxControlModel = new CheckboxControlModel($scope.resources.toList());
+                $scope.displayResources = $scope.checkboxControlModel.getResourcesList()
```
svilenvelikov, line 226: The idea of this ne model/utility is to hide the internal implementation. In other words, there shouldn't have such filtering and other manipulations on the underlying array here. All this must happen in the wrapper. Otherwise it's pointless because the client would in fact know about the underlying structure and would be able to break it at some point.

src/js/angular/models/import/checkbox-control-resource-wrapper.js
```diff
@@ -0,0 +1,122 @@
+import {SortingType} from "./sorting-type";
+import {convertToBytes} from "../../utils/size-util";
+
+export class CheckboxControlModel {
```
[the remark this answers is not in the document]
svilenvelikov: It is a posibility, yes. Although it won't be quite convenient and performant, because the controller works with the array that is returned by the ImportResourceTreeElement#toList method and if you put these functions inside the ImportResourceTreeElement you'd need to call toList internally each time, whereas now it's possible to have a local cache of the `toList` result and just do the filtering, etc

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -210,83 +219,11 @@ function importResourceTreeDirective($timeout, ImportContextService) {
...
-
-                if ($scope.filterByType === TYPE_FILTER_OPTIONS.DIRECTORY) {
-                    return resource.hasTextInDirectoriesName($scope.filterByFileName);
-                }
+                const mainCheckbox = element[0].querySelector('#importSelectCheckboxInput');
 
-                if ($scope.filterByType === TYPE_FILTER_OPTIONS.FILE) {
-                    return resource.hasTextInFilesName($scope.filterByFileName);
+                if (mainCheckbox) {
+                    mainCheckbox.checked = !!($scope.resourceListUtil.areAllDisplayedImportResourcesSelected || $scope.resourceListUtil.areAllDisplayedImportResourcesPartialSelected);
```
boyan-tonchev, line 225: This condition `!!($scope.resourceListUtil.areAllDisplayedImportResourcesSelected || $scope.resourceListUtil.areAllDisplayedImportResourcesPartialSelected) ` could be moved into a function in `CheckboxControlModel`

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -314,6 +251,9 @@ function importResourceTreeDirective($timeout, ImportContextService) {
                 } else {
                     ImportResourceTreeService.mergeResourceTree($scope.resources, newResources, isUserImport);
                 }
+                $scope.resourceListUtil = $scope.resources.getResourceListUtil();
+                $scope.resourceListUtil.setNameFilter($scope.filterByFileName);
```
boyan-tonchev, line 255: The `nameFilter` needs to be updated when the filter changes. Update it in the `filterByFileNameChanged` function.

src/js/angular/import/directives/import-resource-tree.directive.js
```diff
@@ -314,6 +251,9 @@ function importResourceTreeDirective($timeout, ImportContextService) {
...
+                $scope.resourceListUtil.setTypeFilter($scope.filterByType);
```
boyan-tonchev, line 256: The `typeFilter` needs to be updated when the filter changes. Update it in the `filterByTypeChanged` function.

PR #1693, opened 2024-11-08
GDB-11022: User Guides download ttl files with local security ON lead to alternate browser auth.

boyan-tonchev:
When
The issue occurs when users attempt to download a guide resource in a secured workbench. The browser’s default behavior opens a basic authentication dialog box.

Why
Originally, the download link was implemented as a simple HTML anchor tag (<a>) that opens a new tab when clicked. Since the backend requires authorization, the request fails with a 401 Unauthorized status, which triggers the browser's default authentication dialog.

How
To address this, the default link behavior is prevented by using ng-click in place of a standard link action. Now, when the link is clicked, a global function is called instead. This method sends the request with the necessary authorization header, allowing secure access to the resource without triggering the browser’s default login dialog.

## Testing
There are existing tests.

## Screenshots
image

image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] MR name
- [x] MR Description
- [ ] Tests


src/js/angular/guides/guides.service.js
```diff
@@ -162,6 +189,12 @@ function GuidesService($http, $rootScope, $translate, $interpolate, ShepherdServ
         this._subscribeToGuideResumed();
         this._subscribeToGuideCancel();
         this._subscribeToGuidePause();
+        $rootScope.downloadGuidesFile = (resourcePath, resourceFile) => {
```
svilenvelikov, line 192: No! This won't make it! Better move this function in the main controller.

boyan-tonchev: Moved

src/js/angular/guides/guides.service.js
```diff
@@ -164,6 +191,13 @@ function GuidesService($http, $rootScope, $translate, $interpolate, ShepherdServ
         this._subscribeToGuidePause();
     };
 
+    this.downloadGuidesFile = (resourcePath, resourceFile) => {
+        GuidesRestService.downloadGuideResource(`${resourcePath}/${resourceFile}`)
```
svilenvelikov, line 195: where is the return

src/js/angular/controllers.js
```diff
@@ -1023,6 +1024,10 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repos
             onRepositoriesChanged();
         }
     });
+
+    $scope.downloadGuidesFile = (resourcePath, resourceFile) => {
+        GuidesService.downloadGuidesFile(resourcePath, resourceFile);
```
svilenvelikov, line 1029: catch?

src/js/angular/guides/guides.service.js
boyan-tonchev: Done

src/js/angular/controllers.js
boyan-tonchev: done

src/js/angular/controllers.js
```diff
@@ -28,6 +28,7 @@ import 'angularjs-slider/dist/rzslider.min';
 import {debounce} from "lodash";
 import {DocumentationUrlResolver} from "./utils/documentation-url-resolver";
 import {NamespacesListModel} from "./models/namespaces/namespaces-list";
+import {FileUtils} from "./utils/file-utils";
```
svilenvelikov, line 31: remove it

boyan-tonchev: Done

svilenvelikov (approved):

PR #1707, opened 2024-11-18
GDB-10443 Enrich the Translation Service

boyan-tonchev:
## What
Enhance the Translation service to support language bundles in hierarchical structure, allowing for hierarchical keys and nested objects within the translation files.

## Why
This change simplifies the management and maintenance of labels, especially for complex UIs with multiple nested elements. With this approach, translations can be organized into logical groups and sub-groups, making it easier to locate, update, and expand the labels as needed.

A sample translation file might look like this:

```
{
"visual.graph.label": "Visual graph",
"view.rdf.rank.title": "RDF Rank",
"language_selector": {
   "toggle_menu": {
      "label": "Select language",
      "tooltip": ""
   }
 }
}
```

## How
Enhance the translation service to attempt extracting a translation from a hierarchical object if a simple key lookup does not return a value.

## Checklist
- [X] Branch name
- [X] Target branch
- [x] Commit messages
- [X] MR name
- [X] MR Description

### DIFF
--- a/packages/shared-components/src/services/translation.service.ts
+++ b/packages/shared-components/src/services/translation.service.ts
@@ -4,13 +4,18 @@ import {ServiceProvider, LanguageService} from "@ontotext/workbench-api";
 import {Subscription} from '@reactivex/rxjs/dist/package';
 import {TranslationParameter} from '../models/translation/translation-parameter';
 import {TranslationCallback, TranslationObserver} from '../models/translation/translation-observer';
+import {sanitizeHTML} from '../utils/html-utils';
+
+type TranslationBundle = {
+  [key: string]: string | TranslationBundle;
+};
 
 /**
  * Service responsible for translation operations in the component.
  */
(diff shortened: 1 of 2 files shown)


packages/shared-components/src/services/translation.service.ts
```diff
@@ -4,13 +4,18 @@ import {ServiceProvider, LanguageService} from "@ontotext/workbench-api";
 import {Subscription} from '@reactivex/rxjs/dist/package';
 import {TranslationParameter} from '../models/translation/translation-parameter';
 import {TranslationCallback, TranslationObserver} from '../models/translation/translation-observer';
+import {sanitizeHTML} from '../utils/html-utils';
+
+type TranslationBundle = {
```
svilenvelikov, line 9: Convert to `Record<string, TranslationBundle>`

svilenvelikov (approved):

PR #1709, opened 2024-11-19
GDB-10570: Introduce dropdown component

boyan-tonchev:
## What
 Introduce a reusable dropdown component onto-dropdown, designed to provide a flexible, customizable dropdown menu that integrates seamlessly with other components. The dropdown supports features such as configurable labels, tooltips, icons, and item lists.

## Why
 To enhance consistency and reduce repetitive code, a centralized and reusable dropdown component. Simplifies implementation of dropdowns across the application. Ensures uniform styling and behavior.

## How
 The onto-dropdown component was implemented using StencilJS. Key features include:

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] MR name
- [X] MR Description

### DIFF
--- a/packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
+++ b/packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
@@ -1,0 +1,128 @@
+import {Component, Event, h, State, Prop, EventEmitter, Listen} from '@stencil/core';
+import {DropdownItem} from '../../models/dropdown/dropdown-item';
+import {TranslationService} from '../../services/translation.service';
+import {DropdownItemAlignment} from '../../models/dropdown/dropdown-item-alignment';
+
+/**
+ * A reusable dropdown component built using StencilJS. This component supports configurable labels, tooltips, icons,
+ * and items, making it versatile for various use cases. It also integrates with a translation service to handle
+ * internationalization.
+ */
+@Component({
+    tag: 'onto-dropdown',
+    styleUrl: 'onto-dropdown.scss',
+    shadow: false,
+})
+export class OntoDropdown {
+
+    /**
+     * Indicates whether the dropdown menu is open.
+     */
+    @State() open = false;
+
+    /**
+     * The dropdown button name. It will be used if present; otherwise, the {@link OntoDropdown#dropdownButtonNameLabelKey} will be used.
+     */
+    @Prop() dropdownButtonName: string;
+
+    /**
+     * The translation label key for the dropdown button name. It will be used if {@link OntoDropdown#dropdownButtonName} is not present.
+     */
+    @Prop() dropdownButtonNameLabelKey: string;
+
+    /**
+     * The dropdown button tooltip. It will be used if present; otherwise, the {@link OntoDropdown#dropdownButtonTooltipLabelKey} will be used.
+     */
+    @Prop() dropdownButtonTooltip: string;
+
+    /**
+     * The translation label key for the dropdown button tooltip. It will be used if {@link OntoDropdown#dropdownButtonTooltip} is not present.
+     */
+    @Prop() dropdownButtonTooltipLabelKey: string;
+
+    /**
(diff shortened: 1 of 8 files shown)


packages/shared-components/src/models/dropdown-option.ts
```diff
@@ -0,0 +1,63 @@
+export class DropdownOption {
+  private _label: string;
+  private _labelKey: string;
+  private _tooltip: string = '';
+  private _tooltipLabelKey: string;
+  private _iconClass: string;
+  private _value: any;
```
svilenvelikov, line 7: Why `any`? Isn't this just a plain string?

packages/shared-components/src/components/core/onto-dropdown/onto-dropdown.tsx
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Array of dropdown options. Each option includes a value, label, tooltip, and optional icon class.
+     */
+    @Prop() items: DropdownOption[];
+
+    /**
+     * Event emitted when a dropdown item is selected.
+     * The event payload contains the value of the selected item.
+     */
+    @Event() valueChanged: EventEmitter<string>;
```
svilenvelikov, line 58: And here the payload is string, although the value option is set to be `any`. You need to clear this up

packages/shared-components/src/components/core/onto-dropdown/onto-dropdown.tsx
```diff
@@ -0,0 +1,123 @@
...
+                        class='button-name'>{this.nameLabel ? this.nameLabel : TranslationService.translate(this.nameLabelKey)}</span>
+                </button>
+
+                <div class="dropdown-menu pre-scrollable dropdown-menu-right">
+                    {this.items && this.items.map(item =>
+                        <a href="#" class='dropdown-item' title={TranslationService.translate(item.tooltip)}
+                           onClick={() => this.onSelect(item.value)}>
+                            <span class={'dropdown-option-icon ' + item.iconClass}></span>
+                            {item.label ?
+                                <span innerHTML={item.label}></span> : TranslationService.translate(item.labelKey)}
```
svilenvelikov, line 104: I'd prefer to have consistent html structure in both cases here. Add a span tag around the translated label as well.

svilenvelikov: Also, it'd be possible in practice to have html tags in the translated labels as well. So, in both cases the label should be injected via the innerHTML

svilenvelikov:
Why the branch name is called "implement-dropdown-component"

svilenvelikov:
Remove the `core` package!

packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Array of dropdown options.
+     */
+    @Prop() items: DropdownItem[];
+
+    /**
+     * Specifies the dropdown items' alignment. If not provided, the items and the dropdown button will be aligned to the left.
+     *
+     */
+    @Prop() dropdownAlignment: DropdownItemAlignment = DropdownItemAlignment.LEFT;
```
svilenvelikov, line 57: I think it'd be better if we use plain string options here instead of enum values because we don't export the module types out of this module. Or, consider exporting the models as well, although I'm not sure if really want to do this because some of these are for internal use only.

packages/shared-components/src/models/dropdown-option.ts
boyan-tonchev: Change it to string

packages/shared-components/src/components/core/onto-dropdown/onto-dropdown.tsx
boyan-tonchev: fixed

svilenvelikov (approved):

PR #1714, opened 2024-11-22
GDB-10693 - Disable local repo restart in cluster

DesiBorisova:
## What
When the node is in a cluster, restarting the Local repositories from the Repositories view and the Edit Repository view will not be allowed.

## Why
The action is forbidden in the back-end and performing a restart on a Local repo, while in a cluster, causes problems for the GraphDB. 

## How
I added checks to see if the node is in a cluster and if the edited repo is a remote location or not.

## Testing
Tests added.

## Screenshots
N/A

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/src/js/angular/repositories/controllers.js
+++ b/src/js/angular/repositories/controllers.js
@@ -7,6 +7,7 @@ import {
 import {decodeHTML} from "../../../app";
 import './controllers/manage-remote-location-dialog.controller';
 import {RemoteLocationType} from "../models/repository/remote-location.model";
+import 'angular/rest/cluster.rest.service';
 
 export const getFileName = function (path) {
     let lastIdx = path.lastIndexOf('/');
@@ -130,7 +131,8 @@ const modules = [
     'graphdb.framework.utils.localstorageadapter',
     'toastr',
     'ngFileUpload',
-    'graphdb.framework.repositories.controllers.manage-remote-location-dialog'
+    'graphdb.framework.repositories.controllers.manage-remote-location-dialog',
+    'graphdb.framework.rest.cluster.service'
 ];
 
 angular.module('graphdb.framework.repositories.controllers', modules)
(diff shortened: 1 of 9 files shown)


teodossidossev (changes requested):

src/js/angular/repositories/controllers.js
```diff
@@ -405,6 +421,7 @@ function LocationsAndRepositoriesCtrl($scope, $rootScope, $uibModal, toastr, $re
         // Update repositories state
         $repositories.initQuick();
         getLocations();
+        getAndSetClusterStatus();
```
svilenvelikov, line 424: Do we need this to be here? Isn't the initial check enough?

test-cypress/integration/repository/repositories.spec.js
```diff
@@ -503,6 +505,44 @@ describe('Repositories', () => {
...
+        GlobalOperationsStatusesStub.stubGlobalOperationsStatusesResponse(repositoryId);
+        ClusterStubs.stubClusterNodeStatus();
+        // Then go to the repositories page
+        RepositorySteps.visit();
+        // I expect the repository restart button to be disabled
+        RepositorySteps.getRepositoryRestartButton(repositoryId).should('be.visible');
+        RepositorySteps.getRepositoryRestartButton(repositoryId).should('be.disabled');
+    });
+
+    it('should allow restart of remote repositories, even if node is in cluster', () => {
```
svilenvelikov, line 534: Describe this test better, please, because for a person who is unfamiliar with this behavior it won't be clear why this test verifies the permission to restart in cluster and the previous two tests are doing the oppisite.

src/js/angular/repositories/controllers.js
```diff
@@ -171,6 +174,18 @@ function LocationsAndRepositoriesCtrl($scope, $rootScope, $uibModal, toastr, $re
...
     }
 
+    function getAndSetClusterStatus() {
+        ClusterRestService.getNodeStatus().then((response) => {
+            $scope.isInCluster = response.data.nodeState;
+        }).catch((error) => {
+            if (error.status === 404) {
+                $scope.isInCluster = false;
+            } else {
+                console.error('An unexpected error occurred:', error);
```
svilenvelikov, line 184: I guess this is dumped in the log because otherwise it'd spam the user with toastr popups when not cluster. Right? But I'm not convinced that we should constantly check for cluster status here in the first place. We need to confirm that with @avataar

src/js/angular/repositories/controllers.js
```diff
@@ -171,6 +174,18 @@ function LocationsAndRepositoriesCtrl($scope, $rootScope, $uibModal, toastr, $re
             .finally(() => $scope.loader = false);
     }
 
+    function getAndSetClusterStatus() {
+        ClusterRestService.getNodeStatus().then((response) => {
+            $scope.isInCluster = response.data.nodeState;
```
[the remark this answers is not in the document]
svilenvelikov: It's not. It's a string, so it needs a proper conversion to a boolean value.

svilenvelikov (approved):

svilenvelikov:
Jenkins retry build

svilenvelikov:
Jenkins retry build

PR #1721, opened 2024-11-25
GDB-10570 fixes the imports from the api module in shared components

svilenvelikov:
## What
Fixes the imports from the api module in shared components.

## Why
Import from the api module in format like `import {ServiceProvider, LanguageService} from "@ontotext/workbench-api"` weren't resolved properly due to a combination of various reasons listed below:
- The api module was built with an older version of the `"webpack-config-single-spa-ts": "^4.0.0"` which used to export the module in a systemjs format instead of ES6 module format. The latest version of the plugin exports in ES6 by default so we are now using it because it's what we expect.
- In order for the api module to be accessible from the other modules it must be built and started!
- In stenciljs pages we need to add importmap so that the browser to know how to resolve the module name.
- (Not sure if this is truly needed when we have the importmap) And finally we need the nodeResolve plugin in stenciljs so that the module can be properly loaded.

## Checklist
- [ ] Branch name
- [ ] Target branch
- [ ] Commit messages
- [ ] Squash commits
- [ ] MR name
- [ ] MR Description
- [ ] Tests


svilenvelikov:
This was fixed with https://github.com/Ontotext-AD/graphdb-workbench/pull/1751

PR #1722, opened 2024-11-26
fix(GDB-11271) Restrict delete tag button visibility to admin users in cluster configuration

teodossidossev:
## WHAT:
Add `ng-if="isAdmin"` condition to the delete tag button in the multi-region cluster configuration template to restrict its visibility to admin users.

## WHY:
The delete button was visible to basic users, who are not allowed to delete tags. Clicking the button resulted in a "forbidden" toast and banner, which blocked the cluster view until a page refresh.

## HOW:
Added a `ng-if="isAdmin"` directive to the delete tag button in the HTML template. This ensures the button is only rendered for admin users, preventing basic users from interacting with it.

## Screenshots
Admin user
image

Regular user
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/src/js/angular/clustermanagement/templates/cluster-configuration/multi-region.html
+++ b/src/js/angular/clustermanagement/templates/cluster-configuration/multi-region.html
@@ -81,7 +81,8 @@
                         </td>
                         <td class="actions-cell">
                             <div class="actions">
-                                <button ng-click="deleteTag(tag[0])"
+                                <button ng-if="isAdmin"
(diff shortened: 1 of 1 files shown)


svilenvelikov (approved):

svilenvelikov:
Jenkins retry build

PR #1723, opened 2024-11-26
GDB-11267: Implementation of Tooltip Component

boyan-tonchev:
## What
A new tooltip component has been implemented.

## Why
The tooltip component is designed for use within the shared-components module, ensuring consistency and reusability.

## How
Implemented a tooltip component that serves as a wrapper for the tippy.js library to provide a feature-rich and customizable tooltip solution.

## Screenshots
image

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [X] Tests

### DIFF
--- a/packages/shared-components/src/components.d.ts
+++ b/packages/shared-components/src/components.d.ts
@@ -73,6 +73,8 @@ export namespace Components {
          */
         "selectedMenu": string;
     }
+    interface OntoTooltip {
+    }
     /**
      * The purpose of this component is to display translated literals in the DOM. A Stencil component re-renders when a prop or state changes,
      * but it may not re-render when the language changes. In such cases, this component should be used. It handles language change events
@@ -160,6 +162,12 @@ declare global {
         prototype: HTMLOntoNavbarElement;
         new (): HTMLOntoNavbarElement;
     };
+    interface HTMLOntoTooltipElement extends Components.OntoTooltip, HTMLStencilElement {
+    }
+    var HTMLOntoTooltipElement: {
+        prototype: HTMLOntoTooltipElement;
+        new (): HTMLOntoTooltipElement;
+    };
     /**
      * The purpose of this component is to display translated literals in the DOM. A Stencil component re-renders when a prop or state changes,
      * but it may not re-render when the language changes. In such cases, this component should be used. It handles language change events
@@ -183,6 +191,7 @@ declare global {
         "onto-language-selector": HTMLOntoLanguageSelectorElement;
         "onto-layout": HTMLOntoLayoutElement;
         "onto-navbar": HTMLOntoNavbarElement;
+        "onto-tooltip": HTMLOntoTooltipElement;
         "translate-label": HTMLTranslateLabelElement;
     }
 }
@@ -252,6 +261,8 @@ declare namespace LocalJSX {
          */
         "selectedMenu"?: string;
     }
+    interface OntoTooltip {
+    }
     /**
      * The purpose of this component is to display translated literals in the DOM. A Stencil component re-renders when a prop or state changes,
      * but it may not re-render when the language changes. In such cases, this component should be used. It handles language change events
@@ -273,6 +284,7 @@ declare namespace LocalJSX {
         "onto-language-selector": OntoLanguageSelector;
         "onto-layout": OntoLayout;
         "onto-navbar": OntoNavbar;
+        "onto-tooltip": OntoTooltip;
         "translate-label": TranslateLabel;
     }
 }
@@ -291,6 +303,7 @@ declare module "@stencil/core" {
             "onto-language-selector": LocalJSX.OntoLanguageSelector & JSXBase.HTMLAttributes<HTMLOntoLanguageSelectorElement>;
             "onto-layout": LocalJSX.OntoLayout & JSXBase.HTMLAttributes<HTMLOntoLayoutElement>;
             "onto-navbar": LocalJSX.OntoNavbar & JSXBase.HTMLAttributes<HTMLOntoNavbarElement>;
+            "onto-tooltip": LocalJSX.OntoTooltip & JSXBase.HTMLAttributes<HTMLOntoTooltipElement>;
             /**
              * The purpose of this component is to display translated literals in the DOM. A Stencil component re-renders when a prop or state changes,
              * but it may not re-render when the language changes. In such cases, this component should be used. It handles language change events
--- a/packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
+++ b/packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
@@ -91,9 +91,10 @@ export class OntoDropdown {
 
     render() {
         return (
-            <div class={`onto-dropdown ${this.open ? 'open' : 'closed'}`}
-                 title={this.dropdownButtonTooltip ? this.dropdownButtonTooltip : TranslationService.translate(this.dropdownButtonTooltipLabelKey)}>
+            <div class={`onto-dropdown ${this.open ? 'open' : 'closed'}`}>
                 <button class="onto-dropdown-button"
+                        tooltip-content={this.dropdownButtonTooltip ? this.dropdownButtonTooltip : TranslationService.translate(this.dropdownButtonTooltipLabelKey)}
+                        tooltip-placement='left'
                         onClick={() => this.toggleComponent()}>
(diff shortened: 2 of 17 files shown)


packages/shared-components/src/components.d.ts
```diff
@@ -9,11 +9,13 @@ import { DropdownItem } from "./models/dropdown/dropdown-item";
 import { DropdownItemAlignment } from "./models/dropdown/dropdown-item-alignment";
 import { ExternalMenuModel } from "./components/onto-navbar/external-menu-model";
 import { NavbarToggledEvent } from "./components/onto-navbar/navbar-toggled-event";
+import { Placement } from "tippy.js";
 import { TranslationParameter } from "./models/translation/translation-parameter";
 export { DropdownItem } from "./models/dropdown/dropdown-item";
 export { DropdownItemAlignment } from "./models/dropdown/dropdown-item-alignment";
 export { ExternalMenuModel } from "./components/onto-navbar/external-menu-model";
 export { NavbarToggledEvent } from "./components/onto-navbar/navbar-toggled-event";
+export { Placement } from "tippy.js";
```
svilenvelikov, line 18: duplication

svilenvelikov: This is generated code and I wonder why this might be happening

boyan-tonchev: This is very strange. I deleted the duplication, but it reappeared after rebuilding.

packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
```diff
@@ -91,35 +91,43 @@ export class OntoDropdown {
...
     render() {
         return (
-            <div class={`onto-dropdown ${this.open ? 'open' : 'closed'}`}
-                 title={this.dropdownButtonTooltip ? this.dropdownButtonTooltip : TranslationService.translate(this.dropdownButtonTooltipLabelKey)}>
-                <button class="onto-dropdown-button"
-                        onClick={() => this.toggleComponent()}>
-                    <span class={'button-icon ' + this.iconClass}></span>
-                    <span class='button-name'>
+          <div class={`onto-dropdown ${this.open ? 'open' : 'closed'}`}>
+            <onto-tooltip
```
svilenvelikov, line 95: I know that we implemented it this way for the YASGUI, but having the tooltip as a standalone component by itself requires to wrap it around the element on which we need to show the tooltip.  And I need to say that I'm not quite happy with this approach, because it looks quite verbose. I think that we should reconsider it more like an attribute directive in angular instead.

packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
```diff
@@ -0,0 +1,103 @@
...
+  private update() {
+    this.init();
+    this.tooltip.setContent(this.tooltipContent);
+  }
+
+  private init() {
+    if (this.tooltip) {
+      return;
+    }
+    const options = {
```
svilenvelikov, line 56: Create a type for this!
Move the options initialization outside!

packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
```diff
@@ -0,0 +1,103 @@
...
+      }
+    }
+  }
+
+  private createHideFunction(tooltip: Instance) {
+    return () => tooltip.hide();
+  }
+
+  private destroy() {
+    this.el.removeEventListener('mouseover', this.showFunction);
```
svilenvelikov, line 99: Are you sure these really unsubscribe the event handlers? Given how you create them I'm not quite convinced it really works.

packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
boyan-tonchev: I didn’t find a way to do it.

packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
boyan-tonchev: Done

packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
boyan-tonchev: The functions are created at runtime, and the registration and unregistration of the listener use the same function reference. I added some console logging, and it looks like it is working as expected.

packages/shared-components/src/components/onto-tooltip/models/onto-trigger-type.ts
svilenvelikov, line 1: Either rename this to OntoTooltipTriggerEvent or remove it

packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
```diff
@@ -0,0 +1,102 @@
...
+    this.tooltip.setContent(this.tooltipContent);
+  }
+
+  private init() {
+    if (this.tooltip) {
+      return;
+    }
+    this.tooltip = tippy(this.el, this.getTooltipOptions());
+
+    this.showFunction = this.createShowFunction(this.tooltip);
```
svilenvelikov, line 59: And now we reach to the point where we need to have these named consistently as we do it everywhere else. For example, we never name event handlers like `someFunction`... Rename these to `onMouseOver` and `onMouseLeave`

packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
svilenvelikov: image

svilenvelikov: I'll push a branch so that we can discuss on it

svilenvelikov: I have pushed an example component here.

packages/shared-components/src/components/single-tooltip/single-tooltip.tsx
```diff
@@ -0,0 +1,62 @@
+import { Component, State, h } from '@stencil/core';
+
+@Component({
+  tag: 'single-tooltip',
+  styleUrl: 'single-tooltip.scss',
+  shadow: true,
+})
+export class Tooltip {
```
svilenvelikov, line 8: Of course we need to wrap the tippy tooltip inside here.

svilenvelikov: This might help in the re-positioning https://atomiks.github.io/tippyjs/v6/all-props/#getreferenceclientrect

packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
boyan-tonchev: Done

packages/shared-components/src/components/onto-tooltip/models/onto-trigger-type.ts
```diff
@@ -0,0 +1,9 @@
+export enum OntoTriggerType {
```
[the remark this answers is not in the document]
boyan-tonchev: Removed

packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
```diff
@@ -0,0 +1,102 @@
...
+          .setTriggerTarget(this.el)
+          /**
+           * The tippy library has some conflict with the Google Chart Editor. The editor adds a div element with the "jfk-tooltip" class.
+           * When the mouse hovers over a chart editor element, this div is positioned accordingly.
+           * When the mouse leaves the element, a "jfk-tooltip-hidden" class is added, and the div tag is hidden.
+           * For some reason, when a 'onto-tooltip' is open, the "jfk-tooltip-hidden" class is removed, and the Google Chart tooltip is displayed along with
+           * 'onto-tooltip' (maybe Google uses an old version of tippy or popover and there is a conflict in implementations).
+           * When 'onto-tooltip' is open, we add the class 'hidden' to the Google Chart tooltip to hide it.
+           */
+          .setOnShow(() => document.querySelectorAll('.jfk-tooltip').forEach(popper => popper.classList.add('hidden')))
```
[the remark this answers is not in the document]
boyan-tonchev: Done

PR #1728, opened 2024-11-28
GDB-11273 - Allow wildcard usage in ACL role names

DesiBorisova:
## What
When creating a new role in either of the `ACL management tabs`, the user will be able to type just `*` as a single wildcard role and not receive an error that the role name is too short.

## Why
The documentation notes that the wildcard (`*`) is a valid ACL role, but the UI did not allow it, because it was less than the previously agreed length of 2 symbols.

## How

- I extracted the `ngTagsInput` logic from the `custom-role-prefix` directive and added it to its own directive where it is used (in the security module user page).

- I created a new directive to handle the roles entered in the ACL Management view. The old logic (for adding the `CUSTOM_` or `!CUSTOM_` prefix to the model values and removing the prefixes for the view value) remains.

- I added a validator, which checks the length of the user text (without the added prefixes). It also checks if only a wildcard is used, in which case the value is returned as is (no prefix added).

- I updated the language files with a more descriptive message for the user.

## Testing
Test added and a few existing ones altered or extended, in order to cover more cases.

## Screenshots
**The Statements tab with different roles added:**
Screenshot from 2024-11-28 15-43-29

**Altered warning message:**
Screenshot from 2024-11-28 15-31-43

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


src/js/angular/security/directives/custom-prefix-tagsInput.directive.js
svilenvelikov, line 1: Can rename the file to `custom-prefix-tags-input.directive.js` please

svilenvelikov (approved):

PR #1735, opened 2024-12-03
GDB-11294 - Load the default SPARQL query during Agent Edit

DesiBorisova:
## What
When the user edits an `Agent`  that has `Fetch ontology from a named graph` selected, selecting `Provide a SPARQL CONSTRUCT query that fetches the ontology` will display the default SPARQL query in the text field.

## Why
Previously, switching from `Fetch ontology from a named graph` to `Provide a SPARQL CONSTRUCT query that fetches the ontology` loaded an empty text field, because for the named graph option there was no `sparqlQuery` for that extraction method when the `Agent` is loaded initially.

## How
I edited the `extraction methods form mapper` to use the `default Agent` values for the `sparqlQuery` and `ontologyGraph` fields when the current model data has an `undefined` property.

## Testing
N/A

## Screenshots
N/A

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


DesiBorisova:
Jenkins retry build

DesiBorisova:
Jenkins retry build

svilenvelikov (approved):

PR #1737, opened 2024-12-04
GDB-11251 - Close repo info popover on click outside, so content can be copied

DesiBorisova:
## What
The user will be able to copy the information in the repository popover, because it will not disappear on `mouseleave`. Instead it will disappear on click outside.

## Why
The functionality is an improvement suggested by users. There was no way to copy this information before.

## How
I moved the popover to another element, in order to allow it to close on click outside.

## Testing
N/A

## Screenshots
N/A

## Checklist
- [ ] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/src/js/angular/controllers.js
+++ b/src/js/angular/controllers.js
@@ -600,6 +600,7 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repos
     window.addEventListener('storage', localStoreChangeHandler);
 
     $scope.$on('$destroy', () => {
+        document.removeEventListener('click', closeActiveRepoPopoverEventHandler);
         window.removeEventListener('storage', localStoreChangeHandler);
         deregisterMenuWatcher();
         if ($scope.checkMenu) {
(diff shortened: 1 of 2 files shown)


src/js/angular/controllers.js
```diff
@@ -652,6 +659,29 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repos
         }
     };
 
+    $scope.isPopoverOpen = false;
```
boyan-tonchev, line 662: It would be better to rename this property and all related functions to more specific names for better readability. The name isPopoverOpen could be associated with multiple popover components on the page. A name like isActiveRepositoryPopover,Open or something similar, would improve code readability.

DesiBorisova: I renamed the functions and this field for better readability

boyan-tonchev (approved):

DesiBorisova:
Jenkins retry build

DesiBorisova:
Jenkins retry build

PR #1747, opened 2024-12-09
GDB-10569: WB Migration: create MapperProvider

boyan-tonchev:
## What
Introduce MapperProvider.

## Why
To easily retrieve a single instance of a given mapper.

## How
Implement a MapperProvider that holds and returns the created instance of the specified mapper, with instances being created lazily upon the first request.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/api/src/providers/mapper/mapper.provider.ts
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * Retrieves a Mapper instance for the given type. If no instance exists, a new one is created,
+   * stored, and then returned.
+   *
+   * @template T - The specific Mapper type to retrieve or create.
+   * @param type - The constructor of the Mapper to retrieve or create.
+   * @returns The Mapper instance for the given type.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public static get<T extends Mapper<any>>(type: new () => T): T {
```
svilenvelikov, line 26: Here it's better to be avoided usage of `any`. You can use `unknown` instead, but the best would be if you create a custom type for the mapper constructor
`type MapperConstructor<T> = new () => Mapper<T>`
and use it here
`public static get<T>(type: MapperConstructor<T>): Mapper<T> {}`

packages/api/src/services/repository-location/repository-location.service.ts
```diff
@@ -1,18 +1,20 @@
...
  * The RepositoryLocationService class is responsible for fetching repository-location-related data from the backend
  * and mapping the responses to application models.
  */
 export class RepositoryLocationService implements Service {
   private locationRestService: RepositoryLocationRestService;
+  private repositoryLocationMapper: RepositoryLocationMapper;
 
   constructor() {
     this.locationRestService = ServiceProvider.get(RepositoryLocationRestService);
+    this.repositoryLocationMapper = MapperProvider.get(RepositoryLocationMapper);
```
svilenvelikov, line 17: I think we don't need to cache the mappers, but you can leave it this way for now.

packages/api/src/services/repository/mappers/repository-list.mapper.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+      ]
+    };
+
+    const repositoryList = repositoryListMapper.mapToModel(rawRepositoryObjects);
+
+    expect(repositoryList).toBeInstanceOf(RepositoryList);
+    const repositories = repositoryList.getItems();
+    expect(repositories).toHaveLength(3);
+    expect(repositories).toContainEqual(new Repository({id: repoOneId} as Repository));
+    expect(repositories).toContainEqual(new Repository({id: repoTwoId} as Repository));
```
svilenvelikov, line 37: Eventually, check if the two instances above doesn't have a location

packages/api/src/services/repository/mappers/repository.mapper.ts
```diff
@@ -0,0 +1,22 @@
...
+   *
+   * This method checks if the provided data can be considered valid data for a `Repository`.
+   * If valid, it constructs a new `Repository` object from the data. If not, it returns
+   * an empty `Repository` object.
+   *
+   * @param {unknown} data - The raw data to be transformed into a Repository model.
+   * @returns {Repository} - A new Repository instance.
+   */
+  mapToModel(data: unknown): Repository {
+    return Mapper.isTypeOf<Repository>(data) ? new Repository(data) : new Repository();
```
svilenvelikov, line 20: This is a bit confusing. If `data` is not in the expected Repository type format you discard it and return a blank instance. Why is that?!

packages/api/src/providers/mapper/mapper.provider.ts
boyan-tonchev: Done

packages/api/src/providers/mapper/mapper.provider.ts
```diff
@@ -0,0 +1,38 @@
...
+ * The `MapperProvider` is designed to handle the creation and reuse of `Mapper` instances
+ * for different types. Each type of `Mapper` is stored in a map, ensuring that only one instance
+ * per type exists during the application's lifecycle.
+ */
+export class MapperProvider {
+  /**
+   * A static map that stores singleton instances of `Mapper` objects, keyed by their type name.
+   */
+  // eslint-disable-next-line @typescript-eslint/consistent-generic-constructors, @typescript-eslint/no-explicit-any
+  private static readonly MAPPER_INSTANCES: Map<string, Mapper<any>> = new Map();
```
svilenvelikov, line 15: OK, replace this any with unknown as well

packages/api/src/services/repository/mappers/repository-list.mapper.spec.ts
boyan-tonchev: Updated raw data and added a check for locations.

packages/api/src/services/repository/mappers/repository.mapper.ts
boyan-tonchev: Refactored.

packages/api/src/providers/mapper/mapper.provider.ts
boyan-tonchev: Done

packages/api/src/services/repository/mappers/repository-list.mapper.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+        }
+      ]
+    };
+
+    const repositoryList = repositoryListMapper.mapToModel(rawRepositoryObjects);
+
+    expect(repositoryList).toBeInstanceOf(RepositoryList);
+    const repositories = repositoryList.getItems();
+    expect(repositories).toHaveLength(3);
+    expect(repositories).toContainEqual(new Repository({id: repoOneId, location: ''} as Repository));
```
svilenvelikov, line 38: I didn't meant that, but rather mock with an actual response as you did above (without the location property for the local repositories) and verify the the Repository instance is being created without the location property.

PR #1749, opened 2024-12-11
GDB-11369: WB Migration: Implement the Repository Selector Tooltip

boyan-tonchev:
## What
- Fixed an issue where tooltips for dropdown items were not working;
- Implemented tooltips for the repository selector.

## Why
- We replaced item.tooltip ? item.tooltip : this.translate(item.tooltipLabelKey) with item.tooltip ?? this.translate(item.tooltipLabelKey). However, there is a subtle difference: if the tooltip is an empty string, the first approach uses the translation, while the second retains the empty string;
- This change was made to improve usability and provide more contextual information.

## How
- Updated the default value of the tooltip to undefined instead of an empty string;
- Extended the dropdown configuration to support a promise that returns the tooltip string. A function was passed to the dropdown configuration to fetch additional repository information from the backend and convert it into an HTML string to be used as the tooltip content.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [ ] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


svilenvelikov:
Some issues with the menus:
* Both menus are 4px higher than in the old WB
* The menu items are also higher here than in the old WB
* The repository icon has font-size: 24px here and 22.4px in the old WB. On the other side the icon in the language selector is 24px in the old WB which makes it a bit inconsistent there, so I wonder if we should leave it this way here or try to comply with the old WB.

svilenvelikov:
The tooltip syling is also different from the old WB.
* The shadow is darker in the old one and also the border seems like more dark as well
* There is no bottom border between the header and body of the popover in the old or at least it's almost indistinguishable in contrast with this one where the border is clearly visible

packages/api/src/models/repositories/repository-type.ts
```diff
@@ -1,5 +1,6 @@
 export enum RepositoryType {
   GRAPH_DB = 'graphdb',
   ONTOP = 'ontop',
-  FEDEX = 'fedex'
+  FEDEX = 'fedex',
+  SYSTEM = 'System'
```
svilenvelikov, line 5: Is it `System` with capital `S` really?

packages/api/src/services/repository/mappers/repository-size-info.mapper.spec.ts
svilenvelikov, line 1: Didn't we decided to put tests in a test nested test package? Or we stared doing it only in the shared components?

packages/api/src/services/repository/repository-rest.service.ts
```diff
@@ -1,9 +1,18 @@
 import {Service} from '../../providers/service/service';
+import {Repository, RepositorySizeInfo} from '../../models/repositories';
 
 export class RepositoryRestService implements Service {
-  getRepositories(): Promise<Record<string, unknown[]>> {
+
+  static readonly REPOSITORIES_ENDPOINT = 'rest/repositories';
```
svilenvelikov, line 6: Should start with `/`!

packages/api/src/services/repository/repository-rest.service.ts
```diff
@@ -1,9 +1,18 @@
...
+
+  getRepositories(): Promise<Record<string, []>> {
```
svilenvelikov, line 8: Wasn't there a concrete type for the repository list? Use it!

packages/api/src/services/repository/repository.service.spec.ts
```diff
@@ -16,22 +16,36 @@ describe('RepositoryService', () => {
...
-
-      // When the service is called to fetch all repositories,
-      const repositories = await repositoryService.getRepositories();
-
-      // Then I expect it to return an instance of RepositoryList,
-      expect(repositories).toBeInstanceOf(RepositoryList);
-      // with all repositories included.
-      expect(repositories).toEqual(expectedResult);
-    });
+  test('should return repositories', async () => {
```
svilenvelikov, line 19: Always put the method name in the beginning of the description
`getRepositories should return a RepositoryList` or `getRepositorySizeInfo should return an instance of RepositorySizeInfo`

packages/api/src/services/repository/repository.service.spec.ts
```diff
@@ -16,22 +16,36 @@ describe('RepositoryService', () => {
...
+    expect(repositories).toEqual(expectedResult);
   });
 
+  test('should return an instance of RepositorySizeInfo', async () => {
+    const rawRepositorySizeInfo = {
+      inferred: 2,
+      total: 3,
+      explicit: 1
+    };
+    mockRepositoryRestService.getRepositorySizeInfo.mockResolvedValue(rawRepositorySizeInfo as unknown as RepositorySizeInfo);
```
svilenvelikov, line 41: I don't like how these tests are going! You are mocking the wrong part and with wrong data. And then, the inocation below is also wrong which might hide issues.
Lets discuss this tomorrow!

packages/shared-components/src/components/onto-repository-selector/onto-repository-selector.tsx
```diff
@@ -140,4 +153,86 @@ export class OntoRepositorySelector {
...
+                    <div class="value">${TranslationService.translate(this.canWriteRepoInLocation(repository) ? 'repository-selector.tooltip.accesses.read_write' : 'repository-selector.tooltip.accesses.read')}</div>
+                </div>`;
+
+    repositoryTooltipHtml += this.buildRepositorySizeInfoHtiml(repositorySizeInfo);
+
+    repositoryTooltipHtml += '</div>';
+    return repositoryTooltipHtml
+  }
+
+  private buildRepositorySizeInfoHtiml(repositorySizeInfo: RepositorySizeInfo): string {
```
svilenvelikov, line 192: typo in Htiml

svilenvelikov (changes requested):

packages/api/src/services/repository/mappers/repository-size-info.mapper.spec.ts
boyan-tonchev: We discussed it but did not make any decision.

packages/shared-components/src/components/onto-repository-selector/onto-repository-selector.tsx
boyan-tonchev: Done

packages/api/src/services/repository/repository.service.spec.ts
boyan-tonchev: Done

packages/api/src/services/repository/repository-rest.service.ts
boyan-tonchev: Done

packages/api/src/services/repository/repository-rest.service.ts
boyan-tonchev: Done

packages/api/src/models/repositories/repository-type.ts
boyan-tonchev: Fixed

packages/api/src/services/repository/repository.service.spec.ts
boyan-tonchev: Refactored

packages/api/src/services/repository/mappers/repository-size-info.mapper.spec.ts
boyan-tonchev: Moved into nested test packages.

packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
```diff
@@ -37,13 +37,18 @@ export class OntoDropdown {
   /**
    * The dropdown button tooltip. It will be used if present; otherwise, the {@link OntoDropdown#dropdownButtonTooltipLabelKey} will be used.
    */
-  @Prop() dropdownButtonTooltip: string;
+  @Prop() dropdownButtonTooltip: string | (() => Promise<string>);
```
svilenvelikov, line 40: we should think for some custom type for these `() => Promise<...>`

packages/shared-components/src/components/onto-repository-selector/onto-repository-selector.tsx
```diff
@@ -140,4 +153,86 @@ export class OntoRepositorySelector {
...
+      if (!repository) {
+        return Promise.resolve('');
+      }
+      return this.repositoryService.getRepositorySizeInfo(repository)
+        .then((repositorySizeInfo: RepositorySizeInfo) => this.buildRepositoryTooltipHtml(repository, repositorySizeInfo));
+    }
+  }
+
+  private buildRepositoryTooltipHtml(repository: Repository, repositorySizeInfo: RepositorySizeInfo): string {
+    let repositoryTooltipHtml = `
```
svilenvelikov, line 168: I wonder if this is the best way to compose these templates here and the other one below. I also think that these could eventually be a standalone component itself

packages/shared-components/src/components/onto-tooltip/models/onto-tooltip-configuration.ts
```diff
@@ -10,6 +10,9 @@ export class OntoTooltipConfiguration {
    */
   hideOnClick = false;
 
+  interactive = true;
+  appendTo = () => document.body;
```
svilenvelikov, line 14: document these

packages/shared-components/src/models/dropdown/dropdown-item.ts
```diff
@@ -15,13 +15,18 @@ export class DropdownItem<T> {
   /**
    * The dropdown item tooltip. It will be used if present; otherwise, the {@link DropdownItem._tooltipLabelKey} will be used.
    */
-  private _tooltip: string = '';
+  private _tooltip: string | (() => Promise<string>);
```
svilenvelikov, line 18: here again we have 3 instances of this return type! I think we can just have something like `export type Awaitable = () => Promise<T>` and use it instead.

packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
boyan-tonchev: Done

packages/shared-components/src/models/dropdown/dropdown-item.ts
boyan-tonchev: Done

svilenvelikov (approved):

PR #1756, opened 2024-12-16
GDB-11315 implement license notification

plamen-yordanov:
## What
Implement license alert component

## Why
To display a notification to the user, whenever a license is invalid

## How
- Added `license-alert-component`
- The new component is displayed in the header, when license is not valid, or an error occurred, when retrieving the license
- Added request for retrieving license info inside `ontotext-root-config.js` (to be done once globally)
- Added services, mappers and models needed for the new component
- Added `onto-stylesheet.css` for global styles

## Testing
jest unit tests

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/root-config/src/ontotext-root-config.js
+++ b/packages/root-config/src/ontotext-root-config.js
@@ -13,8 +13,10 @@ import {
 import microfrontendLayout from './microfrontend-layout.html';
 import './styles/bootstrap.min.css';
 import 'font-awesome/css/font-awesome.min.css';
+import './styles/onto-stylesheet.css';
 // import "./styles/bootstrap-graphdb-theme.css";
 import { defineCustomElements } from '../../shared-components/loader';
+import { ServiceProvider, LicenseService, LicenseContextService, TranslationService } from '@ontotext/workbench-api';
 
 addErrorHandler((err) => {
   console.error(err);
@@ -65,6 +67,18 @@ layoutEngine.activate();
 
 defineCustomElements();
 
+const loadLicense = () => {
+  const licenseContext = ServiceProvider.get(LicenseContextService);
+  ServiceProvider.get(LicenseService).getLicense().then(license => {
+    licenseContext.updateLicense(license);
+  }).catch(() => {
+    licenseContext.updateLicense({
+      message: TranslationService.translate('license_alert.no_license_set'),
+      valid: false
+    });
+  });
+};
+
 // This is a workaround to initialize the navbar when the root-config is loaded and the navbar is not yet initialized.
 const waitForNavbarElement = () => {
   return new Promise((resolve, reject) => {
@@ -93,6 +107,7 @@ const registerSingleSpaFirstMountListener = () => {
     window.singleSpaFirstMountListenerRegistered = true;
     window.addEventListener('single-spa:first-mount', () => {
       initializeNavbar();
+      loadLicense();
     });
   }
 };
--- a/packages/shared-components/cypress/e2e/header/header.cy.js
+++ b/packages/shared-components/cypress/e2e/header/header.cy.js
@@ -1,4 +1,5 @@
 import {HeaderSteps} from "../../steps/header/header-steps";
+import {TooltipSteps} from "../../steps/tooltip-steps";
 
 describe('Header', () => {
   it('Should render header and various tools inside', () => {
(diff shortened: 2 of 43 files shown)


packages/shared-components/src/assets/i18n/en.json
```diff
@@ -69,5 +69,9 @@
     "btn": {
       "toggle": "Choose repository"
     }
+  },
+  "license-alert": {
```
svilenvelikov, line 73: Use underscore instead of dash

packages/shared-components/src/components/onto-header/onto-header.tsx
```diff
@@ -1,21 +1,53 @@
-import { Component, Host, h } from '@stencil/core';
+import { Component, Host, h, State } from '@stencil/core';
+import { ServiceProvider, LicenseContextService, License } from '@ontotext/workbench-api'
 
+/**
+ * OntoHeader component for rendering the header of the application.
+ * This component includes a search component, license alert (if applicable),
```
svilenvelikov, line 6: Just to point out here, no action is needed. It includes them explicitly, for now. We plan to tweak the plugin system so that we can inject whatever we want declaratively.

packages/shared-components/src/components/onto-header/onto-header.tsx
```diff
@@ -1,21 +1,53 @@
...
   tag: 'onto-header',
   styleUrl: 'onto-header.scss',
   shadow: false,
 })
 export class OntoHeader {
+  /** The current license information */
+  @State() private license: License;
+
+  /** Array of subscription cleanup functions */
+  private readonly subscriptions: (() => void)[] = [];
```
svilenvelikov, line 19: I think it'd be better if we have an explicit type for the subscription, e.g. `type Subscription = () => void`. Also, a SubscriptionList class with methods to add a subscription and unsubscribe all subscriptions. It'd be neat to have these I think.

packages/root-config/src/ontotext-root-config.js
```diff
@@ -65,6 +67,18 @@ layoutEngine.activate();
 
 defineCustomElements();
 
+const loadLicense = () => {
+  const licenseContext = ServiceProvider.get(LicenseContextService);
+  ServiceProvider.get(LicenseService).getLicense().then(license => {
+    licenseContext.updateLicense(license);
+  }).catch(() => {
+    licenseContext.updateLicense({
```
svilenvelikov, line 75: Better create a local variable for this with the Partial<License> type or maybe cast it directly.

svilenvelikov (changes requested):

packages/shared-components/src/assets/i18n/en.json
plamen-yordanov: replaced dashes with underscores

packages/shared-components/src/components/onto-header/onto-header.tsx
plamen-yordanov: created new type `subscription` and class `subscription-list`. Added methods for adding a subscription and unsubscribing from all subscriptions

packages/root-config/src/ontotext-root-config.js
plamen-yordanov: Ignored as discussed, since this is JS

packages/shared-components/cypress/e2e/header/header.cy.js
```diff
@@ -13,4 +13,41 @@ describe('Header', () => {
     // And I should see the search component
     HeaderSteps.getSearch().should('be.visible');
   });
+
+  describe('OntoLicenseAlert', () => {
```
svilenvelikov, line 17: Move this outside in its own directory. Also do the same for the steps related with the license alert component.
Implement separate page for the license alert component and test it in isolation. Extend the header tests to check if the license alert appears there as well.

packages/shared-components/src/pages/js/main.js
```diff
@@ -0,0 +1,18 @@
+let headerContext = document.createElement('onto-test-context');
```
svilenvelikov, line 1: this is trivial, but you can rename this headerContext to testContext

packages/shared-components/src/pages/js/main.js
```diff
@@ -0,0 +1,18 @@
...
+document.body.appendChild(headerContext);
+
+// Mock the navigateUrl function which is exposed by the single-spa via the root-config module
+// in order to allow the menu to work without going anywhere when clicking the menu items
+window.singleSpa = {
+  navigateToUrl: function (url) {
+    console.log('%cnavigate', 'background: blue', url);
```
svilenvelikov, line 8: If this logger is just forgotten, then delete it

packages/shared-components/cypress/e2e/header/header.cy.js
plamen-yordanov: Moved the redirect test into `license-alert` directory and added `steps` file, along with `index.html` and `main.js`. The show/hide tests are left here. Also added assertions for the tooltip in this file. The reason, they are here is because I had Issues with the tooltip in `license-alert.cy.js`. For some reason the tests broke there due to the tooltip

packages/shared-components/src/pages/js/main.js
plamen-yordanov: deleted

packages/shared-components/src/pages/js/main.js
plamen-yordanov: renamed

packages/shared-components/src/pages/js/main.js
plamen-yordanov: deleted

packages/shared-components/src/pages/js/main.js
plamen-yordanov: renamed

svilenvelikov (approved):

svilenvelikov:
Looks good. Rebase, squash commits and lest integrate it!

PR #1757, opened 2024-12-16
GDB-11387 Pass the password property with the expected name to the UpdateUserPayload model

svilenvelikov:
## What
Pass the password property with the expected name to the UpdateUserPayload model.

## Why
A logged-in user can't edit his own password.

## How
Fixed the password property name passed to the user model so that it can be mapped correctly.  

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


boyan-tonchev (approved):

svilenvelikov:
Jenkins retry build

PR #1768, opened 2025-01-02
GDB-10571 Implement workbench footer component

plamen-yordanov:
## What
Implemented dynamic data loading for footer component

## Why
To populate the versions, mentioned in the footer dynamically.

## How
Added a request to `/rest/info/version?local=1` upon `single-spa:first-mount`. If the request is successful, we update the license context with the product info. The `onto-footer` component now gets the data from the context and visualises it, once it is available.

## Testing
Jest unit tests and cypress component tests

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/services/license/license-context.service.ts
+++ b/packages/api/src/services/license/license-context.service.ts
@@ -19,7 +19,7 @@ export class LicenseContextService extends ContextService {
 
   /**
    * Subscribes to changes in the license context
-   * .
+   *
    * @param callbackFn - A callback function that will be called when the license changes.
    * The callback receives the updated License object or undefined as its parameter.
    * @returns A function that, when called, will unsubscribe from the license changes.
--- a/packages/api/src/services/license/license-rest.service.ts
+++ b/packages/api/src/services/license/license-rest.service.ts
@@ -6,13 +6,12 @@ import { HttpService } from '../http/http.service';
  * Extends the HttpService to utilize its HTTP request capabilities.
  */
 export class LicenseRestService extends HttpService {
-
   /**
    * Retrieves the current license information from the GraphDB settings.
-   * 
+   *
    * This method sends a GET request to the '/rest/graphdb-settings/license' endpoint
    * to fetch the license details.
-   * 
+   *
    * @returns A Promise that resolves to a License object containing the current license information.
    */
   getLicense(): Promise<License> {
--- a/packages/shared-components/src/components/onto-footer/onto-footer.tsx
+++ b/packages/shared-components/src/components/onto-footer/onto-footer.tsx
@@ -1,23 +1,57 @@
-import { Component, Host, h } from '@stencil/core';
+import { Component, Host, h, State } from '@stencil/core';
+import { ProductInfo, ServiceProvider, SubscriptionList, ProductInfoContextService } from '@ontotext/workbench-api';
+import { TranslationService } from '../../services/translation.service';
 
+/**
+ * OntoFooter component for rendering the footer of the application.
+ * This component displays information about GraphDB, RDF4J, Connectors, and Workbench versions,
+ * as well as copyright information.
+ */
 @Component({
   tag: 'onto-footer',
   styleUrl: 'onto-footer.scss',
   shadow: false,
 })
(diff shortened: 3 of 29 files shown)


packages/api/src/services/license/license-rest.service.ts
```diff
@@ -1,21 +1,36 @@
...
+  /**
+   * Retrieves the local version information of the product.
+   *
+   * This method sends a GET request to the 'rest/info/version' endpoint with a <code>local=1</code> query parameter
+   * to fetch the local version details of the product. The value of the <code>local</code> query parameter
+   * does not matter. The presence of it will result in the server returning the local version details.
+   *
+   * @returns A Promise that resolves to a ProductInfo object containing the local version information.
+   */
+  getVersionLocal(): Promise<ProductInfo> {
```
svilenvelikov, line 33: rename this to `getProductInfo` which would be more meaningful

packages/shared-components/src/components/onto-footer/onto-footer.tsx
```diff
@@ -1,23 +1,57 @@
...
+ * as well as copyright information.
+ */
 @Component({
   tag: 'onto-footer',
   styleUrl: 'onto-footer.scss',
   shadow: false,
 })
 export class OntoFooter {
+  /** State variable to store product information */
+  @State() private productInfo: ProductInfo;
```
svilenvelikov, line 17: I don't think we need a state for this component. This data is loaded just once on app bootstrap and  is not a subject of a runtime change afterwards, so we can keep it safely as a normal field.

packages/api/src/services/license/license-context.service.ts
```diff
@@ -19,12 +21,32 @@ export class LicenseContextService extends ContextService {
...
   onLicenseChanged(callbackFn: ValueChangeCallback<License | undefined>): () => void {
     return this.subscribe(LicenseContextService.GRAPHDB_LICENSE, callbackFn);
   }
+
+  /**
+   * Updates the product information in the context.
+   *
+   * @param productInfo - The new ProductInfo object to be set in the context.
+   */
+  updateProductInfo(productInfo: ProductInfo): void {
```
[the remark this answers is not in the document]
plamen-yordanov: The reason I put it here is because the legacy workbench used the [license.rest.service.js](https://github.com/Ontotext-AD/graphdb-workbench/blob/eb6865d3fac30caa5020ebcc3af76e25b13fc296/src/js/angular/rest/license.rest.service.js#L94) for this request. Should I still create a new package?

packages/api/src/services/license/license-rest.service.ts
plamen-yordanov: Should it be `getProductInfoLocal`? There are two types of the version requests. One uses the `local` query param, which will get the local installation version info. The other one is without the `local` query param, which may return a different version, if the workbench is connected to a remote location. The second one adds a header and requires a location URI. I was planning to have 2 separate methods for both requests

packages/api/src/services/license/license-context.service.ts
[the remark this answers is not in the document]
svilenvelikov: Yes. The legacy workbench is not well structured and we should not be guided by its design.

packages/api/src/services/license/license-rest.service.ts
svilenvelikov: You are correct. Use the `getProductInfoLocal` name!

packages/api/src/services/license/license-context.service.ts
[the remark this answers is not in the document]
plamen-yordanov: Created product info services and moved the logic there

packages/api/src/services/license/license-rest.service.ts
plamen-yordanov: renamed to `getProductInfoLocal`

svilenvelikov (approved):

PR #1770, opened 2025-01-02
GDB-11282 introduce local storage service

svilenvelikov:
## What
Introduce a local storage service that will be used to persist data in the browser.

## Why
Some features in the workbench require data to be persisted locally in the browser.

## How
* Implemented a Persistence interface that must be implemented by every storage service. The Persistence name was chosen to avoid confusion with the native Storage interface.
* Implemented a LocalStorageService backed by the localStorage API.
* Implemented a StorageData wrapper model which provides convenient methods for access and convert the stored value.

## Testing
Implemented tests for the service and storage data wrapper.

## Screenshots
NA

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/services/storage/test/local-storage.service.spec.ts
+++ b/packages/api/src/services/storage/test/local-storage.service.spec.ts
@@ -1,0 +1,112 @@
+import { LocalStorageService } from '../local-storage.service';
+
+describe('LocalStorageService', () => {
+
+  const prefixedKey = 'ontotext.gdb.domain.testKey';
+
+  beforeEach(() => {
+    const storageMock = new StorageMock();
+    jest.spyOn(TestStorageService.prototype, 'getStorage')
+      .mockImplementation(() => storageMock);
+  });
+
+  test('set should store a string value automatically prefixing the key', () => {
+    const storageService = new TestStorageService();
+    const key = 'testKey';
+    const value = 'testValue';
+
+    storageService.set(key, value);
+
+    expect(storageService.getStorage().getItem(prefixedKey)).toBe(value);
+  });
+
+  test('get should retrieve a string value by an automatically prefixed key', () => {
+    const storageService = new TestStorageService();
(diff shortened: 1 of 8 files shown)


packages/api/src/services/storage/test/local-storage.service.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    expect(storageService.getStorage().getItem(prefixedKey)).toBe(value);
+  });
+
+  test('set should store a null value', () => {
+    const storageService = new TestStorageService();
+    const key = 'testKey';
+
+    // @ts-expect-error - Testing the behavior of storing a null value
+    storageService.set(key, null);
```
boyan-tonchev, line 28: IMO, the signatures of `LocalStorageService.set` and `LocalStorageService.storeValue` should be updated to support null as a value. I noticed that in `LocalStorageService.storeValue`, `localStorage` accepts only strings. However, the method could be modified so that if the value is not null, it sets the value; otherwise, it removes the value.
```
storeValue(key: string, value: string | null): void {
    const prefixedKey = this.getPrefixedKey(key);
    if (value) {
      this.getStorage().setItem(prefixedKey, value);
    } else {
      this.getStorage().removeItem(key);
    }
  }
```
With this change, the @ts-expect-error can be removed, and we won't encounter a compilation error if we need to set null for any reason in the code.

svilenvelikov: I prefer not to do this. There is a method for removing values from the store and I don't like the idea to remove mapped values using the storeValue method. Regarding these tests for set/get null values, I removed them. I added them by mistake, because my storage stub is backed by a js object which allows null values in contrast of the Storage API that doesn't.

boyan-tonchev (approved):

PR #1771, opened 2025-01-06
GDB-11319 - Create denied permission alert banner

DesiBorisova:
## What
The "permission denied" banner has been created in the new WB. Functionality to be added with another MR.

## Why
This was one of the elements due to be migrated.

## How
I created a new shared component with the banner and styled it like the one in the legacy WB.

## Testing
Test added.

## Screenshots
Screenshot from 2025-01-09 12-24-46

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/shared-components/src/components/onto-layout/onto-layout.scss
+++ b/packages/shared-components/src/components/onto-layout/onto-layout.scss
@@ -33,6 +33,10 @@
     margin-left: 0;
   }
 
+  .default-slot-wrapper {
+    display: none;
+  }
+
   footer {
     grid-area: footer;
   }
--- a/packages/shared-components/src/components/onto-layout/onto-layout.tsx
+++ b/packages/shared-components/src/components/onto-layout/onto-layout.tsx
@@ -26,6 +26,7 @@ export class OntoLayout {
   @Element() hostElement: HTMLOntoLayoutElement;
 
   @State() isLowResolution = false;
+  @State() hasPermission = true; // TODO get from local store
 
   /**
    * Event listener for the navbar toggled event. The layout needs to respond properly when the navbar is toggled in
@@ -86,6 +87,10 @@ export class OntoLayout {
   render() {
     return (
       <Host class="wb-layout">
+        {/* Default slot is explicitly defined to be able to hide the main content in the user permission check */}
+        <div class="default-slot-wrapper">
+          <slot name="default"></slot>
+        </div>
         <header class="wb-header">
           <onto-header></onto-header>
         </header>
@@ -94,8 +99,11 @@ export class OntoLayout {
           <onto-navbar navbar-collapsed={this.isLowResolution} selected-menu={this.currentRoute}></onto-navbar>
         </nav>
 
-        <slot name="main"></slot>
-
+        {
+          this.hasPermission ?
+            <slot name="main"></slot> :
+            <onto-permission-banner></onto-permission-banner>
+        }
         <footer class="wb-footer">
           <onto-footer></onto-footer>
         </footer>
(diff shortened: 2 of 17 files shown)


packages/shared-components/src/components/onto-permission-banner/onto-permission-banner.scss
```diff
@@ -0,0 +1,36 @@
...
+  padding: 0 2rem 3rem 2rem;
+}
+
+.label-container {
+  display: flex;
+  flex-direction: column;
+  margin-left: 1rem;
+}
+
+.alert {
```
svilenvelikov, line 19: All alert related styles should go here https://github.com/Ontotext-AD/graphdb-workbench/blob/GDB-10373-workbench-with-single-spa/packages/root-config/src/styles/onto-stylesheet.css and also use different class names (prefix them with `onto-`) to avoid interference with bootstrap

DesiBorisova: Updated

packages/shared-components/src/components/onto-layout/onto-layout.tsx
```diff
@@ -83,8 +84,14 @@ export class OntoLayout {
...
         </nav>
 
-        <slot name="main"></slot>
-
+        {
+          this.hasPermission ?
+            <slot name="main"></slot> :
+            <onto-permission-banner></onto-permission-banner>
+        }
+        <div class="default-slot-wrapper">
```
svilenvelikov, line 92: Nice you found this solution. I just can suggest if you can move this in the beginning of the component template.

packages/shared-components/src/components/onto-layout/onto-layout.scss
```diff
@@ -33,6 +33,10 @@
     margin-left: 0;
   }
 
+  .default-slot-wrapper {
```
svilenvelikov, line 36: I played a bit with this and I think that you probably don't need this.

packages/shared-components/src/components/onto-layout/onto-layout.tsx
svilenvelikov: And also, drop a comment above why this was needed for further reference if you don't mind

packages/shared-components/src/components/onto-permission-banner/onto-permission-banner.scss
svilenvelikov, line 1: I know that currently our styles are a bit screwed, but can you confirm that the banner looks this way on your side as well
image
For reference, this is how it looks like in current WB
image

packages/shared-components/src/components/onto-layout/onto-layout.tsx
DesiBorisova: Moved and added descriptive comment

packages/shared-components/src/components/onto-layout/onto-layout.scss
DesiBorisova: It appears to be required. While the main content doesn't render if this styling is removed, the `default slot` remains empty and present, creating the view in the screenshot: 
Screenshot from 2025-01-09 10-49-36

packages/shared-components/src/components/onto-permission-banner/onto-permission-banner.scss
DesiBorisova: Yes, it's the same on my side as well. I stretched out the banner more, so it resembles the legacy WB one. Currently it's like this: 
Screenshot from 2025-01-09 12-24-46

I will update the MR description.

svilenvelikov (approved):

PR #1772, opened 2025-01-08
GDB-11385 Fix issues with d3

plamen-yordanov:
## What
Fix issues with d3

## Why
Previously, some files directly accessed the `d3` property, which was global. Now with ESM, that is no longer possible and Visual graph, Class relationships and Class hierarchy views are not loading due to that.

## How
Added necessary import statements in the files, which use `d3` 
Added property to skip `onpopstate` event in `graphs-visualizations.controller.js` since it caused double loading of links

## Testing
existing

## Screenshots
image
image
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


boyan-tonchev:
LGTM

svilenvelikov (approved):

PR #1776, opened 2025-01-13
GDB-11319 - Add functionality to permission banner

DesiBorisova:
## What
- The user denied permission banner will appear when the page is unauthorized;
- The onto-navbar remains aligned to the top of the page regardless of the margin-top set on the onto-layout. For instance, if onto-layout has margin-top: 100px, the navbar will still stay fixed at the top of the page.
image

## Why
- The banner was added as a component, but the functionality was not;
- This behavior occurs because there is a top: 0 declaration on the onto-navbar container.

## How
- I added a context service, which communicates between the new and old WB. The onto-layout also listens for route changes, in order to know that the flag for the banner should be re-calculated. The UserPermissionStorageService remains from a previous implementation of the solution, but is not used here otherwise;
- Remove the top: 0 CSS declaration.

## Testing
Tests added.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/services/event-service/event.service.ts
+++ b/packages/api/src/services/event-service/event.service.ts
@@ -1,0 +1,76 @@
+import {Event} from '../../models/events';
+import {Service} from '../../providers/service/service';
+import {ObjectUtil} from '../utils';
+import {EventObserver} from '../../models/events/event-observer';
+
+/**
+ * A service that manages event subscriptions and emissions.
+ *
+ * Allows components or modules to subscribe to specific events and be notified when those events are emitted.
+ */
+export class EventService implements Service {
+
+  /**
+   * A map of event names to their corresponding list of subscribers.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private eventSubscribers: Map<string, EventObserver<any>[]> = new Map<string, EventObserver<any>[]>();
+
+  /**
+   * Emits an event to all subscribers of the specified event type.
+   *
+   * @template T - The type of the event payload.
+   * @param event - The event to emit, containing its name and payload.
+   */
+  emit<T extends {} | undefined>(event: Event<T>): void {
+    const subscribers = this.eventSubscribers.get(event.NAME);
+    if (subscribers) {
+      subscribers.forEach((subscriber) => {
+        subscriber.notify(ObjectUtil.deepCopy(event.payload));
+      });
+    }
+  }
+
+  /**
+   * Subscribes to a specific event with a callback function.
+   *
+   * @template T - The type of the event payload.
+   * @param eventName - The name of the event to subscribe to.
+   * @param callback - A callback function to handle the event notifications.
+   *
+   * @returns A function to unsubscribe from the event.
+   */
+  subscribe<T extends {} | undefined>(eventName: string, callback: (payload: T) => void): () => void {
+    const observer = new EventObserver<T>(callback);
+    const eventSubscribers = this.eventSubscribers.get(eventName);
+
+    if (eventSubscribers) {
+      eventSubscribers.push(observer);
+    } else {
+      this.eventSubscribers.set(eventName, [observer]);
+    }
+    return () => this.unsubscribe(eventName, observer);
+  }
+
+  /**
+   * Unsubscribes an observer from a specific event type.
+   *
+   * @template T - The type of the event payload.
+   * @param eventName - The name of the event to unsubscribe from.
+   * @param observer - The observer to remove from the subscribers list.
+   *
+   * @private
+   */
+  private unsubscribe<T extends {} | undefined>(eventName: string, observer: EventObserver<T>): void {
+    const eventSubscribers = this.eventSubscribers.get(eventName);
+    if (eventSubscribers) {
+      const index = eventSubscribers.indexOf(observer);
+      if (index !== -1) {
+        eventSubscribers.splice(index, 1);
+      }
+      if (eventSubscribers.length === 0) {
+        this.eventSubscribers.delete(eventName);
+      }
+    }
+  }
+}
--- a/packages/api/src/services/security/security-context.service.ts
+++ b/packages/api/src/services/security/security-context.service.ts
@@ -1,0 +1,47 @@
+import {ContextService} from '../context';
+import {ValueChangeCallback} from '../../models/context/value-change-callback';
+import {DeriveContextServiceContract} from '../../models/context/update-context-method';
+import {RestrictedPages} from '../../models/security';
+
+type SecurityContextFields = {
+  readonly RESTRICTED_PAGES: string
+}
+
+type SecurityContextFieldParams = {
+  readonly RESTRICTED_PAGES: RestrictedPages;
+}
+
+/**
+ * The SecurityContextService class manages the various fields in the security context.
+ */
+export class SecurityContextService extends ContextService<SecurityContextFields> implements DeriveContextServiceContract<SecurityContextFields, SecurityContextFieldParams> {
+  readonly RESTRICTED_PAGES = 'restrictedPages';
+
+  /**
+   * Retrieves the restricted pages for the user.
+   *
+   * @return a map with restricted pages.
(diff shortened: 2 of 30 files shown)


packages/api/src/services/security/security-context.service.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/**
+ * The SecurityContextService class manages the various fields in the security context.
+ */
+export class SecurityContextService extends ContextService<SecurityContextFields> implements DeriveContextServiceContract<SecurityContextFields, SecurityContextFieldParams> {
+  readonly HAS_PERMISSION = 'permission';
+
+  getPermissions(): DeniedPermissions | undefined {
+    return this.getContextPropertyValue('permission');
```
svilenvelikov, line 21: use the constant

packages/api/src/services/user-permission/user-permission-storage.service.ts
```diff
@@ -0,0 +1,14 @@
...
+import {StorageData} from '../../models/storage';
+
+export class UserPermissionStorageService extends LocalStorageService {
+  readonly NAMESPACE = 'auth';
+
+  set(key: string, value: string): void {
+    this.storeValue(key, value);
+  }
+
+  get(key: string): StorageData {
```
svilenvelikov, line 11: There is no need to override this method

packages/shared-components/src/components/onto-layout/onto-layout.tsx
```diff
@@ -66,6 +76,21 @@ export class OntoLayout {
     service.handleStorageChange(event);
   }
 
+  private setPermission(permissions: DeniedPermissions) {
+    const mainContent = document.querySelector('.wb-layout main') as HTMLElement;
```
svilenvelikov, line 80: move this down, before the code where it's used

packages/api/src/models/security/denied-permissions.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Represents a map of denied paths.
+ * - Each key is a path that is denied.
+ * - The value is always `true`, indicating the path is denied.
+ */
+export class DeniedPermissions {
```
svilenvelikov, line 6: There is no value DeniedPermissions to be a class in the way how it's used currently and is kind of misleading. If the idea was to just define the type, then better make it simply an interface or a type (`type DeniedPermissions = Record<string, true>`). If the impact in the legacy-workbech wouldn't be big, then my oppinion is that it'd be better if we back this by a map, and expose some methods for getting and setting values in the map inside, like `denyPermission(location: string): void` and `hasPermission(location: string): boolean`. This would encapsulate the logic and make it more flexible and typesafe.

packages/api/src/services/security/test/security-context.service.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+describe('SecurityContextService', () => {
+  let securityContextService: SecurityContextService;
+
+  beforeEach(() => {
+    securityContextService = new SecurityContextService();
+  });
+
+  test('Should call onUserPagePermissionChange callback with correct denied routes', () => {
+    const onPermissionChangedCallBackFunction = jest.fn();
+    const newPermissions = {'/testPage': true} as DeniedPermissions;
```
svilenvelikov, line 13: Give the suggestion for this class, this can become for example:
```
const newPermissions = new DeniedPermissions();
newPermissions.denyPermission('/testPage');
```

packages/shared-components/src/components/onto-layout/onto-layout.tsx
```diff
@@ -2,7 +2,8 @@ import {Component, Host, h, Listen, Element, State} from '@stencil/core';
 import {NavbarToggledEvent} from "../onto-navbar/navbar-toggled-event";
 import {debounce} from "../../utils/function-utils";
 import {WINDOW_WIDTH_FOR_COLLAPSED_NAVBAR} from "../../models/constants";
-import {ServiceProvider, LocalStorageSubscriptionHandlerService} from "@ontotext/workbench-api";
+import {ServiceProvider, LocalStorageSubscriptionHandlerService, SecurityContextService} from "@ontotext/workbench-api";
+import {DeniedPermissions} from '../../../../api/src/models/security/denied-permissions';
```
svilenvelikov, line 6: fix this import!!!

packages/shared-components/src/components/onto-layout/onto-layout.tsx
```diff
@@ -21,12 +24,13 @@ export class OntoLayout {
 
   constructor() {
     this.windowResizeObserver = debounce(() => this.windowResizeHandler(), 50);
+    this.securityContextService = ServiceProvider.get(SecurityContextService);
```
svilenvelikov, line 27: this can be assigned directly to the field declaration

svilenvelikov (changes requested):

packages/shared-components/src/components/onto-layout/onto-layout.tsx
```diff
@@ -52,6 +54,12 @@ export class OntoLayout {
     this.windowResizeObserver();
   }
 
+  @Listen('single-spa:routing-event',  {target: 'window'})
```
svilenvelikov, line 57: I prefer single-spa stuff to be handled in root-config and not to be propagated in the other modules

packages/api/src/services/event-service/event.service.ts
```diff
@@ -0,0 +1,89 @@
+import {Event} from '../../models/events';
+import {Service} from '../../providers/service/service';
+import {ObjectUtil} from '../utils';
+
+/**
+ * Represents an observer for events, encapsulating a callback to handle event notifications.
+ *
+ * @template T - The type of the event payload.
+ */
+export class EventObserver<T> {
```
svilenvelikov, line 10: move this in a separate file

packages/api/src/services/event-service/event.service.ts
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Allows components or modules to subscribe to specific events and be notified when those events are emitted.
+ */
+export class EventService implements Service {
+
+  /**
+   * A map of event names to their corresponding list of subscribers.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private eventSubscribers: Record<string, EventObserver<any>[]> = {};
```
svilenvelikov, line 35: Can we use a WeakMap here

packages/shared-components/src/components/onto-layout/onto-layout.tsx
boyan-tonchev: Refactored

packages/api/src/services/security/test/security-context.service.spec.ts
boyan-tonchev: done

packages/api/src/services/event-service/event.service.ts
boyan-tonchev: done

packages/api/src/services/event-service/event.service.ts
boyan-tonchev: Added logic that removes the key if the value array is empty

packages/shared-components/cypress/e2e/permission-banner/restricted-pages.cy.js
```diff
@@ -0,0 +1,61 @@
+import {RestrictedPagesSteps} from "../../steps/restricted-pages/restricted-pages-steps";
+import {PermissionBannerSteps} from "../../steps/permission-banner/permission-banner-steps";
+
+describe("Registered pages", () => {
```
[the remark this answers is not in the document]
boyan-tonchev: fixed

svilenvelikov (approved):

boyan-tonchev:
Jenkins retry build

PR #1784, opened 2025-01-21
GDB-11508 Add transloco to workbench module

plamen-yordanov:
## What
Add transloco i18n to workbench module

## Why
So the workbench can have dynamic i18n support

## How
- Added `@jsverse/transloco` package
- Added `transloco-bootstrap.ts`, where transloco is configured and set to listen to language change
- Moved single-spa `start()` after `bootstrap.js` promises are resolved, to ensure bundles are available, before other modules start.
- Added `en.json` and `fr.json` bundle files for the workbench module

Also:
 - Split bootstrap functions into separate modules, to avoid cluttering `ontotext-root-config.js`
 - Removed modules from the `workbench` micro-frontend to have a standalone component structure
 - Moved the `path` of `@ontotext/workbench-api` inside the base `tsconfig.json`, since both `app` and `spec` files use the same path. This fixed the issue, where typescript wasn't able to resolve imports from `@ontotext/workbench-api`

## Testing
None, as there is no testable new code. Only an example of the new functionality

## Screenshots
GraphQL in english: 
image

GraphQL: in french:
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/workbench/src/app/app.routes.ts
```diff
@@ -1,3 +1,9 @@
-import { Routes } from '@angular/router';
+import {Routes} from '@angular/router';
+import {SparqlComponent} from './sparql/sparql.component';
+import {GraphqlComponent} from './graphql/graphql.component';
 
-export const routes: Routes = [];
+export const routes: Routes = [
+  {path: 'sparql', component: SparqlComponent},
```
boyan-tonchev, line 6: Could you change this path from 'sparql' to 'sparql-new' or something else? We have a 'sparql' URL that is currently being processed by the legacy workbench.

packages/workbench/src/app/app.routes.ts
```diff
@@ -1,3 +1,9 @@
...
+  {path: 'graphql', component: GraphqlComponent},
+  // {path: "**", component: EmptyRouteComponent}
```
boyan-tonchev, line 8: Uncomment this line and implement the component. It should be a component with empty HTML. Without this default path definition, the Angular router will throw an exception and break navigation to the view if we click on a link leading to a view managed by the legacy workbench.

packages/workbench/src/app/app.routes.ts
plamen-yordanov: Added `/sparql-new` path

packages/root-config/src/bootstrap/language/language-bootstrap.js
```diff
@@ -0,0 +1,61 @@
...
+        languageContextService.updateDefaultBundle(defaultBundle);
+        if (isStoredAndDefaultLangEqual) {
+          languageContextService.updateLanguageBundle(defaultBundle);
+        }
+      }
+    })
+    .catch((error) => console.error('Could not load language configuration', error));
+};
+
+const onLanguageChange = () => {
```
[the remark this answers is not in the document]
plamen-yordanov: returned `Promise.resolve()`

svilenvelikov (approved):

boyan-tonchev (approved):

PR #1813, opened 2025-02-11
GDB-11563 set endpoint default state

svilenvelikov:
## What
Allow changing which is the default graphql endpoint. There can be only one default endpoint.

## Why
This allows the user to switch which is the default endpoint from the UI.

## How
Implemented operation handler in the controller and a UI control for the user to interact with.

## Testing
Implemented tests.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/src/js/angular/graphql/controllers/graphql-endpoint-management-view.controller.js
+++ b/src/js/angular/graphql/controllers/graphql-endpoint-management-view.controller.js
@@ -3,6 +3,7 @@ import '../services/graphql-context.service';
 import './graphql-endpoint-configuration-modal.controller';
 import {GraphqlEventName} from "../services/graphql-context.service";
 import {endpointUrl} from "../models/endpoints";
+import {resolvePlaygroundUrlWithEndpoint} from "../services/endpoint-utils";
 
 const modules = [
     'graphdb.framework.core.services.graphql-service',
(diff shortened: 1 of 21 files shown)


src/js/angular/graphql/controllers/graphql-endpoint-management-view.controller.js
```diff
@@ -86,6 +93,38 @@ function GraphqlEndpointManagementViewCtrl($scope, $location, $interval, $reposi
...
+        GraphqlService.editEndpointConfiguration($repositories.getActiveRepository(), endpointInfo.endpointId, updateEndpointRequest)
+            .then(() => {
+                toastr.success(
+                    $translate.instant('graphql.endpoints_management.table.actions.set_as_default.success',
+                        {endpointId: endpointInfo.endpointId})
+                );
+                return loadEndpointsInfo(false);
+            })
+            .catch((error) => {
+                // something wen wrong while setting the default endpoint so we need to revert the changes
```
boyan-tonchev, line 116: Previously, before sending the request, we set `endpointInfo.default = true`. We may need to revert this setting if an error occurred.

svilenvelikov: Well. I'm doing it. See line 119

svilenvelikov: done

boyan-tonchev (approved):

PR #1824, opened 2025-02-17
GDB-11536 Fix broken layout and restricted pages tests

plamen-yordanov:
## What
Fix broken layout and restricted pages tests

## Why
They started failing after a change in loading menu items in the layout.

## How
 - Added mock items for the `window.PluginRegistry.get()` function
 - Increase cypress' `defaultCommandTimeout` to 10000 milliseconds
 - Removed global event listener and added `onClick` events directly on the links instead
 - Replaced string DOM manipulation with document fragments to reduce querying and repainting. Now only the `main` element content is changed
 - Added `renderPage()` before restricting/unrestricting a page, to make sure the layout is set up correctly, before the event
 - Added extra step to wait for `onto-layout` to load, as it's a heavy operation. This gives the test some extra time to complete

## Tests
cypress

## Screenshots
Local test run
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


plamen-yordanov:
Jenkins retry build

plamen-yordanov:
Jenkins retry build

plamen-yordanov:
Jenkins retry build

plamen-yordanov:
Jenkins retry build

packages/shared-components/src/pages/restricted-pages/main.js
```diff
@@ -1,9 +1,10 @@
-let bannerElement = document.querySelector("onto-permission-banner");
+const app = document.getElementById('onto-layout');
 
 const setDefaultRestrictedPages = () => testContext.updateRestrictedPage({
     '/pages/restricted-pages/restricted-page': true});
 
 const makeCurrentPageRestricted = () => {
+    renderPage();
```
boyan-tonchev, line 7: The `renderPage(); `function should be called from both `makeRestrictedPagesMapUndefined` and `makeRestrictedPagesMapEmpty` , as well as from.

plamen-yordanov: I was wondering, weather this is needed. I will add it

plamen-yordanov: > as well as from.

I assume this should be `as well as from setDefaultRestrictedPages`?
 Added to all 3 functions

boyan-tonchev (approved):

PR #1834, opened 2025-02-21
GDB-11245 Introduce entity index size for repositories

plamen-yordanov:
## What
Introduce `entityIndexSize` property configuration for repositories

## Why
Following the use case where thousands of repositories with little to no data can exist in a single GraphDB, we need to allow users that create repositories through the workbench to be able to configure the entity index size. This property impacts the initial size of the in-memory collection used to map entity ids and directly impacts the per-repository heap consumption on each created repository.

## How
Added new number input field, below `Enable full-text search (FTS) index` checkbox The value for the field is the `entityIndexSize` from the repository configuration Added validation for min value of the field, which is 10K. Refactored some functions regarding this, to be more generic and effective

## Tests
cypress

## Screenshots
valid value:
image

invalid value:
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


src/js/angular/repositories/controllers.js
```diff
@@ -22,22 +24,21 @@ export const getFileName = function (path) {
...
-            params.leftJoinWorkerThreads.value = parseInt(params.leftJoinWorkerThreads.value);
-            params.boundJoinBlockSize.value = parseInt(params.boundJoinBlockSize.value);
-            params.joinWorkerThreads.value = parseInt(params.joinWorkerThreads.value);
-            params.unionWorkerThreads.value = parseInt(params.unionWorkerThreads.value);
-        }
+    if (!params) {
+        return;
     }
+
+    Object.keys(params).forEach(key => {
```
boyan-tonchev, line 31: I think this have to be reverted, and extended only with what is needed. The new function parses all values of string to numbers, but I think this is not the truly for all properties. I looked for usage of this params object and found that there is a string that not are number. I this implementation we can some string that is valid number but have to be string, for example. I know why this is made like this, but IMO we have to leave the old variant for now, during migration process we will refactoring this, then we will have better understanding of all posible situations.

 @svilenvelikov, what do you think?

svilenvelikov: I haven't looke in detail this MR yet, but generally that's the idea. We refactor and improve only the minimum which we understand and are sure to not break something.

plamen-yordanov: I could add `&& numberParamToErrorKey[key]` to line 35, which will ensure only number properties, which are in the map, will be parsed. The result will be the same properties, plus `entityIndexSize` e.g. :
```
    queryLimitResults
    queryTimeout
    validationResultsLimitTotal
    validationResultsLimitPerConstraint
    joinWorkerThreads
    leftJoinWorkerThreads
    unionWorkerThreads
    boundJoinBlockSize
    entityIndexSize
```

Are you OK with this, or should I revert?

svilenvelikov: ok

svilenvelikov (approved):
don't forget to squash

PR #1837, opened 2025-02-21
GDB-11561 import graphql endpoint definition

svilenvelikov:
## What
Implemented import graphql endpoint definitions operation.

## Why
This allows users to quickly generate graphql endpoints given they already have endpoint definitions in SOML format.

## How
Implemented the import command in the graphql endpoint management view. The command opens a modal dialog where the user can select and upload definitions, either single or multiple files at once. Supported file formats are yaml and zip containing yaml files.

## Testing
Tests implemented.

## Screenshots
image
image
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


src/js/angular/core/filters/bytes-filter.js
```diff
@@ -0,0 +1,29 @@
...
+ * @returns {string} Formatted string with appropriate unit (bytes, kB, MB, GB, TB, PB)
+ * @example
+ * // returns "1.5 MB"
+ * {{ 1572864 | bytes }}
+ *
+ * // returns "1.50 MB"
+ * {{ 1572864 | bytes:2 }}
+ */
+angular.module('graphdb.framework.core.filters.bytes', [])
+    .filter('bytes', function () {
```
yordanalexandrov, line 17: There is similar logic in `heap-memory-chart.js`

Can something be reused?

src/js/angular/import/controllers/import-view.controller.js
```diff
@@ -807,6 +807,7 @@ importViewModule.controller('UploadCtrl', ['$scope', 'toastr', '$controller', '$
     const uploadFile = (file, startImport, nextCallback) => {
         $scope.settings.name = file.name;
         const data = UploadRestService.createUploadPayload(file, $scope.settings);
+        console.log('%cuploaddata', 'background: yellow', data, file);
```
yordanalexandrov, line 810: Remvove or change to debug

src/js/angular/rest/graphql.rest.service.js
```diff
@@ -175,6 +175,25 @@ function GraphqlRestService($http) {
...
+     * @param {FormData|File} payload The request payload.
+     * @param {*} canceler The canceler object.
+     * @returns {Promise<unknown>} The response from the backend.
+     */
+    const importEndpointDefinition = (repositoryId, payload, canceler) => {
+        return Upload.http({
+            url: `${REPOSITORIES_ENDPOINT}/${repositoryId}/graphql/manage/endpoints/import`,
+            data: payload,
+            headers: {
+                'Content-Type': undefined
```
yordanalexandrov, line 191: Why is it undefined?

src/js/angular/rest/upload.rest.service.js
```diff
@@ -37,6 +37,7 @@ function UploadRestService($http, Upload, $translate) {
      * @return {*}
      */
     function uploadUserDataFile(repositoryId, file, data) {
+        console.log('%cupload service', 'background: yellow', file, data);
```
yordanalexandrov, line 40: Same

src/js/angular/rest/graphql.rest.service.js
svilenvelikov: This allows the type to be obtained by the browser. Not sure if it's strictly mandatory though.

svilenvelikov: Yeah, the content-type header must be set, and as far as we allow different file types to be imported, it's easier to set it this way and allow the appropriate type to be set from the browser.

src/js/angular/core/filters/bytes-filter.js
svilenvelikov: Well, It's possible to reuse the filter I added in the chart and not the other way round because the convertor from the chart is kind of tied with it. Anyway. I don't wan't to deal with refactoring in other modules in this MR.

boyan-tonchev (approved):

PR #1862, opened 2025-03-06
GDB-11826 Extend create/edit agent options

yordanalexandrov:
## What
Extend the create and edit agent steps to be able to configure which fields to be filled

## Why
This will allow the guide creator to add specificity to the guide steps

## How
- added options for name, model and user instructions to both create and edit steps
- added `disable` option for extraction method, which specifies if the method should be enabled or disabled
- extracted agent configuration into its own complex step to reduce repeating
- extracted `end-on-api-key-error` and added a check if the toaster is for the api-key
- added another core step `wait-for-element-to-show` which waits for an element to be shown before proceeding with next step

## Testing
N/A

Definitions to test:
```json
{
      "guideBlockName": "create-ttyg-agent",
      "editName": true,
      "model": "gpt-4o-mini",
      "userInstructions": "Talk like a frog",
      "options": {
        "methods": [
          {
            "guideBlockName": "sparql-search-method",
            "options": {
              "ontologyGraph": "https://kg.ontotext.com/resource/ontology/"
            }
          }
        ]
      }
    },
    {
      "guideBlockName": "edit-ttyg-agent",
      "model": "gpt-4o",
      "options": {
        "methods": [
          {
            "guideBlockName": "sparql-search-method",
            "options": {
              "disable": true
            }
          },
          {
            "guideBlockName": "fts-search-method"
          }
        ]
      }
    }
```

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [-] Tests


src/js/angular/guides/steps/complex/ttyg/common/configure-agent/plugin.js
```diff
@@ -0,0 +1,68 @@
...
+                        content: 'guide.step_plugin.configure-agent.name-input',
+                        class: 'input-agent-name-guide-dialog',
+                        url: '/ttyg',
+                        beforeShowPromise: () => GuideUtils.waitFor(GuideUtils.getGuideElementSelector('agent-form'), 5)
+                            .catch((error) => {
+                                services.toastr.error(services.$translate.instant('guide.unexpected.error.message'));
+                                throw error;
+                            }),
+                        elementSelector: GuideUtils.getGuideElementSelector('agent-name'),
+                        showOn: () => options.name,
```
boyan-tonchev, line 38: The name of the option is too generic. 'editName' or something like that would be more meaningful.

boyan-tonchev, line 38: Refactor the method 'getSteps' to conditionally add this step if the option name ('editName') is passed, and remove 'showOn'. There's no need to add the steps and skip them at runtime; we can conditionally add them at 'compile' time. The same applies to other steps that depend on the option and not on the system state.

src/js/angular/guides/steps/complex/ttyg/create-ttyg-agent/plugin.js
```diff
@@ -5,16 +5,8 @@ PluginRegistry.add('guide.step', [
...
-
-                return methods.map((method) => {
-                    return {
-                        guideBlockName: method.guideBlockName,
-                        options: method.options
-                    };
-                });
-            };
+            // Set name field as mandatory for creation
+            options.name = true;
```
boyan-tonchev, line 9: Don't do this here. The configuration should only be on the 'configure-agent' step:
```
{
    guideBlockName: 'configure-agent',
    options: angular.extend({}, options, {editName: true})
}

```

src/js/angular/guides/steps/complex/ttyg/common/configure-agent/plugin.js
yordanalexandrov: Renamed to `editName`

src/js/angular/guides/steps/complex/ttyg/create-ttyg-agent/plugin.js
yordanalexandrov: Passed only to the `configure-agent`

boyan-tonchev (approved):

PR #1867, opened 2025-03-07
fix(GDB-11838) Add Cypress admin login & free access toggle commands

teodossidossev:
## WHAT
- Introduce new Cypress commands (loginAsAdmin, switchOffSecurity, switchOffFreeAccess) that capture and reuse the Authorization header.
- Update user, repository, and security commands to pass the admin token in requests.
- Adjust tests to use the new commands and reduce repetitive cleanup logic.

## WHY
- We need a standardized way to authenticate and manage security features in our Cypress tests.

## HOW
- Added loginAsAdmin to capture the Authorization token from the response and store it in Cypress.env.
- Reworked existing commands (switchOffSecurity, switchOffFreeAccess, etc.) to use the stored token.

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [X] Tests

### DIFF
--- a/src/js/angular/graphql/plugin.js
+++ b/src/js/angular/graphql/plugin.js
@@ -7,8 +7,7 @@ PluginRegistry.add('route', [
         'controller': 'GraphqlEndpointManagementViewCtrl',
         'title': 'menu.graphql-endpoint-management.label',
         'helpInfo': 'graphql.endpoints_management.helpInfo',
-        'documentationUrl': 'graphql-endpoint-management.html',
-        'allowAuthorities': ['WRITE_REPO_{repoId}', 'WRITE_REPO_{repoId}:GRAPHQL']
+        'documentationUrl': 'graphql-endpoint-management.html'
     },
     {
         'url': '/graphql/endpoint/create',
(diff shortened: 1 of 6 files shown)


src/js/angular/graphql/plugin.js
```diff
@@ -7,8 +7,7 @@ PluginRegistry.add('route', [
         'controller': 'GraphqlEndpointManagementViewCtrl',
         'title': 'menu.graphql-endpoint-management.label',
         'helpInfo': 'graphql.endpoints_management.helpInfo',
-        'documentationUrl': 'graphql-endpoint-management.html',
-        'allowAuthorities': ['WRITE_REPO_{repoId}', 'WRITE_REPO_{repoId}:GRAPHQL']
```
svilenvelikov, line 11: Why this is changed in this MR?

test-cypress/support/security-command.js
```diff
@@ -11,15 +11,56 @@ Cypress.Commands.add('switchOnSecurity', () => {
     });
 });
 
-Cypress.Commands.add('switchOffSecurity', () => {
+Cypress.Commands.add('loginAsAdmin', () => {
```
svilenvelikov, line 14: This command is not implemented correctly. It must return the promise so that the chaining to work properly
```
     cy.loginAsAdmin().then(()=> {
                cy.deleteUser(user, true);
                cy.switchOffSecurity(true);
            });
```
Right now I'm not convinced that this would work reliably

svilenvelikov: Same applies for the other new commands below!

svilenvelikov (changes requested):

test-cypress/support/security-command.js
teodossidossev: returns added

svilenvelikov (approved):

PR #1868, opened 2025-03-07
fix(GDB-11795) Fix cookie saving in free access mode

teodossidossev:
## WHAT:
- Ensures that local storage is used properly for free-access users.
- Updates the TrackingService to read from local storage or backend consistently, depending on the access mode

## WHY:
- Cookie handling was not working in freaa access mode.

## HOW:
- Removes the internal getCookieConsent() call in the CookiePolicyModalController and injects the same instance from the directive
- Ensures that isTrackingAllowed() and getCookieConsent() are used consistently in TrackingService, and re-initializes tracking when user consent updates.
- Adds LSKeys.COOKIE_CONSENT as a known local storage key and leverages LocalStorageAdapter appropriately.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/src/js/angular/core/services/tracking/tracking.service.js
+++ b/src/js/angular/core/services/tracking/tracking.service.js
@@ -12,7 +12,8 @@ const modules = [
 
 ];
 angular.module('graphdb.framework.core.services.trackingService', modules)
-    .service('TrackingService', ['$window', '$jwtAuth', '$translate', '$licenseService', 'toastr', 'InstallationCookieService', 'GoogleAnalyticsCookieService', TrackingService]);
+    .factory('TrackingService', ['$window', '$jwtAuth', '$licenseService', 'InstallationCookieService', 'GoogleAnalyticsCookieService', 'LocalStorageAdapter', 'LSKeys',
+        TrackingService]);
 
 /**
  * Service for managing user tracking based on their cookie consent preferences.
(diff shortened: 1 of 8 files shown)


src/js/angular/core/services/tracking/tracking.service.js
```diff
@@ -12,7 +12,8 @@ const modules = [
 
 ];
 angular.module('graphdb.framework.core.services.trackingService', modules)
-    .service('TrackingService', ['$window', '$jwtAuth', '$translate', '$licenseService', 'toastr', 'InstallationCookieService', 'GoogleAnalyticsCookieService', TrackingService]);
+    .service('TrackingService', ['$window', '$jwtAuth', '$translate', '$licenseService', 'toastr', 'InstallationCookieService', 'GoogleAnalyticsCookieService', 'LocalStorageAdapter', 'LSKeys',
+        TrackingService]);
```
svilenvelikov, line 16: why the service is injecting itself?

teodossidossev: this is the service's function

svilenvelikov: hmm I now noticed that you've used a service and not a factory.

svilenvelikov (approved):

PR #1880, opened 2025-03-12
GDB-11877 Implement interactive guide autostart

yordanalexandrov:
## What
Implement logic to autostart specific guides when user opens the workbench with a query parameter `autostartGuide` pointing to a guideId

## Why
This will allow to generate links which will automatically start a specific guide. This is required by the GDB Sandbox project, where users are given a link which will open the workbench and start a guide

## How
- add new LSKeys for storing a flag which allows auto started guide to be disabled upon completion or upon user request
- add logic in `mainCtrl` to check for the query param `autostartGuide` and if found to call `GuidesService.autoStartGuide`
- add new method `GuidesService.autoStartGuide` to fetch the specified guide and run it
- refactored some logic in `ShepherdService`
  - removed default alert when canceling the guide and replaced it with a new custom modal
  - add checks if the auto start of the guide is disabled
  - changed the cancellation logic a bit

## Testing
N/A

## Screenshots
image
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


src/js/angular/guides/steps/core/plugin.js
```diff
@@ -174,10 +180,10 @@ PluginRegistry.add('guide.step', [
                 content: options.content || 'guide.step_plugin.guide-ended.content',
                 lastStep: true,
                 show: (guide) => () => {
-                    guide.options.confirmCancel = false;
+                    // guide.options.confirmCancel = false;
```
svilenvelikov, line 183: Document why this is commented out or delete it

src/js/angular/guides/guides.service.js
```diff
@@ -236,12 +236,28 @@ function GuidesService(
...
             ShepherdService.resumeGuide(guide.guideId, stepsDescriptions, startStepId);
         } else {
-            ShepherdService.startGuide(guide.guideId, stepsDescriptions, startStepId);
+            ShepherdService.startGuide(guide.guideId, stepsDescriptions, startStepId, isAutoStarted);
         }
 
         $rootScope.$broadcast('guideStarted');
     };
 
+    this.autoStartGuide = (autoStartGuideId) => {
```
svilenvelikov, line 245: Add normalization or check for this argument because right now you rely on the guideId and the request parameter to be of the same type which is string, but the guideId can be also a number and then the logic below might break.

src/js/angular/guides/tour-lib-services/shepherd.service.js
```diff
@@ -179,15 +207,65 @@ function ShepherdService($location, $translate, LocalStorageAdapter, $route, $in
...
+
+        if (currentStep.options.isLastStep) {
+            this._completeGuide(guide);
+            return;
+        }
+
+        guide.hide();
+
+        return ModalService.openCustomModal({
+            templateUrl: 'js/angular/guides/templates/modal-confirm-exit.html',
```
svilenvelikov, line 230: For some reason when I try to stop the guide via the [x] button I get this error and the guide remains active
image

svilenvelikov (changes requested):

src/js/angular/guides/steps/core/plugin.js
yordanalexandrov: Forgot to remove

src/js/angular/guides/tour-lib-services/shepherd.service.js
yordanalexandrov: Could it be that you did not rebuid the app when you switched the workspace?

The tamplate must be coppied by webpack, so a rebuild is needed.

src/js/angular/guides/guides.service.js
yordanalexandrov: Added check and documentation

yordanalexandrov: Added check and conversion to string if `autoStartGuideId` is number.

Also added explicit conversion of `guide.guideId` to string as it is used as a further in the services.

svilenvelikov (approved):

PR #1882, opened 2025-03-12
GDB-11318 implement global notification component

plamen-yordanov:
## What
Implement global notification component

## Why
To be able to display statuses of different ongoing operations

## How
- Created `onto-operations-notification` dropdown button, which displays different operations as icons and respective count. The button is a dropdown toggle, so when clicked, it opens a list of different operations. When an individual operation is clicked, it redirects to a page, related to the operation

- Refactored `onto-header` to display/hide the new notification component and poll the operations on every 2 seconds

- Migrated `fibonacci-generator`, which is currently used to slow down requests, if errors occur

- Added models and services

## Testing
jest and cypress

## Screenshots
Screenshot from 2025-03-12 13-49-16

Screenshot from 2025-03-12 13-52-08

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/shared-components/src/components/onto-operations-notification/onto-operations-notification.scss
```diff
@@ -0,0 +1,165 @@
...
+  & .operations-dropdown {
+    position: absolute;
+    right: 0;
+    z-index: 1000;
+    border: none;
+    background-color: var(--primary-color, #9e2b0a);
+    border-radius: 0;
+    list-style: none;
+    padding: 0;
+    -moz-transition: all 150ms ease-out; /* FF4+ */
```
svilenvelikov, line 60: we should add the autoprefixer library and write just the standard's compliant css rules

packages/shared-components/src/components/onto-operations-notification/onto-operations-notification.tsx
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  private toggleDropdown = (event: Event) => {
+    event.stopPropagation();
+    this.isOpen = !this.isOpen;
+  };
+
+
+  private toOperationsGroupSummary() {
+    const groupToOperationSummary = new Map<OperationGroup, OperationGroupSummary>();
```
svilenvelikov, line 113: Is it possible this mapping building to be moved in the activeOperations model maybe. Better avoid exposing the model internals in components to reduce the coupling

svilenvelikov (changes requested):

packages/api/src/models/monitoring/operation-group-summary.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export class OperationGroupSummary extends Model<OperationGroupSummary> {
+  group: OperationGroup;
+  totalOperations: number;
+  status: OperationStatus;
+  id: string;
+
+  constructor(data: OperationGroupSummary) {
+    super();
+    this.id = window.crypto.randomUUID();
```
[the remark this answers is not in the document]
plamen-yordanov: Can the service/util use `window.crypto.randomUUID()` internally, or should I implement something?

packages/api/src/models/monitoring/operation-status-summary.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Currently, these could be backup, sparql querying and cluster operations
+ */
+export class OperationStatusSummary extends Model<OperationStatusSummary> {
+  status: OperationStatus;
+  allRunningOperations: OperationList;
+
+  constructor(operationStatusSummary: Partial<OperationStatusSummary>) {
+    super();
+    this.status = operationStatusSummary.status || OperationStatus.INFORMATION;
+    this.allRunningOperations = MapperProvider.get(OperationListMapper).mapToModel(operationStatusSummary.allRunningOperations || []);
```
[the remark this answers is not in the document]
plamen-yordanov: This won't default to a native array. It will pass a native empty array to the mapper, which will result in an empty `OperationList`

packages/shared-components/src/components/onto-header/onto-header.tsx
```diff
@@ -43,11 +77,56 @@ export class OntoHeader {
...
   }
 
-  /**
-   * Lifecycle method called when the component is about to be removed from the DOM.
-   * Cleans up all subscriptions.
-   */
-  disconnectedCallback(): void {
-    this.subscriptions.unsubscribeAll();
+  private startOperationPolling() {
+    this.pollingInterval = setInterval(() => {
```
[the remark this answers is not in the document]
plamen-yordanov: The legacy workbench currently starts polling, when the component is loaded, without clearing the interval beforehand. The interval is cleared upon `$destroy`.

image

The current implementation will stop polling, if the repositoryContext sends `repositoryId: undefined` as well. Otherwise it will start and stop, like the legacy

packages/shared-components/src/components/onto-operations-notification/onto-operations-notification.scss
plamen-yordanov: In a different task, I assume?

packages/api/src/models/monitoring/operation-group-summary.ts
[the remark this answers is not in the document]
svilenvelikov: Yes. Just beware that our services are singletons and we don't want to cache the window inside.

packages/api/src/models/monitoring/operation-status-summary.ts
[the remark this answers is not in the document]
svilenvelikov: I see. My bad.

packages/shared-components/src/components/onto-operations-notification/onto-operations-notification.scss
svilenvelikov: yes

packages/api/src/models/monitoring/operation-group-summary.ts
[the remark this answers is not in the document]
plamen-yordanov: moved to `generator-utils`

packages/shared-components/src/components/onto-header/onto-header.tsx
```diff
@@ -9,30 +22,51 @@ import { ServiceProvider, LicenseContextService, License, SubscriptionList } fro
...
 export class OntoHeader {
+  private readonly monitoringService = ServiceProvider.get(MonitoringService);
+  private readonly repositoryContextService = ServiceProvider.get(RepositoryContextService);
+  private readonly securityContextService = ServiceProvider.get(SecurityContextService);
+  private readonly authenticationService = ServiceProvider.get(AuthenticationService);
+  private readonly UPDATE_ACTIVE_OPERATION_TIME_INTERVAL = 2000;
+  private readonly fibonacciGenerator = new FibonacciGenerator();
+
+  private repositoryId?: string;
+  private pollingInterval: NodeJS.Timeout;
```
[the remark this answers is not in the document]
plamen-yordanov: added `number` type. Also added `window.setInterval` to ensure typescript knows it's a browser interval and not a nodejs. If you don't like depending directly on window here, the other option is //@ts-ignore, since typescript doesn't know, which interval it is.

packages/shared-components/src/components/onto-header/onto-header.tsx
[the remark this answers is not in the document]
plamen-yordanov: added `clearInterval` at the start just in case

packages/shared-components/src/components/onto-operations-notification/onto-operations-notification.tsx
plamen-yordanov: moved inside the model

plamen-yordanov:
Jenkins retry build

packages/api/src/models/monitoring/operation-status-summary.ts
```diff
@@ -19,4 +23,35 @@ export class OperationStatusSummary extends Model<OperationStatusSummary> {
...
+      const summary = groupToOperationSummary.get(operation.group)!;
+      summary.totalOperations += operation.count;
+
+      if (STATUS_ORDER[operation.status] > STATUS_ORDER[summary.status]) {
+        summary.status = operation.status;
+      }
+    });
+
+    const operationGroupArray = Array.from(groupToOperationSummary.values());
+    return MapperProvider.get(OperationGroupSummaryListMapper).mapToModel(operationGroupArray);
```
svilenvelikov, line 55: OK, but now you put a service dependency in the model. Although it is a mapper, I don't think it's a good idea. You can return the array from this method and mode this mapping in the client component.

plamen-yordanov: Should I limit mapper usage only to constructors? I've used list mappers in other models and I'm not sure, where to draw the line

svilenvelikov: Models should not have dependencies than other models. If you've used mappers in models, then those should also be refactored.

plamen-yordanov: Removed the dependency entirely. Moved the `OperationGroupSummaryListMapper#mapToModel` inside `OperationSummaryMapper`

plamen-yordanov:
Also: 
- Replaced `onto-operations-monitoring` with `onto-header` in the page file to use the same response mock and model mapping.
- Changed the component dropdown to close upon link click, since this is the original behavior

svilenvelikov (approved):

PR #1885, opened 2025-03-13
GDB-11722 Fix broken file upload when maxUploadSize is less than 1MB

plamen-yordanov:
## What
Fix broken file upload, when maxUploadSize is less than 1MB

## Why
File upload wasn't working

## How
Previously the maxUploadSize bytes were converted into integer MB. That value was also added as `ngf-max-size` to the import. Values, which were less than 1MB resulted in 0 as max value, hence nothing was allowed. Now the exact `maxUploadSize` value is used to set `ngf-max-size` and the value is converted into a readable label only for display purposes

## Testing
existing

## Screenshots
exceeded size
Screenshot from 2025-03-13 11-41-49

within boundaries
Screenshot from 2025-03-13 11-47-24

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/src/pages/import.html
+++ b/src/pages/import.html
@@ -52,7 +52,7 @@
                                ngf-select
                                ngf-keep="false"
                                ngf-change="fileSelected($files, $file, $newFiles, $duplicateFiles, $invalidFiles)"
-                               ngf-max-size="maxUploadFileSizeMB + 'MB'">
+                               ngf-max-size="maxUploadFileSizeBytes">
                                 <div class="grid-container">
                                     <em class="icon-upload icon-lg"></em>
                                     <!-- Keep this a label, it points to the input element created by ngFileUpload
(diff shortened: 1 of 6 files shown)


src/pages/import.html
```diff
@@ -61,7 +61,7 @@ <h1>
                                         <div>{{'upload.rdf.files.label' | translate}}</div>
                                         <small
                                             class="text-muted">{{'all.rdf.formats.label' | translate}}{{'up.to' | translate}}
-                                            {{maxUploadFileSizeMB | number}} MB
+                                            {{maxUploadFileSizeBytes | bytes}}
```
svilenvelikov, line 64: I think that the filter can be improved a little bit more to round to whole number when under 1kB. See the screenshot
image

plamen-yordanov: Only for bytes? Is it OK to have this for example 
image

plamen-yordanov: done
image

svilenvelikov (approved):

PR #1894, opened 2025-03-19
GDB-11283 Implement toastr component

plamen-yordanov:
## What
Implement toastr component for notifications

## Why
To have the ability to display info, success, warning and error notification

## How
- Created `onto-toastr-service`, which holds methods for `info`, `success`, `warning` and `error` toast messages. When called, each of these methods emits an event with the message and type

- Created `onto-toastr` component, which listens for those events, and displays the relevant notification.

- Added config file, which currently contains config for position of the notifications, as well as timeout, until the toasts disappear

Toasts can display regular text, as well as markup If a toast is hovered, the timeout is removed, so the toast remains visible. When the cursor is removed, the timeout is re-added, so the notification can disappear

## Testing
jest/cypress

## Screenshots
Example, just for showcase:
Screenshot from 2025-03-19 16-36-10

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/shared-components/src/components/onto-toastr/onto-toastr.scss
```diff
@@ -0,0 +1,60 @@
...
+    bottom: 12px;
+    right: 12px;
+  }
+
+  &.bottom-left {
+    bottom: 12px;
+    left: 12px;
+  }
+
+  & .onto-toast {
```
svilenvelikov, line 17: Also consider that if you have a longer unbreakable string, the toastr won't render it properly
image

packages/shared-components/src/components/onto-toastr/onto-toastr.tsx
```diff
@@ -0,0 +1,108 @@
...
+   */
+  @State() private toasts: ToastMessageList = new ToastMessageList([]);
+
+  /**
+   * Map to track timeout IDs for each toast message
+   */
+  private toastToTimeout = new Map<ToastMessage, number>();
+
+  /** Configuration options, such as toast timeout */
+  private readonly config = ToastrConfig.getDefaultConfig();
```
svilenvelikov, line 34: * I'd prefer if we get an external configuration and merge with this internal one so that we can reconfigure behavior without the need of a new build here.
Also I wonder if it's a good idea to have some tests for runtime reconfiguration as well as having different configurations on different pages in an SPA.
Another case is: show a toastr, navigate to another page without redirect. Verify if the toastr is still visible.

plamen-yordanov: > I'd prefer if we get an external configuration and merge with this internal one so that we can reconfigure behavior without the need of a new build here.

will do.

> Also I wonder if it's a good idea to have some tests for runtime reconfiguration as well as having different configurations on different pages in an SPA.

Shouldn't the tooltip behavior be identical throughout the entire app?

> Another case is: show a toastr, navigate to another page without redirect. Verify if the toastr is still visible.

Currently the legacy workbench clears tooltips upon route change
https://github.com/Ontotext-AD/graphdb-workbench/blob/master/src/app.js#L218

svilenvelikov: > Currently the legacy workbench clears tooltips upon route change

... which is stupid idea, because there are bunch of places where they invented hacks with setTimeout to workaround it. I've no idea why they did it, but IMO it's better if we don't do that.

svilenvelikov: > Shouldn't the tooltip behavior be identical throughout the entire app?
It will. Most likely. But there might be cases where on the same page you might want to customize the behavior for one particular notification, although  this is a different scenario and is more related with how you configure each notification. Currently you can only provide a type and a message. I believe that we need a separate configuration for each type and also to be able to pass runtime config with the notification request.

plamen-yordanov: - Added option to pass config for individual toasts. Configuration for individual toasts (currently timeout), will override the default config, if provided
- Added `word-break: break-word` for longer strings

packages/shared-components/src/components/onto-toastr/onto-toastr.scss
```diff
@@ -24,6 +24,7 @@
     padding: 25px 0;
     color: white;
     margin-bottom: 5px;
+    word-break: break-word;
```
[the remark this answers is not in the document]
plamen-yordanov: changed

PR #1901, opened 2025-03-24
GDB-11317 Implement user menu

plamen-yordanov:
## What
Implement user menu component inside the header

## Why
To have visibility of logged-in user, as well as settings and logout

## How
Created `onto-user-menu` component, which displays a button with the currently logged-in user. The component will be displayed, when security is on, and there is an authenticated user When clicked, it shows a dropdown with `My-settings` and `Logout` options. The logout is currently not implemented When an external user is logged, there is no logout option The header is subscribed to changes for user and security configuration, and determines, weather to display the new component

## Testing
cypress

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/shared-components/cypress/e2e/user-menu/user-menu.cy.js
+++ b/packages/shared-components/cypress/e2e/user-menu/user-menu.cy.js
@@ -1,0 +1,52 @@
+import {UserMenuSteps} from "../../steps/user-menu/user-menu-steps";
+
+describe('User Menu', () => {
+  it('should render user menu', () => {
+    // Given I've opened a page with the user menu
+    UserMenuSteps.visit();
+
+    // And the user menu is visible
+    UserMenuSteps.getUserMenu()
+      .should('be.visible')
+      .and('have.text', 'john.doe');
+
+    // When, I click on the menu
+    UserMenuSteps.openUserMenu();
+
+    const dropdownOptions = ['My settings', 'Logout'];
+    // Then the dropdown should be displayed and have My Settings and Logout options
+    UserMenuSteps.getDropdownItems()
+      .each(($item, index) => {
+        cy.wrap($item).should('have.text', dropdownOptions[index]);
+      });
+
+    // When, I am logged in as external user
+    UserMenuSteps.setExternalUser();
+
+    // Then I expect to not see a logout option in the dropdown
+    UserMenuSteps.getDropdownItems()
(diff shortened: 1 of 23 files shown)


packages/shared-components/cypress/e2e/user-menu/user-menu.cy.js
```diff
@@ -0,0 +1,55 @@
...
+
+  it('should redirect to the correct URL when clicking on a dropdown option', () => {
+    // Given I've opened a page with the user menu
+    UserMenuSteps.visit();
+
+    // And the user menu is visible
+    UserMenuSteps.getUserMenu().should('be.visible');
+
+    // When, I click on the menu
+    UserMenuSteps.getUserMenu().click();
```
svilenvelikov, line 41: It's better to abstract DOM access in the steps. This is their purpose. Instead of doing clicks, getting `dropdown.children().first().click()` etc., created methods like `openUserMenu()`, `selectSettings()`, etc.

packages/shared-components/src/components/onto-user-menu/onto-user-menu.scss
```diff
@@ -0,0 +1,96 @@
...
+.onto-user-menu {
+  position: relative;
+  width: fit-content;
+
+  button {
+    display: flex;
+    flex-direction: row;
+    line-height: 1.5;
+    font-weight: 400;
+    font-size: 1rem;
```
svilenvelikov, line 14: The menu button doesn't scale down on lower resolutions
image

svilenvelikov: Actually it looks entirely broken and not just for this menu.

plamen-yordanov: Added global `@media (max-width: 1439px)`, which changes `font-size: 14px`. This comes from the `bootstrap-graphdb-theme.css`. The language and repository select need to have their media queries either removed, or set identical to the global one. Which one should it be?

svilenvelikov: Remove those from shared components. They should rely on the client and not to enforce their own hooks when possible.

PR #1902, opened 2025-03-24
GDB-11865: Remove built-in GraphDB Free license and related things

avataar:
## What
- Handle situations where GraphDB is running without any license
- Improved usability of license errors and validation

## Why
GraphDB 11 will be distributed without a built-in Free license.

## How
- Recognize the state of "no license" explicitly
- Moved license error (no license, expired license, etc) to the top of the home view for better visibility
- Replaced "Revert to Free" with "Remove license"
- When setting a license via the license view, show the license even if it's expired/wrong version/etc (with a disabled OK button)
- Cookie tracking enabled for all non-commercial licenses (no license, Free, Sandbox, evaluation)
- Reconciled some differences between various views/modals that show license info
- Decreased font size in license info details
- Extract a GraphDB Free license from Jenkins for the tests


docker-compose.yml
```diff
@@ -15,6 +15,9 @@ services:
       - 7200
     volumes:
       - ./test-cypress/fixtures/graphdb-import:/opt/home/import-data/
+      - type: bind
```
teodossidossev, line 18: can be shortened to `- ./graphdb.license:/opt/graphdb/home/conf/graphdb.license`

src/js/angular/core/services/license.service.js
```diff
@@ -46,33 +46,25 @@ function licenseService($window, $document, LicenseRestService, $translate) {
             .catch(() => {
                 _isLicenseHardcoded = true;
                 _license = {
-                    message: $translate.instant('no.license.set.msg'),
+                    message: "No license was set",
```
teodossidossev, line 49: this should be a flag/constant, not a string

src/js/angular/core/services/license.service.js
```diff
@@ -46,33 +46,25 @@ function licenseService($window, $document, LicenseRestService, $translate) {
...
-        } else if (_productType === "enterprise") {
-            _productTypeHuman = "Enterprise";
-        } else if (_productType === "free") {
-            _productTypeHuman = "Free";
-        } else if (_productType === "graphdb") {
-            _productTypeHuman = "GraphDB";
-        }
+    const isTrackableLicense = () => {
+        const licenseTypeOfUse = _license && _license.typeOfUse.toLowerCase();
+        return !isLicensePresent() || productType() === "free" || productType() === "sandbox"
```
teodossidossev, line 62: should be extracted to constants

src/js/angular/core/services/license.service.js
```diff
@@ -83,6 +75,14 @@ function licenseService($window, $document, LicenseRestService, $translate) {
         return _license;
     };
 
+    const licenseErrorMsg = () => {
+        if (!_license?.present && _license?.message === "No license was set") {
+            return $translate.instant('no.license.set.msg');
+        } else {
+            return $translate.instant("error.license", {message: _license?.message});
```
teodossidossev, line 82: do this actually translate? Does `instant` interpolate  `{{'common.error'|translate}}`?

teodossidossev (changes requested):

src/js/angular/core/services/license.service.js
avataar: Yes

teodossidossev (approved):

PR #1904, opened 2025-03-26
GDB-11287 - fix chart axis resetting when using chart filters

DesiBorisova:
## What
When filtering the `Monitoring`  charts, the axis will show the correct ranges, without cutting off a part of the data.
The y-axis names will not overlap long tick labels.

## Why
No value was passed in for the `multiplier` and the wrong function parameter was used to calculate the y-axis max value. Hence, the y-axis max value would reset to 1 when the chart data was toggled.

## How
I made sure the function receives the `multiplier` value where it is expected.
I added a bigger gap between the axis name and the tick label.

## Testing
N/A

## Screenshots
After the fix. The y-axis values won't reset to 0 and 1 when the charts are toggled.

https://github.com/user-attachments/assets/b235ab8a-9b61-4e7f-aae9-fab36bb94965

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/src/js/angular/resources/chart-models/performance/connections-chart.js
+++ b/src/js/angular/resources/chart-models/performance/connections-chart.js
@@ -46,7 +46,7 @@ export class ConnectionsChart extends ChartData {
     }
 
     updateRange(dataHolder, multiplier) {
-        const [max, minInterval] = ChartData.getIntegerRangeForValues(dataHolder, this.selectedSeries)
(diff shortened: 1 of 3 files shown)


svilenvelikov (approved):

svilenvelikov:
The branch has wrong JIRA ID

PR #1906, opened 2025-03-27
GDB-11267 destroy tooltip on node removal

plamen-yordanov:
## What
Add functionality to remove existing tooltip, when its target element has been removed from the dom

## Why
Because currently, when an element, which has a tooltip is removed, the tooltip stays visible

## How
Added a mutation observer, which listens to changes in the body and monitors removed nodes. When a node is removed, it is checked to see, if it has a tooltip. If a tooltip is present, it is destroyed.

## Testing
cypress

## Screenshots
before: 
before-fix

after:
after-fix

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
+++ b/packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
@@ -15,6 +15,8 @@ export class OntoTooltip {
     private static readonly ATTR_TRIGGER = 'tooltip-trigger';
     private static readonly ATTR_APPEND_TO = 'tooltip-append-to';
 
+    private observer: MutationObserver;
+
     /**
      * Checks if the target of the 'mouseover' event has tooltip configuration.
      * If it does, creates a tooltip instance and displays it.
@@ -60,6 +62,14 @@ export class OntoTooltip {
     }
     }
 
+    componentWillLoad() {
+        this.handleRemovedNodes();
+    }
+
+    disconnectedCallback() {
+        this.observer?.disconnect();
+    }
+
     render() {
         return <Host></Host>;
     }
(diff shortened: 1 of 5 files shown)


packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
```diff
@@ -72,6 +78,37 @@ export class OntoTooltip {
...
+            });
+        });
+    };
+
+    /**
+     * Initializes a MutationObserver to monitor the document body for removed nodes.
+     * When nodes are removed, it triggers the clearing of tooltips associated with those nodes.
+     */
+    private handleRemovedNodes(): void {
+        this.observer = new MutationObserver(this.clearTooltipsForRemovedElements);
```
svilenvelikov, line 108: You should take care of removing this observer on destroy.

packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
```diff
@@ -72,6 +78,37 @@ export class OntoTooltip {
...
+     */
+    private clearTooltipsForRemovedElements = (records: MutationRecord[]) => {
+        records.forEach((record) => {
+            record.removedNodes.forEach((node) => {
+                if (node instanceof HTMLElement) {
+                    const target = this.getTooltipTarget(node);
+                    if (target) {
+                        const tooltipInstance = this.getTooltipInstance(target);
+                        if (tooltipInstance) {
+                            tooltipInstance.destroy();
```
svilenvelikov, line 95: Should we also remove the _tippy from the element?

plamen-yordanov: No, according to documentation the [destroy()](https://atomiks.github.io/tippyjs/v5/methods/#destroy:~:text=The%20_tippy%20property%20is%20deleted%20from%20the%20reference%20element%20upon%20destruction.) method does that

packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
plamen-yordanov: added disconnect

packages/shared-components/src/components/onto-tooltip/onto-tooltip.tsx
```diff
@@ -60,6 +62,14 @@ export class OntoTooltip {
     }
     }
 
+    componentWillLoad() {
+        this.handleRemovedNodes();
+    }
+
+    disconnectedCallback() {
+        this.observer.disconnect();
```
svilenvelikov, line 70: Check if it's present. Better to be safe :)

svilenvelikov (approved):

PR #1914, opened 2025-03-31
GDB-11314 Implement rdf search view

plamen-yordanov:
## What
Implement RDF search component view

## Why
To have the ability to search rdf resources. It is split in view and logic parts, to reduce the load of large MRs

## How
Implemented a base `onto-search-resource-input`, which is will be the reusable part in the application. It contains the input field and will contain the results of the search

Implemented `onto-rdf-search.tsx`, which uses the reusable search and adds a search icon on top of it and is responsible for opening and closing the component in the header.

Added functionality in the header to intercept the click on the icon, and if it is done, while the user is on the homepage, then he is shown a toast message to use the existing visual search

Added font awesome light font and stylesheet, as the currently used icons look closer to the legacy ones.

Refactored the tooltip to cleanup children of removed nodes

## Testing
cypress

## Screenshots
open rdf search:
Screenshot from 2025-03-31 14-10-51

closed rdf search:
Screenshot from 2025-03-31 14-11-04

rdf search on home page:
Screenshot from 2025-03-31 14-13-55

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/shared-components/src/components/onto-header/onto-header.tsx
```diff
@@ -43,6 +49,15 @@ export class OntoHeader {
   /** Menu should appear, when security is enabled and user is authenticated */
   @State() private showUserMenu: boolean;
 
+  /** Whether the search component should appear */
+  @State() private shouldShowSearch: boolean = true;
+
+  @Listen('single-spa:routing-event', {target: 'window'})
```
svilenvelikov, line 55: It'd be better if we don't tie with the single-spa events directly. For this particular event we have our own which we can use here 
image

plamen-yordanov:
Jenkins retry build

packages/shared-components/src/components/onto-header/onto-header.tsx
plamen-yordanov: Replaced with `NavigationEnd` subscription

plamen-yordanov:
Also added conditions to show/hide the rdf search icon. Conditions are presence of repositoryId, active repository location and not having a loading request for active repository location (unless the user is on `/repository`

packages/shared-components/src/components/onto-header/onto-header.tsx
```diff
@@ -144,4 +187,27 @@ export class OntoHeader {
...
     return this.securityConfig.enabled && this.authenticationService.isAuthenticated(this.securityConfig, this.user);
   }
+
+  private captureSearchClick() {
+    return (event: MouseEvent) => {
+      const target = event.target as HTMLElement;
+      if (target.id !== this.searchIconId) {
+        return;
+      }
+      if (isHomePage()) {
```
svilenvelikov, line 197: I shomehow think that this should not be done here in the header. Can we have some config comming from above that controls the visibility of the rdf search component.
It'd have been even better if we had the plugin system in place so we can control this, but we are not there yet.

packages/shared-components/src/utils/routing-utils.ts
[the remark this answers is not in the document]
plamen-yordanov: moved to `api`

packages/shared-components/src/components/onto-search-resource-input/onto-search-resource-input.scss
```diff
@@ -0,0 +1,82 @@
...
+
+      &:hover:not(.selected) {
+        background-color: #d4d4d4;
+        border-color: #d4d4d4;
+      }
+    }
+  }
+
+  & .hint {
+    font-size: 12px;
```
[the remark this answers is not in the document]
plamen-yordanov: replaced with `rem`

packages/shared-components/src/components/onto-header/onto-header.tsx
plamen-yordanov: refactored to show a dummy icon, when on the homepage and removed the coupling between header and rdf search

svilenvelikov (approved):

PR #1926, opened 2025-04-03
GDB-11955 - Introduce cluster Create dialog improvements

DesiBorisova:
## What
Several improvements added to cluster create dialog:

- field for new node gets auto-focused;
- current node replace is not allowed;
- added new locations can be removed during cluster creation (exception is the Current node);
- copy functionality added for locations.

## Why
The changes improve the user experience.

## How
I extended the existing copy-to-clipboard directive with a custom tooltip for the copy action notification. I added the necessary logic for input focus of the dynamic fields.

## Testing
N/A

## Screenshots
Cluster creation

https://github.com/user-attachments/assets/c438e6c7-acbf-4ff2-9646-6bac92521733

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/src/css/clustermanagement.css
+++ b/src/css/clustermanagement.css
@@ -225,10 +225,18 @@ div.nodetooltip {
     -webkit-user-select: none;
     user-select: none;
     word-break: break-all;
-
     display: flex;
     align-items: center;
     justify-content: flex-start;
+}
+
+.location-item .text-only:hover {
+    text-decoration: 1px underline dashed var(--color-success-dark);
+    text-underline-offset: 4px;
+}
+
+.location-item:hover .custom-link-icon {
+    opacity: 1;
 }
 
 .location-item.hoverable:not(.list-group-item-danger):hover {
--- a/src/i18n/locale-en.json
+++ b/src/i18n/locale-en.json
@@ -194,7 +194,8 @@
             "actions": {
                 "replace_node": "Replace node",
                 "delete_node": "Delete node",
-                "cannot_delete_node": "You cannot remove more than half of the healthy nodes at the same time",
+                "cannot_delete_node": "The current node cannot be removed",
+                "cannot_replace_node": "You cannot replace this node",
                 "add_node": "Add node",
                 "add_node_tooltip": "Add node to current cluster",
                 "cancel": "Cancel",
@@ -208,6 +209,11 @@
             },
             "field_placeholders": {
                 "location": "http://my-hostname:7200"
+            },
+            "copy_to_clipboard": {
+                "copied": {
+                    "tooltip": "IRI copied"
+                }
             }
         }
     },
@@ -2474,6 +2480,9 @@
         "dates": {
             "today": "Today",
             "yesterday": "Yesterday"
+        },
+        "messages": {
+            "copied_to_clipboard": "Copied to clipboard"
         }
     },
     "common.confirm.delete": "Confirm delete",
--- a/src/js/angular/clustermanagement/directives/cluster-nodes-configuration.directive.js
+++ b/src/js/angular/clustermanagement/directives/cluster-nodes-configuration.directive.js
@@ -33,6 +33,13 @@ function ClusterNodesConfigurationComponent($translate, $timeout, productInfo, t
             $scope.addNode = () => {
                 $scope.addNewLocation = true;
                 $scope.newLocation = new Location();
+                // Timeout used because field is added dynamically on "Add Node" and needs to render before it can be focused
+                $timeout(function () {
+                    const input = document.querySelector("input[name='location']");
+                    if (input) {
+                        input.focus();
+                    }
+                }, 0);
             };
 
             /**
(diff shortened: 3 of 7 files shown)


src/css/clustermanagement.css
```diff
@@ -225,12 +225,27 @@ div.nodetooltip {
     -webkit-user-select: none;
     user-select: none;
     word-break: break-all;
-
     display: flex;
     align-items: center;
     justify-content: flex-start;
 }
 
+.location-item .text-only:hover {
```
svilenvelikov, line 233: Where the idea for this styling came from?

src/js/angular/clustermanagement/directives/cluster-nodes-configuration.directive.js
```diff
@@ -33,8 +33,47 @@ function ClusterNodesConfigurationComponent($translate, $timeout, productInfo, t
...
+                // Timeout used because field is added dynamically on "Add Node" and needs to render before it can be focused
+                $timeout(function () {
+                    const input = document.querySelector("input[name='location']");
+                    if (input) {
+                        input.focus();
+                    }
+                }, 0);
             };
 
+            $scope.copyToClipboard = (textToCopy, event) => {
```
svilenvelikov, line 45: Generally I like the idea on having the copy url button, but I have also some concerns about it. To begin with, this is not something that was requested. The problem with the current behavior was that the url was styled in a way that it wouldn't allow to be copied, so the easiest change was to find and fix this. 
image

The other concern that I have is that this implementation here duplicated our copy-to-clipboard.directive.js which we use on multiple places.
The third one is the url styling which has no anologue in the application as far as I can tell.

DesiBorisova: Like we discussed, I extended the existing directive with the custom logic. The original logic in the directive will remain as the default behavior. The new behavior can be observed in the Cluster view. The default can be verified for example in the Help section in the Import view.

If changes are OK, I will update the MR description.

src/js/angular/core/directives/copy-to-clipboard/copy-to-clipboard.directive.js
```diff
@@ -41,42 +59,100 @@ copyToClipboard.$inject = ['$translate', 'toastr'];
...
-                        toastr.success($translate.instant('import.help.messages.copied_to_clipboard'));
-                    }).catch((err) => {
-                        console.error('Could not copy text: ', err);
+                function showCustomTooltip() {
+                    const tooltip = document.createElement("span");
+                    tooltip.innerText = $translate.instant('common.copied');
+
+                    const iconElement = element[0].querySelector('.custom-link-icon');
+
+                    tooltip.style.fontFamily = "Arial, sans-serif";
```
svilenvelikov, line 111: I'd prefer the styling to be done in the stylesheet and not inline.

src/i18n/locale-en.json
```diff
@@ -2506,7 +2506,7 @@
     "common.show_all": "Show all",
     "common.hide": "Hide",
     "common.learn_more_in_documentation": "Learn more in the GraphDB documentation",
-    "common.copied": "Copied",
+    "common.copied": "IRI copied",
```
svilenvelikov, line 2509: Create a more specific key for the copy directive, e.g. `copy_to_clipboard.copied.tooltip`.

src/js/angular/core/directives/copy-to-clipboard/copy-to-clipboard.directive.js
```diff
@@ -41,42 +59,100 @@ copyToClipboard.$inject = ['$translate', 'toastr'];
...
                 const textToCopy = $scope.textToCopy ? $scope.textToCopy : element.parent().find('.copyable').text();
 
-                if (navigator.clipboard) {
-                    navigator.clipboard.writeText(textToCopy).then(() => {
-                        toastr.success($translate.instant('import.help.messages.copied_to_clipboard'));
-                    }).catch((err) => {
-                        console.error('Could not copy text: ', err);
+                function showCustomTooltip() {
+                    const tooltip = document.createElement("span");
+                    tooltip.innerText = $translate.instant('common.copied');
```
svilenvelikov, line 107: This is not OK, because this directive is generic and can be used for copying not only IRI's, but any text. So, better consider providing the tooltip key as parameter maybe

src/js/angular/core/directives/copy-to-clipboard/copy-to-clipboard.directive.js
```diff
@@ -41,42 +60,111 @@ copyToClipboard.$inject = ['$translate', 'toastr'];
...
+                        tooltip.classList.remove("show");
+                        setTimeout(() => tooltip.remove(), 200);
+                    }, 1500);
+                }
+
+                function showSuccessFeedback() {
+                    if ($scope.customTooltipStyle) {
+                        showCustomTooltip();
+                    } else {
+                        toastr.success($translate.instant('import.help.messages.copied_to_clipboard'));
```
svilenvelikov, line 152: This label should not be used here in this directive

src/i18n/locale-en.json
```diff
@@ -2505,6 +2506,7 @@
     "common.show_all": "Show all",
     "common.hide": "Hide",
     "common.learn_more_in_documentation": "Learn more in the GraphDB documentation",
+    "copy_to_clipboard.copied.tooltip": "IRI copied",
```
svilenvelikov, line 2509: This must be in the cluster_management section because it's used in the cluster management view

svilenvelikov (approved):

PR #1929, opened 2025-04-08
GDB-12117 Define HTTP interceptor structure

plamen-yordanov:
## What
Define a structure for HTTP interceptors

## Why
To have the ability to pre- and post-process HTTP requests

## How
Created an abstract class, which holds the contract methods of an HTTP interceptor. An interceptor can pre- and post-process a request. To do that, it needs to implement the `preProcess` and/or `postProcess` methods. For those methods to be considered, the interceptor must implement `shouldPreProcess` and `shouldPostProcess`, which will determine, what must be intercepted

Added `interceptor.service`, which chains all `preProcess` and `postProcess`

Interceptors have a priority (number). If an interceptor needs to be executed in a specific order, this can be done, by modifying that property. Interceptors with higher priority are executed first

Modified `http-service` to call the `preProcess` chain, before making the request, and `postProcess`, after receiving the response.

Created `auth-http-interceptor`, which adds security headers to each request. This hasn't been attached, however, since some security features are missing

## Testing
jest

## Screenshots
none, as the functionality is not active yet

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/interceptor/interceptors.ts
+++ b/packages/api/src/interceptor/interceptors.ts
@@ -1,0 +1,18 @@
+import {HttpInterceptor} from '../models/interceptor/http-interceptor';
+import {HttpRequest} from '../models/http/http-request';
+import {ModelList} from '../models/common';
+
+/**
+ * An array of HTTP request interceptors to be used in the application.
+ */
+export const REQUEST_INTERCEPTORS = new ModelList<HttpInterceptor<HttpRequest>>([
+  // Request interceptors go here
+  // new AuthRequestInterceptor()
+]);
+
+/**
+ * An array of HTTP response interceptors to be used in the application.
+ */
+export const RESPONSE_INTERCEPTORS = new ModelList<HttpInterceptor<Response>> ([
+  // Response interceptors go here
+]);
--- a/packages/api/src/models/interceptor/http-interceptor.ts
+++ b/packages/api/src/models/interceptor/http-interceptor.ts
@@ -1,0 +1,30 @@
+import {HttpRequest} from '../http/http-request';
+
+/**
+ * Defines the structure for an HTTP interceptor.
+ * This class allows for processing of HTTP requests or responses.
+ * A class extending this can override the `process` and `shouldProcess` methods to customize the HTTP
+ * request and response flow.
+ */
+export abstract class HttpInterceptor<T extends HttpRequest | Response> {
+  /** Priority of the interceptor. Higher values indicate earlier execution. */
+  priority = 0;
+
+  /**
+   * Processes an HTTP request or response.
+   * This method can be used to modify or enhance the request or response.
+   *
+   * @param data - The original HTTP request or response to be processed.
+   * @returns A promise that resolves to the processed HTTP request or response.
+   */
+  abstract process(data: T): Promise<T>
+
+  /**
+   * Determines whether the processing step should be applied to the given HTTP request or response.
+   * This method allows for conditional processing based on the request's or response's properties.
+   *
+   * @param data - The HTTP request or response to be evaluated.
+   * @returns A boolean indicating whether the processing should be applied (true) or skipped (false).
+   */
+  abstract shouldProcess(data: T): boolean;
+}
--- a/packages/api/src/services/http/http.service.ts
+++ b/packages/api/src/services/http/http.service.ts
@@ -1,9 +1,13 @@
 import {HttpOptions} from '../../models/http/http-options';
+import {InterceptorService} from '../interceptor/interceptor.service';
+import {HttpRequest} from '../../models/http/http-request';
+import {ServiceProvider} from '../../providers';
 
 /**
  * A service class for performing HTTP requests.
  */
 export class HttpService {
+  private readonly interceptorService = ServiceProvider.get(InterceptorService);
 
   /**
    * Performs an HTTP GET request.
@@ -93,15 +97,22 @@ export class HttpService {
   private request<T>(url: string, method: 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH', options: HttpOptions = {}): Promise<T> {
     const queryString = this.buildQueryParams(options.params);
     const fullUrl = `${url}${queryString ? `?${queryString}` : ''}`;
+    const headers = {
+      'Content-Type': 'application/json',
+      ...options.headers,
+    };
 
-    return fetch(fullUrl, {
-      method,
-      headers: {
-        'Content-Type': 'application/json',
-        ...options.headers,
-      },
-      body: options.body ? JSON.stringify(options.body) : null,
-    })
+    return this.interceptorService.preProcess(new HttpRequest({url: fullUrl, method, headers, body: options.body}))
+      .then((request) => {
+        return fetch(request.url, {
+          method: request.method,
+          headers: request.headers as HeadersInit,
+          body: request.body ? JSON.stringify(request.body) : null,
+        });
+      })
+      .then((response) => {
+        return this.interceptorService.postProcess(response);
+      })
       .then((response) => {
         if (!response.ok) {
           return Promise.reject(response);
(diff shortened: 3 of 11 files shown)


packages/api/src/models/interceptor/http-interceptor.ts
```diff
@@ -0,0 +1,52 @@
+import {HttpRequest} from '../http/http-request';
+
+/**
+ * Defines the structure for an HTTP interceptor.
+ * This class allows for pre-processing of HTTP requests and post-processing of HTTP responses.
+ * A class, extending this, can override the `preProcess` and `postProcess` methods to customize the HTTP
+ * request and response flow.
+ */
+export abstract class HttpInterceptor {
```
svilenvelikov, line 9: I think that we could simplify the API here. For example. HttpInterceptor can have just two methods: `shouldProcess` and `process` plus the priority. The reason for this is that you would need to think and implement  simpler interface and also the processing in the http service would be simplified as well. What purpose an interceptor will have would depend on how you register it. If you consider the stateful interceptor service, we could just have two lists for pre and post processors and methods for registering interceptors or if we keep you current approach, we could have pre and post interceptor arrays.

packages/api/src/interceptor/interceptors.ts
```diff
@@ -0,0 +1,9 @@
+import {HttpInterceptor} from '../models/interceptor/http-interceptor';
+
+/**
+ * An array of HTTP interceptors to be used in the application.
+ */
+export const INTERCEPTORS: HttpInterceptor[] = [
+  // Add interceptors here
```
svilenvelikov, line 7: It'd be good if you also write down a little explanation for the interceptors and their usage in the developer's guide.

packages/api/src/services/http/http.service.ts
```diff
@@ -93,15 +97,22 @@ export class HttpService {
...
-    return fetch(fullUrl, {
-      method,
-      headers: {
-        'Content-Type': 'application/json',
-        ...options.headers,
-      },
-      body: options.body ? JSON.stringify(options.body) : null,
-    })
+    return this.interceptorService.preProcess(new HttpRequest({url: fullUrl, method, headers, body: options.body}))
+      .then((request) => {
```
svilenvelikov, line 106: The only way I can skip the fetcher currently is if I throw error or reject. Also, if I do that, next interceptors would be skipped as well. Probably this is irrelevant at this point, but it's worth mentioning in the dev guide as well

svilenvelikov: Also, I don't see tests for such scenarios: interceptor is throwing or rejecting...

packages/api/src/models/interceptor/http-interceptor.ts
yordanalexandrov: If we go with the two lists/arrays approach we could name them `RequestInterceptors` and `ResponseInterceptors` or `Handlers`

packages/api/src/interceptor/auth/auth-http-interceptor.ts
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * This method performs the following tasks:
+   * 1. Adds an authorization token to the request headers if available.
+   * 2. Adds repository ID and location to the headers if not already present.
+   *
+   * @param request - The HTTP request to be processed.
+   * @returns A Promise that resolves to the modified HTTP request.
+   */
+  override preProcess(request: HttpRequest): Promise<HttpRequest> {
+    //TODO: when using OpenId this should be skipped
```
yordanalexandrov, line 29: This should go in the `shouldPreProcess`

packages/api/src/models/interceptor/http-interceptor.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  /**
+   * Determines whether the pre-processing step should be applied to the given HTTP request.
+   * This method allows for conditional pre-processing based on the request's properties.
+   *
+   * @param request - The HTTP request to be evaluated.
+   * @returns A boolean indicating whether the pre-processing should be applied (true) or skipped (false).
+   */
+  abstract shouldPreProcess(request: HttpRequest): boolean;
```
yordanalexandrov, line 31: I can see the benefit of this if the conditions are streight forward, buty ou might have different conditions for processing based on different properties/scenarios/service calls.

Most of the more complex interceptors I have worked with/on would put the logic in the `process` method and just return the `Promise.resolve(request)` if no proceessing is required.

packages/api/src/services/interceptor/interceptor.service.ts
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Chains all interceptors, which {@link HttpInterceptor#shouldPreProcess should pre-process} the request.
+   * @param request The initial HTTP request.
+   * @returns A promise that resolves to the final processed HTTP request.
+   */
+  async preProcess(request: HttpRequest): Promise<HttpRequest> {
+    let processedRequest = request;
+    for (const interceptor of this.interceptors) {
+      if (interceptor.shouldPreProcess(processedRequest)) {
+        processedRequest = await interceptor.preProcess(processedRequest);
```
yordanalexandrov, line 24: Just a thought, but I wouldn't it be better if we do not mutate the original request since the `HttpRequest` constructor accepts an `HttpRequest` we could simply instantiate a new object when passing it for processing.

Can't think of a case which would require the request to be immutable.

plamen-yordanov: A [new HttpRequest](https://github.com/Ontotext-AD/graphdb-workbench/pull/1929/files#diff-e2ebaa84b374fb300a44af05521f191826c2e28b7bdfce8fa20e48e39a386160R105) is being created specifically for the interceptor chain to handle

packages/api/src/models/interceptor/http-interceptor.ts
plamen-yordanov: Now, that I think about it, after the changes, there will be only 2 methods: `process` and `shouldProcess`. Having a default `promise.resolve` will no longer be needed, as there will be different interceptors for responses and requests. I'll remove the default `promise.resolve`, since there won't be a scenario, where it's needed

packages/api/src/interceptor/interceptors.ts
plamen-yordanov: Added explanation in the developers guide

packages/api/src/services/http/http.service.ts
plamen-yordanov: I could implement a mechanism for graceful cancel of the request if you like. For now I added tests for error and reject, and also mentioned the behavior in the developers guide

packages/api/src/interceptor/auth/auth-http-interceptor.ts
plamen-yordanov: moved to `preProcess`

packages/api/src/models/interceptor/http-interceptor.ts
plamen-yordanov: Removed the default `promise.resolve`. Now you will be forced to implement both `process` and `shouldProcess`

packages/api/src/services/http/http.service.ts
yordanalexandrov: I think this should be transparent. I don't think it would be a good idea to consume or rethrow any errors. Let the client deal with it.

plamen-yordanov: Maybe I didn't word myself correctly. I'm not referring to consuming rejections or errors, but implementing a mechanism, which gives the ability to cancel the request

plamen-yordanov:
Jenkins retry build

plamen-yordanov:
Jenkins retry build

plamen-yordanov:
Jenkins retry build

plamen-yordanov:
Jenkins retry build

PR #1933, opened 2025-04-09
(cherry-pick) GDB-11287 - Fix chart axis resetting when using chart filters

DesiBorisova:
## What
When filtering the Monitoring charts, the axis will show the correct ranges, without cutting off a part of the data.
The y-axis names will not overlap long tick labels.

## Why
No value was passed in for the multiplier and the wrong function parameter was used to calculate the y-axis max value. Hence, the y-axis max value would reset to 1 when the chart data was toggled.

## How
I made sure the function receives the multiplier value where it is expected.
I added a bigger gap between the axis name and the tick label.

## Testing
N/A

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/src/js/angular/resources/chart-models/performance/connections-chart.js
+++ b/src/js/angular/resources/chart-models/performance/connections-chart.js
@@ -46,7 +46,7 @@ export class ConnectionsChart extends ChartData {
     }
 
     updateRange(dataHolder, multiplier) {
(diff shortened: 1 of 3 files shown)


DesiBorisova:
Jenkins retry build

PR #1936, opened 2025-04-10
GDB-12078 - Update text snippet correctly after Edit

DesiBorisova:
## What
When editing and updating an imported `text snipped`, the changes will not be lost.

## Why
The update call and the import call sometimes happened before the list updates. This meant that the old file was used and any new changes the user made to the snippet would be lost.

## How
This was a little hard to reproduce, because the update and import actions had to be done very quickly. The solution was to add a list update call after the successful update.

## Testing
N/A

## Screenshots
N/A

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/src/js/angular/import/controllers/import-view.controller.js
+++ b/src/js/angular/import/controllers/import-view.controller.js
@@ -708,14 +708,20 @@ importViewModule.controller('UploadCtrl', ['$scope', 'toastr', '$controller', '$
                 }
                 importResource.data = data.text;
                 importResource.format = data.format;
-                updateTextImport(importResource);
+                updateTextImport(importResource)
+                    .then(() => $scope.updateListHttp(true))
+                    .then(() => {
+                        if (data.startImport) {
+                            $scope.setSettingsFor(importResource.name, false, importResource.format, Operation.IMPORT_SNIPPET);
+                        }
+                    })
             } else {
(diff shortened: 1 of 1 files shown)


src/js/angular/import/controllers/import-view.controller.js
```diff
@@ -781,7 +781,8 @@ importViewModule.controller('UploadCtrl', ['$scope', 'toastr', '$controller', '$
     const updateTextImport = (settings) => {
         $scope.updating = true;
         ImportRestService.updateTextSnippet($repositories.getActiveRepository(), settings).success(function (data) {
-            // its updated
+            // Update list after file updates
+            $scope.updateListHttp(true);
```
boyan-tonchev, line 785: This is not enough to fix the issue. You can simulate the problem by slowing down your internet connection when loading the page, then try editing the query quickly.
IMO, we should return only the promise from the ImportRestService.updateTextSnippet call.
``` javascript
const updateTextImport = (settings) => {
        $scope.updating = true;
        return ImportRestService.updateTextSnippet($repositories.getActiveRepository(), settings).success(function (data) {
```
and where the updateTextImport function is called to be somethink like
``` javascript
updateTextImport()
  .then(() => $scope.updateListHttp(true))
  .then(() => {
  // import with updated model
});
```

boyan-tonchev (approved):

PR #1948, opened 2025-04-15
GDB-10684: Cannot navigate away from the SPARQL editor page

boyan-tonchev:
## What
When the "SPARQL Query & Update" view is loaded and the user tries to navigate to another view, the navigation does not occur.

## Why
The "SPARQL Query & Update" view includes functionality that displays a confirmation dialog if there are running queries. If the user confirms, those queries are aborted. This functionality relies on methods from the ontotext-yasgui-web-component, which are asynchronous. Because of this, the original navigation event is prevented, and navigation is manually triggered using the $location service after user confirmation. However, this approach does not work in a single-spa environment—after the user confirms, the navigation is not performed.

## How
Replaced navigation using the $location service with the navigateToUrl method provided by single-spa.

## Testing
N/A

## Screenshots
N/A

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/sparql-editor/controllers.js
+++ b/packages/legacy-workbench/src/js/angular/sparql-editor/controllers.js
@@ -15,6 +15,7 @@ import {VIEW_SPARQL_EDITOR} from "../models/sparql/constants";
 import {CancelAbortingQuery} from "../models/sparql/cancel-aborting-query";
 import {QueryMode} from "../models/ontotext-yasgui/query-mode";
 import 'angular/core/services/event-emitter-service';
+import {navigateTo} from '@ontotext/workbench-api';
 
 const modules = [
     'ui.bootstrap',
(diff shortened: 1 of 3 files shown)


packages/legacy-workbench/src/js/angular/sparql-editor/controllers.js
```diff
@@ -528,11 +528,13 @@ function SparqlEditorCtrl($rootScope,
             .then(() => ontotextYasguiElement.abortAllRequests())
             .then(() => {
                 queriesAreCanceled = true;
-                $location.path(newPath).search(searches);
+                navigateTo(newUrl)();
             })
             .catch((error) => {
                 if (!(error instanceof CancelAbortingQuery)) {
-                    throw error;
+                    console.log(error)
```
svilenvelikov, line 535: console.error

svilenvelikov (approved):

packages/legacy-workbench/src/js/angular/sparql-editor/controllers.js
boyan-tonchev: Done

PR #1950, opened 2025-04-17
GDB-11314 Implement RDF suggestion selection

plamen-yordanov:
## What
Implement RDF search suggestion selection

## Why
To have the ability to view a resource in detail

## How
- Added loading of repository namespaces on repositoryId change. This is needed to be able to expand prefixes Added support for `application/sparql-results+json` content type in `http-service`, since this is what the namespaces request returns.
- Added event, when a suggestion is clicked. If the suggestion is a prefix it will be "expanded" into the input menu but not selected. If the suggestion is a complete result, the event will be thrown and notify listeners.
-  Added context property, since there could be more than one search component in one page. This way the event listener knows which event to capture. When the `rdf-search` component captures the event it will open the resource in a new tab. Depending on the value of the radio buttons, the resource can be opened in table view or graph visualization view.

## Testing
cypress/jest

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/api/src/models/repositories/namespace/namespace-map.ts
```diff
@@ -0,0 +1,22 @@
...
+  setNamespaces(namespaces: Record<string, string>): void {
+    this.namespaces = namespaces;
+  }
+
+  /**
+   * Retrieves the URI for a given prefix.
+   * @param prefix - The prefix for which to retrieve the URI.
+   */
+  getByPrefix(prefix: string): string {
+    return this.namespaces[prefix] || '';
```
svilenvelikov, line 20: I'm not sure that the empty string is OK as a default here

packages/api/src/services/repository/mappers/test/namespace-map.mapper.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+          {prefix: {value: 'yet-another'}, namespace: {value: 'http://yet-another.com/'}},
+          {prefix: {value: 'another'}, namespace: {value: 'http://another.com/'}}
+        ]
+      }
+    } as NamespacesResponse;
+
+    const expected: Record<string, string> = {
+      example: 'http://example.com/',
+      'yet-another': 'http://yet-another.com/',
+      another: 'http://another.com/'
```
svilenvelikov, line 25: No case for the prefix without uri

packages/api/src/services/http/http.service.ts
```diff
@@ -3,6 +3,8 @@ import {InterceptorService} from '../interceptor/interceptor.service';
 import {HttpRequest} from '../../models/http/http-request';
 import {ServiceProvider} from '../../providers';
 
+const JSON_CONTENT_TYPES = ['application/json', 'application/sparql-results+json'];
```
svilenvelikov, line 6: I think that we could have some more types here, but you have made it extensible, so it's OK if we test and add them later

packages/api/src/services/repository/rdf4j-repository-rest.service.ts
```diff
@@ -0,0 +1,19 @@
+import {HttpService} from '../http/http.service';
+import {NamespacesResponse} from '../../models/repositories/namespace/api/namespaces-response';
+
+/**
+ * Service for interacting with the RDF4J repository REST API.
+ */
+export class Rdf4jRepositoryRestService extends HttpService {
```
svilenvelikov, line 7: Maybe NamespaceRestService, because there are quite a lot of endpoints in the RDF4J API and I prefer to have them separated by type. And same for the services.

packages/shared-components/src/components/onto-rdf-search/onto-rdf-search.tsx
```diff
@@ -112,4 +120,15 @@ export class OntoRdfSearch {
       selected
     } as SearchButton);
   }
+
+  private onSuggestionSelected() {
+    this.subscriptions.add(
+      this.eventService.subscribe<SuggestionSelectedPayload>(SUGGESTION_SELECTED_EVENT, (payload) => {
+        if (payload.getContext() === this.RDF_CONTEXT) {
+          const fullUrl = `${this.redirectUrl}?uri=${encodeURIComponent(payload.getSuggestion().getValue())}`;
```
svilenvelikov, line 128: I'd prefer this to be in a service or some utility in the api maybe. Also I wonder the same about the redirectUrl's

packages/shared-components/src/components/onto-search-resource-input/onto-search-resource-input.tsx
```diff
@@ -58,6 +80,7 @@ export class OntoSearchResourceInput {
 
   componentWillLoad() {
     this.onAutocompleteEnabledChange();
+    this.onRepositoryIdChange();
```
svilenvelikov, line 83: These shared components should not be aware neither of the web applications stuff nor other shared components (unless they are somehow composed or extending themselves). So, better abstract this and move the responsibility to trigger the change to the component client

packages/shared-components/src/components/onto-search-resource-input/onto-search-resource-input.tsx
```diff
@@ -188,4 +208,45 @@ export class OntoSearchResourceInput {
         });
     }
   }
+
+  private onRepositoryIdChange() {
+    this.subscriptions.add(
+      this.repositoryContextService.onSelectedRepositoryIdChanged((repositoryId) => {
+        if (repositoryId) {
+          this.rdf4jRepositoryService.getNamespaces(repositoryId)
```
svilenvelikov, line 216: TBH I'd prefer this to come from outside as well, but let's leave it here for now

svilenvelikov (changes requested):

packages/api/src/services/repository/mappers/test/namespace-map.mapper.spec.ts
plamen-yordanov: I didn't know that was a valid scenario. I didn't see it, while testing and I'm not sure, what should happen in that case

packages/api/src/services/repository/rdf4j-repository-rest.service.ts
svilenvelikov: Yes

packages/api/src/services/http/http.service.ts
plamen-yordanov: added tests

packages/api/src/services/repository/rdf4j-repository-rest.service.ts
plamen-yordanov: renamed to `NamespacesService` and moved into its own folder

packages/shared-components/src/components/onto-rdf-search/onto-rdf-search.tsx
plamen-yordanov: Moved into `uri-util`

packages/shared-components/src/components/onto-search-resource-input/onto-search-resource-input.tsx
plamen-yordanov: Moved into `onto-rdf-search`

packages/shared-components/src/components/onto-search-resource-input/onto-search-resource-input.tsx
plamen-yordanov: The subscription is moved into the parent component and the namespaces are now a `Prop`

packages/shared-components/src/components/onto-rdf-search/onto-rdf-search.tsx
```diff
@@ -83,14 +96,14 @@ export class OntoRdfSearch {
         this.createSearchButton(
           'rdf_search.buttons.table',
           () => {
-            this.searchViewType = SearchViewType.TABLE;
+            this.redirectUrl = 'resource';
```
svilenvelikov, line 99: It's not OK that these urls are hardcodded here in a component somewhere in the middle of the component hierarchy.  Ideally it's better if these component are not coupled with such stuff. For now you could create some file with view url constants and reuse them. It's not perfect but it's better.

packages/shared-components/src/components/onto-search-resource-input/onto-search-resource-input.tsx
svilenvelikov: Well, this not a good idea too IMO. I think this should be moved even higher.

plamen-yordanov: I thought you meant the parent component, when you said `component client`. So the namespaces should be passed via context, if we are going higher?

plamen-yordanov: moved into `onto-header` and added a `namespaces-context.service` to pass the value. Added subscription in the search component and refactored tests to use namespaces, instead of repoId change

packages/shared-components/src/components/onto-rdf-search/onto-rdf-search.tsx
plamen-yordanov: Added them in `uri-utils`

svilenvelikov (approved):

plamen-yordanov:
Jenkins retry build

PR #1960, opened 2025-04-28
GDB-7219 - Allow edit of autocomplete labels

DesiBorisova:
## What
The labels in the Autocomplete page can be edited entirely.

## Why
Previously, only the languages could be edited.

## How
I made the IRI field editable and integrated the new endpoint for editing a label.

## Testing
Added test.

## Screenshots
N/A

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


src/js/angular/rest/autocomplete.rest.service.js
```diff
@@ -90,6 +92,24 @@ function AutocompleteRestService($http) {
         return $http.put(AUTOCOMPLETE_LABELS_ENDPOINT, label);
     }
 
+    function editLabelConfig(updatedLabel, originalLabel) {
```
svilenvelikov, line 95: One more thing I'd ask you to do is to add an AutocompleteService and and use that in controllers instead of the rest client directly. For now you could expose only the add/edit label methods there where you can mode this logic for assembling the params below.

svilenvelikov (approved):

yordanalexandrov (approved):

PR #1978, opened 2025-05-09
GDB-11314 Add preserve state to RDF search

plamen-yordanov:
## What
Added state preservation for the RDF search

## Why
To have the ability to preserve the input, selected element and view type, upon reload

## How
- Added new methods in the autocomplete-context.service to set and clear the input, view type and last selected element
- Added `preserveSearch` in `search-resource-input` to be able to configure the component to preserve the search
- Added `HtmlUtil#waitForElement` to monitor the dom for a given element, with a specified selector. The method returns a promise, which resolves, when the given element is inside the DOM. Currently, this is used, when loading the saved state and scrolling to a previously selected suggestion.
- Changed `onMouseEnter` to `onMouseMove`, since it was causing issues with arrow navigation, whenever the cursor is somewhere inside the suggestion list
- Replaced function calls with references in the view, for better performance

 ## Testing
 cypress/jest

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/api/src/services/autocomplete/autocomplete-storage.service.ts
```diff
@@ -1,10 +1,14 @@
 import {LocalStorageService} from '../storage';
+import {Suggestion} from '../../models/rdf-search';
 
 /**
  * Service for managing autocomplete-related storage operations.
  */
 export class AutocompleteStorageService extends LocalStorageService {
-  private readonly autocompleteEnabledKey = 'enabled';
+  private readonly enabledKey = 'enabled';
```
yordanalexandrov, line 8: Uppercase if constants like the `NAMESPACE`

packages/api/src/services/autocomplete/autocomplete-storage.service.ts
```diff
@@ -1,10 +1,14 @@
 import {LocalStorageService} from '../storage';
+import {Suggestion} from '../../models/rdf-search';
 
 /**
  * Service for managing autocomplete-related storage operations.
  */
 export class AutocompleteStorageService extends LocalStorageService {
```
yordanalexandrov, line 7: IMO it would make more sense, for the rdf search to have it's own `StorageService` and not get mixed up with the autocomplete

packages/api/src/services/autocomplete/autocomplete-storage.service.ts
plamen-yordanov: renamed to uppercase

packages/api/src/services/autocomplete/autocomplete-storage.service.ts
yordanalexandrov: Ok but everything added is conneceted to the `onto-search-resource-input`, not to the autocomplete although they are shown by using it.

Are the `INPUT_KEY` or the `LAST_SELECTED_KEY` used anywhere in the context of the Autocomplete and not in by `onto-search-resource-input`

plamen-yordanov: no, they don't appear to be used anywhere else

yordanalexandrov (dismissed):

PR #1995, opened 2025-05-13
GDB-12247 Fix navigation link not updating

plamen-yordanov:
## What
Fix navigation links not updating upon certain navigations

## Why
They are not updated, when a non-reloading navigation is triggered inside the app

## How
- Added a subscription to `navigation-end` inside `onto-navbar`. Whenever a navigation ends, the current route is taken from the URL. It is then used to match the corresponding menu item and select it. If the navigation is towards the home page, then all navbar links are closed and cleared.
- Added `getCurrentRoute` to get the current pathname with removed leading `/`. This is used to match the menu items via href

## Testing
existing

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/shared-components/src/components/onto-navbar/onto-navbar.tsx
+++ b/packages/shared-components/src/components/onto-navbar/onto-navbar.tsx
@@ -15,7 +15,13 @@ import {NavbarToggledEvent} from "./navbar-toggled-event";
 import {NavbarService} from "./navbar-service";
 import {NavbarItemModel, NavbarModel} from "./navbar-model";
 import {TranslationService} from '../../services/translation.service';
-import {navigateTo} from '@ontotext/workbench-api';
+import {
+  EventName,
+  EventService, getCurrentRoute, isHomePage,
+  navigateTo,
+  ServiceProvider,
+  SubscriptionList
+} from '@ontotext/workbench-api';
 
 const labelKeys = {
   EXPAND: 'menu.buttons.expand',
@@ -34,6 +40,8 @@ export class OntoNavbar {
     [labelKeys.COLLAPSE]:TranslationService.translate(labelKeys.COLLAPSE),
     [labelKeys.LOGO_LINK]: TranslationService.translate(labelKeys.LOGO_LINK)
   }
+
+  private readonly subscriptions = new SubscriptionList();
 
   /**
    * Flag indicating whether the navbar is collapsed in result of toggle action initiated by the user. This is needed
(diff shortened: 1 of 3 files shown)


packages/shared-components/src/components/onto-navbar/onto-navbar.tsx
```diff
@@ -131,7 +139,25 @@ export class OntoNavbar {
...
+        EventName.NAVIGATION_END, () => {
+          this.selectItemByUrl();
+        }
+      )
+    );
+  }
+
+  private selectItemByUrl() {
+    this.menuModel.deselectAll();
+    isHomePage()
```
svilenvelikov, line 157: Doesn't navigation to the home page trigger full page reload? Is this really necessary?

plamen-yordanov: navigation to the home page uses `singleSpa.navigateToUrl`, which does not do a full page reload

svilenvelikov: Why? This wasn't how it was done before. I think that we spoke about this some time ago but I can't recall what was the reason

packages/shared-components/src/components/onto-navbar/onto-navbar.tsx
```diff
@@ -131,7 +139,25 @@ export class OntoNavbar {
...
 
-    this.refreshNavbar()
+    this.refreshNavbar();
+  }
+
+  private subscribeToNavigationEnd() {
+    this.subscriptions.add(
+      ServiceProvider.get(EventService).subscribe(
+        EventName.NAVIGATION_END, () => {
+          this.selectItemByUrl();
```
boyan-tonchev, line 149: After these changes, the selected menu item is updated only after navigation has completed. Please revisit the method [select(event: MouseEvent, menuItem: NavbarItemModel)](https://github.com/Ontotext-AD/graphdb-workbench/pull/1995/files#diff-d4cd33d1f899ad89850acbd4264984373b3f7a3181397e1e506329c32af5aaa6L100). There is logic in this method that selects the clicked item immediately, which is now unnecessary and is causing issues.

In some pages, there is logic that prevents navigation when clicking on a menu item, but the item still becomes selected, even though the navigation was blocked. For example, when a long-running SPARQL query is running and the user tries to navigate away, a confirmation dialog appears. If the user clicks "Cancel", the wrong menu item is selected.

plamen-yordanov: Removed `this.menuModel.deselectAll()` and `this.menuModel.selectItem(menuItem)`, when clicking a menu item, which doesn't have children. Now such an item will be selected only after successful navigation. If the menu item has children, then they are just opened and closed, so no change is needed there

yordanalexandrov (approved):

boyan-tonchev (dismissed):

boyan-tonchev (approved):

PR #1999, opened 2025-05-14
GDB-12253 Fix repository info covering RDF search button

yordanalexandrov:
## What
Fixed a problem with the repository info popover of the repository selector covering the RDF search button.

## Why
As it is very easy to trigger the info popover (on mouseenter) it becomes annoying when the user wants to use the search box but opens the info. He is forced to somehow navigate out of it and then precisely navigate to the RDF search without triggering it again.

## How
- defined a new type `HTMLElementWithTooltip`, which is a union between a `HTMLElement` with an added property `hideTooltip` which is a function that hides the tooltip. This type is basically an HTMLElement which has the `OntoTooltip` component set on it.
- set `hideTooltip` to a function which hides the tooltip instance
- exposed the tooltip placement property in `OntoDropdown` so user can set where the tooltip is shown. It accepts a string or a function to which the `isOpen` argument is passed
- modified info popover positioning in `OntoRepositorySelector` to display below the button when closed and to the left of dropdown items when open, improving visibility and preventing UI overlap

## Testing
added test

## Screenshots
image
image
image

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [X] Tests


packages/shared-components/cypress/e2e/tooltip/tooltip.cy.js
```diff
@@ -65,4 +65,26 @@ describe('Tooltip', () => {
     // Then, I expect the tooltip to be removed along with the element
     TooltipSteps.getTooltip().should('not.exist');
   });
+
+
+  it.only('Should remove tooltip from the DOM, when hideTooltip is called on element with tooltip', () => {
```
svilenvelikov, line 70: only

packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
```diff
@@ -65,6 +66,14 @@ export class OntoDropdown {
    */
   @Prop() tooltipTheme: string;
 
+  /**
+   * Specifies the items tooltip placement. Accepts a string of the placement or a function that returns the placement.
+   * The function takes the isOpen parameter as a boolean and returns the placement as a string.
+   * If not provided, the tooltip will be placed to the left.
+   *
+   */
+  @Prop() tooltipPlacement: string | ((isOpen: boolean) => void) = 'left';
```
svilenvelikov, line 75: Can we have some constant with strings union maybe for the placement variations we are going to support?

packages/shared-components/src/components/onto-repository-selector/onto-repository-selector.tsx
```diff
@@ -84,13 +84,21 @@ export class OntoRepositorySelector {
...
           tooltipTheme='light-border'
           items={this.items}>
         </onto-dropdown>
       </Host>
     );
   }
 
+  private tooltipAlignment(isOpen: boolean) {
+    if (isOpen) {
+      return 'left';
```
svilenvelikov, line 97: Constants for these as well

packages/shared-components/cypress/e2e/tooltip/tooltip.cy.js
yordanalexandrov: removed

packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
yordanalexandrov: Added `OntoTooltipPlacement`

packages/shared-components/src/components/onto-repository-selector/onto-repository-selector.tsx
yordanalexandrov: Added `OntoTooltipPlacement`

svilenvelikov (dismissed):

boyan-tonchev (approved):

PR #2000, opened 2025-05-14
fix(GDB-12315): Fix language selector not updating after route change

teodossidossev:
## What
Fixes an issue where the language selector does not reflect the selected language after navigating to a new route.

## Why
Previously, the language selector’s internal state was initialized only in the constructor, which is not re-invoked on route changes. As a result, when the language was changed post-routing, the selector did not update

## How
- Moved initialization logic from the constructor to connectedCallback() in both TranslateLabel and OntoLanguageSelector components to ensure proper setup on each mount.
- Replaced redundant service retrieval with readonly properties directly initialized from ServiceProvider.
- Ensured fallback to default language when no language is selected.

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [ ] Tests

### DIFF
--- a/packages/shared-components/src/components/onto-language-selector/onto-language-selector.tsx
+++ b/packages/shared-components/src/components/onto-language-selector/onto-language-selector.tsx
@@ -4,7 +4,6 @@ import {
   ServiceProvider,
   LanguageService,
   LanguageContextService,
-  LanguageStorageService
 } from "@ontotext/workbench-api";
 import {DropdownItemAlignment} from '../../models/dropdown/dropdown-item-alignment';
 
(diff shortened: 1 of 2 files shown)


packages/shared-components/src/components/onto-language-selector/onto-language-selector.tsx
```diff
@@ -30,11 +29,7 @@ export class OntoLanguageSelector {
    */
   @State() currentLanguage: string;
 
-  constructor() {
-    this.languageService = ServiceProvider.get(LanguageService);
-    this.languageContextService = ServiceProvider.get(LanguageContextService);
-    const selectedLanguage = ServiceProvider.get(LanguageStorageService).get(this.languageContextService.SELECTED_LANGUAGE);
-    this.changeLanguage(selectedLanguage?.getValueOrDefault(this.languageService.getDefaultLanguage()));
```
plamen-yordanov, line 37: shouldn't the initial call to `this.changeLanguage` remain?

packages/shared-components/src/components.d.ts
```diff
@@ -714,6 +734,26 @@ declare namespace LocalJSX {
     interface OntoRdfSearch {
     }
     interface OntoRepositorySelector {
+        /**
+          * Determines whether the current user has write access to the repository.
+         */
+        "canWriteRepo"?: (repo: Repository) => boolean;
```
plamen-yordanov, line 740: These probably belong in a different MR

yordanalexandrov (approved):

PR #2008, opened 2025-05-15
GDB-10490: Move common stylesheets from legacy-workbench to shell application so that they can be reused

boyan-tonchev:
## What
- Fixed application of legacy Workbench styling;
- Fixed visual issues with the main menu.

## Why
The visual appearance was broken during the migration process due to missing styles.

## How
- The import of bootstrap.min.css was accidentally removed during the migration;
- Common styles (for icons, basic button and link formatting, status messages like success/error, etc.) have been moved to the root config package. This ensures that core styles are loaded consistently and shared across all modules.

## Testing
N/A

## Screenshots
image

image

image

image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/root-config/src/ontotext-root-config.js
```diff
@@ -11,11 +11,14 @@ import {
   constructLayoutEngine,
 } from 'single-spa-layout';
 import microfrontendLayout from './microfrontend-layout.html';
-import './styles/bootstrap.min.css';
+import './styles/css/common.css';
```
plamen-yordanov, line 14: Can you move these in the existing `scss` structure, or maybe in `onto-vendor.js`, instead of importing them individually here?

boyan-tonchev: Moved to onto-vendor.js

svilenvelikov (approved):

PR #2009, opened 2025-05-15
GDB-12322 Fix external link navigation

plamen-yordanov:
## What
Fix error, when clicking external link in the navbar menu

## Why
It was throwing an error and not navigating.

## How
Added `documentationHref` property to the plugin items, which redirects to external resources. In the legacy workbench, a hrefFun is used, but that requires each menu item to pass a resolver and product info. This way, the plugin item is decoupled form that responsibility. Added `resolveDocumentationUrl` in `uri-util`, which is used to resolve the complete url, when an element has `documentationHref` present

## Testing
manual

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


yordanalexandrov (approved):

packages/api/src/services/utils/uri-util.ts
```diff
@@ -118,6 +122,33 @@ export class UriUtil {
...
+   * @example
+   * const docUrl = UriUtil.resolveDocumentationUrl('10.0.0', 'sparql-endpoint');
+   * // Returns: 'https://graphdb.ontotext.com/documentation/10.0.0/sparql-endpoint'
+   */
+  static resolveDocumentationUrl(productVersion: string, endpointPath: string): string {
+    if (!productVersion || !endpointPath) {
+      throw new Error('Product version and endpoint path are required for documentation URL resolution.');
+    }
+
+    const isUnofficialVersion = productVersion.includes('-');
```
svilenvelikov, line 147: Unofficial versions might be formatted without hyphen, so IMO, a better approach would be to match with a regex validating that the product version is a string containing only a valid semantic versioning number.

svilenvelikov: You may leave it as is, that's why I approved the MR

PR #2019, opened 2025-05-19
GDB-12338 Fix incorrect legacy autocomplete status on startup

plamen-yordanov:
## What
Fix incorrect value of the autocomplete `enabled` status on application startup

## Why
It wasn't set correctly. It's always false on startup, regardless if the selected repository has autocomplete enabled

## How
Made `updateAutocompleteStatus` to be called regardless, if the newly selected repository is the same as the `WorkbenchContextService` repository. The problem is that now, the `onRepositoriesChanged` method is called within a subscription to repositoryIdChange (the migrated context). The new repository context service emits twice on startup (first with undefined and a second time with the actual value).
 The first problem is that the first call to `onRepositoriesChanged` has different values for new and old repositories, which sets the context to the new value. When new and old value are the same `updateAutocompleteStatus` is not called.
 The second part of the problem is that at the time of the first 2 emits of the context, repositories and license data is not yet loaded, which causes conditions in `updateAutocompleteStatus` to return true and skip loading and updating the autocomplete status.
 Then `onRepositoriesChanged` is called for a third time. This time from the regular legacy workbench flow. At that point the missing data is loaded, but new and old repositories return the same value, which doesn't call `updateAutocompleteStatus`

Now `updateAutocompleteStatus` is called always, regardless if new and old value are the same. During the initial two calls of it, there are no repositories and license, which protects from extra calls to `/enabled`. The third call now works as before and correctly sets the `enabled value in the local store

 ## Testing
 manual

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/root-config/src/bootstrap/autocomplete/autocomplete.js
+++ b/packages/root-config/src/bootstrap/autocomplete/autocomplete.js
@@ -1,4 +1,10 @@
-import {ServiceProvider, AutocompleteService, AutocompleteContextService, RepositoryStorageService} from '@ontotext/workbench-api';
+import {
+  ServiceProvider,
+  AutocompleteService,
+  AutocompleteContextService,
+  RepositoryStorageService,
+  RepositoryContextService
+} from '@ontotext/workbench-api';
 
 /**
  * Check if autocomplete is enabled, when loading(reloading) the application.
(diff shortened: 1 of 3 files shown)


boyan-tonchev (approved):

packages/root-config/src/bootstrap/autocomplete/autocomplete.js
```diff
@@ -6,7 +6,7 @@ import {ServiceProvider, AutocompleteService, AutocompleteContextService, Reposi
  * If there is no selected repository, the request will not be made.
  */
 const isAutocompleteEnabled = () => {
-  const currentRepository = ServiceProvider.get(RepositoryStorageService).get('selecterRepositoryId').getValue();
+  const currentRepository = ServiceProvider.get(RepositoryStorageService).get('selectedRepositoryId').getValue();
```
yordanalexandrov, line 9: Please change to `RepositoryContextService.SELECTED_REPOSITORY_ID` to mitigate further errors like this

yordanalexandrov: I see this pattern used in other places

yordanalexandrov (approved):

PR #2032, opened 2025-05-22
GDB-12362: Old local storage keys for repository and location are not updated across repository-related actions

boyan-tonchev:
## What
Removed usages of the RepositoryStorage methods.

## Why
RepositoryStorage is an outdated service used for reading and writing repository and repository location data. It should no longer be used in favor of the newer implementation.

## How
Replaced usages of the old RepositoryStorage service with the new RepositoryStorageService.

## Testing
N/A

## Screenshots
N/A

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/legacy-workbench/src/js/angular/core/services/jwt-auth.service.js
```diff
@@ -93,6 +91,15 @@ angular.module('graphdb.framework.core.services.jwtauth', [
 
             const that = this;
 
+            const getActiveRepositoryObjectFromStorage = function() {
```
yordanalexandrov, line 94: There is a method like this in the `RepositoriesService`, is it not possible to be reused?

packages/legacy-workbench/src/js/angular/core/services/jwt-auth.service.js
```diff
@@ -456,14 +459,14 @@ angular.module('graphdb.framework.core.services.jwtauth', [
                 }
 
                 // Get the selected repository's ID from the current context.
-                const repoId = RepositoryStorage.getActiveRepository();
+                const repo = getActiveRepositoryObjectFromStorage
```
yordanalexandrov, line 462: missing parentheses

packages/legacy-workbench/src/js/angular/core/services/jwt-auth.service.js
boyan-tonchev: I didn't try it because there's an injection of $jwtAuth, and I think it will cause a cyclic injection — but I’ll try it.

packages/legacy-workbench/src/js/angular/controllers.js
```diff
@@ -124,13 +122,19 @@ function homeCtrl($scope,
     const subscriptions = [];
 
     const onSelectedRepositoryIdUpdated = (repositoryId) => {
+        let selectedRepositoryId = repositoryId;
+        if (!selectedRepositoryId) {
+            const repositoryStorageService = ServiceProvider.get(RepositoryStorageService);
```
plamen-yordanov, line 127: `$repositories.getActiveRepository()` has almost the same implementation. The difference is it will default to `undefined`. Can we use it here?

packages/legacy-workbench/src/js/angular/core/services/jwt-auth.service.js
boyan-tonchev: Done

plamen-yordanov (approved):

yordanalexandrov (approved):

PR #2034, opened 2025-05-22
GDB-12173 Add initial state tests for GraphQL views

plamen-yordanov:
## What
Add initial state tests for GraphQL views

## Why
To ensure the views start up correctly with and without selected repository and with configured endpoints

## How
Added spec files with and without selected repository, as well as with configured endpoints, which assert the initial state of both views

## Testing
cypress

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


boyan-tonchev (dismissed):

boyan-tonchev (approved):

plamen-yordanov:
> Why do we have each in a separate file? Is this how we have agreed to do it? Because some can be in one spec in different `describe` sections eg. with/without repo

This is how we have agreed to do it, yes. Not in describes, but separate files

PR #2056, opened 2025-05-30
feat(GDB-12347): integrate context updates and AngularJS change detection

teodossidossev:
## WHAT
- Introduced LifecycleHooks interface with onCreated() method.
- Centralized context service registration via ContextServiceRegistry.
- Enabled global context subscriptions using subscribeToAllRegisteredContexts.
- Ensured afterChangeCallback is invoked after main callback.
- Hooked AngularJS $rootScope.$apply() to context changes during mount.
- Unsubscribed automatically on unmount to prevent memory leaks.

## WHY
- Eliminate tight coupling and manual registration of context services.
- Support reactive cross-cutting logic through lifecycle hooks.
- Trigger AngularJS digest cycles on external context changes.

## HOW
- Added LifecycleHooks and integrated into ServiceProvider.get.
- Context services self-register via onCreated() into the registry.
- Replaced legacy static subscription utility with registry-based logic.
- Mounted subscribeToAllRegisteredContexts inside single-spa-angularjs and stored unsub fn.
- Updated and extended tests

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [X] Tests

### DIFF
--- a/packages/api/src/providers/service/lifecycle-hooks.ts
+++ b/packages/api/src/providers/service/lifecycle-hooks.ts
@@ -1,0 +1,30 @@
+/**
+ * Interface for defining lifecycle hooks for {@link Service} instances.
+ *
+ * Classes implementing this interface can hook into specific lifecycle moments,
+ * such as when the instance is created by the {@link ServiceProvider}.
+
+ */
+export interface LifecycleHooks {
+  /**
+   * Hook called immediately after the service instance is created and before it's cached.
+   * Useful for registration or custom setup that requires the fully constructed object.
+   *
+   * Services may use the `onCreated` method to perform additional registrations,
+   * for example registering themselves in context or event registries.
+   *
+   * @example
+   * ```ts
+   * import { ServiceProvider } from '../service.provider';
+   * import { ContextSubscriptionManager } from '../../services/context/context-service-registry';
+   *
+   * class MyService implements Service, LifecycleHooks {
+   *   onCreated(): void {
+   *     // Perform some task when fully created
+   *     foo();
+   *   }
+   * }
+   * ```
+   */
+  onCreated?(): void;
+}
--- a/packages/api/src/providers/service/service.provider.ts
+++ b/packages/api/src/providers/service/service.provider.ts
@@ -1,4 +1,5 @@
 import {Service} from './service';
+import {LifecycleHooks} from './lifecycle-hooks';
 
 /**
  * Service provider for all {@link Service} instances. Services in the API are singletons, meaning that there is only
@@ -13,8 +14,11 @@ export class ServiceProvider {
   private static readonly SERVICE_INSTANCES = new Map<string, Service>;
 
   /**
-   * Returns the instance of the given service type. If the service has not been created yet, it will be created and
-   * cached.
+   * Returns the instance of the given service type. If the service has not been created yet, this method:
+   *   1. Instantiates the service via `new type()`.
+   *   2. Calls its `onCreated()` hook if implemented.
+   *   3. Caches and returns the instance.
+   *
    * @param type The service type to retrieve.
    * @returns The instance of the service.
    * @template T The type of the service to retrieve.
@@ -23,9 +27,16 @@ export class ServiceProvider {
   public static get<T extends Service>(type: { new (): T }): T {
     if (!ServiceProvider.SERVICE_INSTANCES.has(type.name)) {
       const instance = new type();
+      if (ServiceProvider.implementsLifecycleHooks(instance)) {
+        instance.onCreated?.();
+      }
       ServiceProvider.SERVICE_INSTANCES.set(type.name, instance);
     }
     return this.SERVICE_INSTANCES.get(type.name) as T;
+  }
+
+  private static implementsLifecycleHooks(instance: Service): instance is LifecycleHooks {
+    return typeof instance === 'object' && typeof (instance as LifecycleHooks).onCreated === 'function';
   }
 
   /**
--- a/packages/api/src/services/context/context.service.ts
+++ b/packages/api/src/services/context/context.service.ts
@@ -2,6 +2,10 @@ import {ValueContext} from '../../models/context/value-context';
 import {ValueChangeCallback} from '../../models/context/value-change-callback';
 import {Service} from '../../providers/service/service';
 import {BeforeChangeValidationPromise} from '../../models/context/before-change-validation-promise';
+import {ContextSubscriptionManager} from './context-subscription-manager';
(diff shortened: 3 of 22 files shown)


packages/api/src/providers/service/lifecycle-hooks.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Interface for defining lifecycle hooks for {@link Service} instances.
+ *
+ * Classes implementing this interface can hook into specific lifecycle moments,
+ * such as when the instance is created by the {@link ServiceProvider}.
+ *
+ * Services may use the `onCreated` method to perform additional registrations,
```
svilenvelikov, line 7: Lines from here to the end is better to be moved to the onCreated doc

packages/api/src/services/context/context-service-registry.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+type GlobalSubscription = {
+  callback: ValueChangeCallback<unknown>;
+  beforeChangeValidationPromise?: BeforeChangeValidationPromise<unknown>;
+  afterChangeCallback?: ValueChangeCallback<unknown>;
+};
+
+/**
+ * ContextServiceRegistry is responsible for tracking all registered ContextService instances in the application.
+ * It supports global subscription to context changes and notifies newly registered services of existing global listeners.
```
svilenvelikov, line 18: I don't quite understand this doc. Tracking services? For what? And what is a global subscription?

packages/api/src/services/context/context-service-registry.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+export class ContextServiceRegistry implements Service {
+  private readonly services: ContextService<Record<string, unknown>>[] = [];
+  private readonly subscribers: GlobalSubscription[] = [];
+
+  registerContextService(service: ContextService<Record<string, unknown>>): void {
```
svilenvelikov, line 24: Public methods must be documented

packages/api/src/services/context/context-service-registry.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.services.push(service);
```
svilenvelikov, line 25: I don't like this. We already keep services in the service provider. Why do we need to keep the here as well?!

packages/api/src/providers/service/service.provider.ts
```diff
@@ -23,11 +27,18 @@ export class ServiceProvider {
   public static get<T extends Service>(type: { new (): T }): T {
     if (!ServiceProvider.SERVICE_INSTANCES.has(type.name)) {
       const instance = new type();
+      if (ServiceProvider.implementsLifecycleHooks(instance)) {
```
plamen-yordanov, line 30: if we move`onCreated` inside the `Service` interface, we should be able to just `instance?.onCreated()` I think

packages/api/src/services/context/context.service.ts
```diff
@@ -67,18 +76,51 @@ export abstract class ContextService<TFields extends Record<string, unknown>> im
...
+   * Subscribes globally to all fields defined in TFields.
+   */
+  public subscribeAll<T>(
+    callbackFunction: ValueChangeCallback<T | undefined>,
+    beforeChangeValidationPromise?: BeforeChangeValidationPromise<T>,
+    afterChangeCallback?: ValueChangeCallback<T | undefined>
+  ): () => void {
+    const unsubscribeFns: (() => void)[] = [];
+    // iterate through service-defined fields
+    for (const key of Object.values(this as unknown as TFields)) {
```
plamen-yordanov, line 111: this should iterate the keys of `this.context`. Currently, it will add subscribe to any string property of the class. `if (typeof key === 'string')` should not be needed after this as well

packages/api/src/services/context/context-service-registry.ts
teodossidossev: renamed and documented

packages/api/src/services/context/context-service-registry.ts
teodossidossev: fixed

packages/api/src/services/context/context-service-registry.ts
```diff
@@ -0,0 +1,84 @@
...
+   * @param beforeChangeValidationPromise - Optional validation function called before value is applied.
+   * @param afterChangeCallback - Optional function called after value is updated.
+   * @returns A function that unsubscribes from all context changes.
+   */
+  subscribeToAllRegisteredContexts(
+    callback: ValueChangeCallback<unknown>,
+    beforeChangeValidationPromise?: BeforeChangeValidationPromise<unknown>,
+    afterChangeCallback?: ValueChangeCallback<unknown>
+  ): () => void {
+    const unsubFns: (() => void)[] = [];
```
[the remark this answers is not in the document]
teodossidossev: done

packages/api/src/services/context/context.service.ts
```diff
@@ -67,18 +76,51 @@ export abstract class ContextService<TFields extends Record<string, unknown>> im
...
+
+  /**
+   * Subscribes globally to all fields defined in TFields.
+   */
+  public subscribeAll<T>(
+    callbackFunction: ValueChangeCallback<T | undefined>,
+    beforeChangeValidationPromise?: BeforeChangeValidationPromise<T>,
+    afterChangeCallback?: ValueChangeCallback<T | undefined>
+  ): () => void {
+    const unsubscribeFns: (() => void)[] = [];
```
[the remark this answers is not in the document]
teodossidossev: done

packages/api/src/services/context/context.service.ts
teodossidossev: fixed

packages/api/src/providers/service/service.provider.ts
```diff
@@ -23,11 +27,18 @@ export class ServiceProvider {
...
+        instance.onCreated?.();
+      }
       ServiceProvider.SERVICE_INSTANCES.set(type.name, instance);
     }
     return this.SERVICE_INSTANCES.get(type.name) as T;
   }
 
+  private static implementsLifecycleHooks(instance: unknown): instance is LifecycleHooks {
```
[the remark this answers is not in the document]
teodossidossev: fixed

packages/api/src/providers/service/service.provider.ts
teodossidossev: this will break the Interface Segregation Principle (solId)

PR #2057, opened 2025-05-30
GDB-12388 Fix repository dropdown text not changing

plamen-yordanov:
## What
Fixed the repository dropdown text, which didn't change, depending on the availability/absence of repositories

## Why
The text should be different, depending on whether there are available repositories or not

## How
Added another `@State` label for `No accessible repositories` and added a condition to switch labels based on the presence/absence of repositories

## Testing
cypress

## Screenshots
Screenshot from 2025-05-30 10-44-41
Screenshot from 2025-05-30 10-45-04

## Checklist

- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


svilenvelikov (approved):

plamen-yordanov:
Changed `home-page-without-repositories.spec` to assert `No accessible repositories` instead of `Choose repository`

PR #2063, opened 2025-06-02
GDB-12361 Fix missing saved repository on startup

plamen-yordanov:
## What
Fix missing saved repository on startup

## Why
Upon bootstrap, a request for autocomplete is made, if there is a saved repository in the local store. That repository might have been deleted at the time of startup. The local store shows a saved repository, which prompts the autocomplete-bootstrap to make a request to `/enabled`. Since the repository doesn't exist, the backend throws an error.

## How
Moved the repository bootstrap functions ahead of all other bootstrap functions. Whenever they are resolved, the rest of the bootstrap functions will be triggered. This way we first get the list of existing repositories and then check to see if the last saved repository still exists.

## Testing
n/a

## Screenshots
n/a

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/root-config/src/bootstrap/autocomplete/autocomplete.js
+++ b/packages/root-config/src/bootstrap/autocomplete/autocomplete.js
@@ -2,7 +2,7 @@ import {
   ServiceProvider,
   AutocompleteService,
   AutocompleteContextService,
-  RepositoryStorageService
+  RepositoryContextService
 } from '@ontotext/workbench-api';
 
 /**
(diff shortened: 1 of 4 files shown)


packages/root-config/src/bootstrap/bootstrap.js
```diff
@@ -1,15 +1,17 @@
...
-import {repositoryBootstrap} from './repository/repository-bootstrap';
 import {autoCompleteBootstrap} from './autocomplete/autocomplete';
 import {securityBootstrap} from './security/security-bootstrap';
 
 export const bootstrapPromises = [
   ...languageBootstrap,
   ...licenseBootstrap,
   ...productInfoBootstrap,
-  ...repositoryBootstrap,
   ...autoCompleteBootstrap,
```
boyan-tonchev, line 11: I added a TODO in autocomplete.js. Could you consider changing the flow of how we update the autocomplete status for the selected repository?
I think it would be better to subscribe to the repository change event, and once it's triggered, update the autocomplete flag based on the newly selected repository.

On bootstrap, we retrieve the selected repository from local storage and update the repository context with it.
If we add a subscription, we can check the autocomplete status using the correct repository ID and location.

plamen-yordanov: Refactored `autocomplete.js` to update the `enabled` status based on repository change

packages/root-config/src/bootstrap/repository/repository-bootstrap.js
```diff
@@ -16,11 +16,13 @@ const loadRepositories = () => {
       repositoryContextService.updateRepositoryList(repositories);
 
       // Initializing the repository context service with the selected repository.
-      const repositoryReference = ServiceProvider.get(RepositoryStorageService).getRepositoryReference();
-      repositoryContextService.updateSelectedRepository(repositoryReference);
+      const lastSavedRepo = ServiceProvider.get(RepositoryStorageService).getRepositoryReference();
+      // The last saved repository could have been deleted, so we check if it still exists in the list.
+      const existingRepo = repositories.find((repo) => repo.id === lastSavedRepo.id);
```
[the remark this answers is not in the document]
plamen-yordanov: reverted

packages/root-config/src/bootstrap/autocomplete/autocomplete.js
```diff
@@ -2,25 +2,27 @@ import {
...
-  if (!currentRepository.id) {
-    return Promise.resolve();
-  }
-  return ServiceProvider.get(AutocompleteService).isAutocompleteEnabled()
-    .then((enabled) => {
-      ServiceProvider.get(AutocompleteContextService).updateAutocompleteEnabled(enabled);
+  return new Promise((resolve, reject) => {
+    ServiceProvider.get(RepositoryContextService).onSelectedRepositoryChanged((selectedRepository) => {
+      if (!selectedRepository?.id) {
+        return resolve();
```
boyan-tonchev, line 17: I'm not sure, but should we set autocomplete to false when no repository is selected?

PR #2066, opened 2025-06-03
GDB-12375: Page state does not reload after repository change

boyan-tonchev:
## What
The page state does not reload after a repository change.

## Why
After switching to the new repository selector component, the repositoryIsSet event stopped being propagated. However, some pages rely on this event to reload their internal state according to the newly selected repository.

## How
Added functionality in mainCtrl to fire the "repositoryIsSet" event when the repository is changed. This resolves issues on pages that rely on subscribing to this event to update their state.

### Additional work
Modified the repository persistence logic to store only the id and location of the selected repository.

## Testing
N/A

## Screenshots
N/A

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/api/src/services/repository/repository-storage.service.ts
```diff
@@ -54,8 +54,11 @@ export class RepositoryStorageService extends LocalStorageService {
     this.remove(this.SELECTED_REPOSITORY);
   }
 
-  private serializeRepositoryReference(repositoryReference: RepositoryReference): string {
-    return JSON.stringify(repositoryReference);
+  private serializeRepositoryReference(repositoryReference: RepositoryReference): string | null {
+    if (repositoryReference) {
+      return `{"id": "${repositoryReference.id}", "location": "${repositoryReference.location}"}`;
```
svilenvelikov, line 59: Two things here:
* Assembling JSON as a string like this is a bad practive. It's not typesafe.
* As far as I can see, the RepositoryReference interface declares only `id` and `location` fields. Why this logic was even needed. If the old code `JSON.stringify(repositoryReference)`  doesn't work here, then the problem is somewhere else most likely.

boyan-tonchev: Fixed

svilenvelikov (approved):

PR #2068, opened 2025-06-03
WB Migration Improve Cypress test cleanup and artifact handling

teodossidossev:
## What:
Refine Docker Compose cleanup and save Jenkins artifacts on failed builds only

## Why:
- Relying on currentBuild.result was unreliable, because it remains null for a “green” build and we archived all videos on green builds 

## HOW:
- Added a catchError wrapper around dockerCompose.upCmd(...) in both Cypress stages so that on failure we archive screenshots/videos and stop the job without tearing down containers.
- Introduced a flag (caughtError) to detect test failures and invoke error(...) only after archiving artifacts.
- Ensured dockerCompose.downCmd(...) is called only on success, leaving containers/volumes intact when tests fail.
- In cypress.config.js, added an on('after:spec') hook that deletes any spec’s video file if none of its tests failed.

image

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [ ] Tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -58,7 +58,7 @@ pipeline {
                     sh 'npm run validate'
                 }
             }
-             post {
+            post {
                 failure {
                     archiveArtifacts allowEmptyArchive: true, artifacts: 'translation-report.json'
                 }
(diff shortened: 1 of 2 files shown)


packages/shared-components/cypress.config.js
```diff
@@ -46,6 +46,23 @@ module.exports = defineConfig({
           return null;
         }
       });
+
+      on('after:spec', (spec, results) => {
```
svilenvelikov, line 50: Isn't this necessary for the workbench cypress tests as well?

teodossidossev: This is already implemented in e2e-tests/plugins/index.js

Jenkinsfile
```diff
@@ -136,17 +142,39 @@ pipeline {
...
+                                }
+                            }
+
+                            if (caughtError) {
+                                echo "Tests failed — archiving Cypress video artifacts."
+                                archiveArtifacts allowEmptyArchive: true, artifacts: 'cypress/screenshots/**/*.png, cypress/videos/**/*.mp4'
+                                error("Cypress tests failed, job failed.")
                             }
+
+                            echo "Tests passed — skipping video artifacts."
```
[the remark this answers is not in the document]
teodossidossev: will leave it

svilenvelikov (approved):

PR #2076, opened 2025-06-05
GDB-12484: Clicking on User Guides blocks the entire workbench until refresh

boyan-tonchev:
## What
The Workbench crashes when attempting to start a guide.

## Why
A SPARQL editor step includes functionality that allows a query to be set in the SPARQL editor by clicking a button displayed in the guide dialog. This button is implemented using AngularJS functionality.

In shepherd.service.js, guide steps are compiled as they are AngularJS templates. After compilation, the scope is expected to be bound to the element. However, after the migration, this functionality broke — the scope of the element is now always undefined.

## How
A temporary check was added to prevent the Workbench from crashing when the scope is undefined. This avoids a hard failure, but the affected copy-to-editor functionality no longer works.

To restore this functionality, it needs to be reimplemented in a way that does not depend on AngularJS.

## Testing
N/A

## Screenshots
<div>Before</div>

<div>After</div>

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/guides/tour-lib-services/shepherd.service.js
+++ b/packages/legacy-workbench/src/js/angular/guides/tour-lib-services/shepherd.service.js
@@ -921,11 +921,17 @@ function ShepherdService($location, $translate, LocalStorageAdapter, LSKeys, $ro
 
         // We have some Angular bits on the elements and this is needed to make them work
         const scope = angular.element(currentStepElement).scope();
-        if (angular.isFunction(stepDescription.onScope)) {
-            // Step definitions may want to add functions to the scope
-            stepDescription.onScope(scope);
+        // After the migration, the scope here is always undefined.
+        // This broke the "copy query to editor" functionality in the execute-sparql-query plugin.
+        // Note: The scope is used only in the execute-sparql-query plugin.
+        // TODO: To restore this functionality, it needs to be reimplemented in a way that does not depend on AngularJS.
+        if (scope) {
+            if (angular.isFunction(stepDescription.onScope)) {
+                // Step definitions may want to add functions to the scope
+                stepDescription.onScope(scope);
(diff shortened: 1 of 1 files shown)


packages/legacy-workbench/src/js/angular/guides/tour-lib-services/shepherd.service.js
```diff
@@ -921,11 +921,18 @@ function ShepherdService($location, $translate, LocalStorageAdapter, LSKeys, $ro
 
         // We have some Angular bits on the elements and this is needed to make them work
         const scope = angular.element(currentStepElement).scope();
-        if (angular.isFunction(stepDescription.onScope)) {
+        // After the migration, the scope here is always undefined.
+        // This broke the "copy query to editor" functionality in the execute-sparql-query plugin.
+        // Note: The scope is used only in the execute-sparql-query plugin.
+        // TODO: To restore this functionality, it needs to be reimplemented in a way that does not depend on AngularJS.
+        if (scope) {
+            console.log("%c Guide step scope: ", "color: green", scope);
```
svilenvelikov, line 929: I guess that this logger is left intentionally. Also the formatting below is wrong.

svilenvelikov (dismissed):

packages/legacy-workbench/src/js/angular/guides/tour-lib-services/shepherd.service.js
plamen-yordanov, line 929: is the console log left on purpose?

boyan-tonchev: Fixed

PR #2080, opened 2025-06-05
GDB-12484: Updates guides services according to migration changes

boyan-tonchev:
## What
- Removed usage of AngularJS-specific functions from guide services.
- Fixed issues where some steps did not navigate to the correct step URL.

## Why
- After the infrastructure migration, certain functionalities relying on AngularJS stopped working. For example, angular.element(...) was used to wrap DOM elements, but the resulting elements could no longer be interacted with (e.g., clicking didn't work as expected).
 - Since we now use single-spa, routing is handled differently, and in some cases, navigation did not behave as expected due to outdated AngularJS-based logic.

## How
- All functions that had AngularJS-specific implementations were refactored to use native JavaScript instead;
- shepherd.service.js was updated to use single-spa’s navigation instead of AngularJS’s $location service.

## Testing
N/A

## Screenshots
N/A

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [ ] Tests


packages/legacy-workbench/src/js/angular/guides/guide-utils.js
```diff
@@ -2,6 +2,20 @@ import {select} from 'd3';
...
 const d3 = {select};
 const GuideUtils = (function () {
+
+    /**
+     * Checks whether a given DOM element is visible in the document.
+     *
+     * @param {HTMLElement|null} element - The DOM element to check.
+     * @returns {boolean} `true` if the element is visible; otherwise, `false`.
+     */
+    const checkElementVisibility = (element) => {
```
svilenvelikov, line 12: it's a nitpicking, but can you consider renaming this method to isElementVisible?

packages/legacy-workbench/src/js/angular/guides/tour-lib-services/shepherd.service.js
```diff
@@ -986,7 +987,10 @@ function ShepherdService($location, $translate, LocalStorageAdapter, LSKeys, $ro
         typeIcon.setAttribute('gdb-tooltip', $translate.instant('guide.step-type.' + iconTooltip));
         typeIcon.setAttribute('tooltip-placement', 'bottom-right');
 
-        angular.element(currentStepElement.querySelector('.shepherd-header')).prepend(typeIcon);
+        const header = currentStepElement.querySelector('.shepherd-header');
```
svilenvelikov, line 990: I can see that the css class used in this selector has a typo and that you haven't changed it. Can you confirm that this expression is properly doing what's expected?

svilenvelikov (dismissed):

packages/legacy-workbench/src/js/angular/guides/guide-utils.js
boyan-tonchev: Done

packages/legacy-workbench/src/js/angular/guides/tour-lib-services/shepherd.service.js
boyan-tonchev: The class "shepherd-header" is part of the native ShepherdJS library and is applied to the header of the dialog.

svilenvelikov: I see. It's my bad. :)

svilenvelikov (approved):

PR #2081, opened 2025-06-05
GDB-12432 Fix missing cookie banner modal

plamen-yordanov:
## What
Fix missing cookie banner modal

## Why
It wasn't showing up, when a license is deleted

## How
- Synchronized the logic, which shows/hides the banner with master. Recent changes, were not propagated to the migration branch and the logic differed.
- Added a subscription to license change in `onto-footer` to recalculate visibility, whenever a license has changed
- Moved `updateGraphdbLicense` from `then` into `finally`, because the license object will be different if the request to the backend returns an error

## Testing
cypress

## Screenshots
Screenshot from 2025-06-05 15-08-47
Screenshot from 2025-06-05 15-09-07

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/services/license/license.service.ts
+++ b/packages/api/src/services/license/license.service.ts
@@ -11,6 +11,8 @@ import {LicenseContextService} from './license-context.service';
 export class LicenseService implements Service {
   private readonly licenseRestService: LicenseRestService = ServiceProvider.get(LicenseRestService);
   private readonly licenseMapper: LicenseMapper = ServiceProvider.get(LicenseMapper);
+  private readonly trackableProductTypes = ['free', 'sandbox'];
+  private readonly trackableTypesOfUse = ['evaluation', 'this is an evaluation license'];
 
   /**
    * Retrieves the current license information.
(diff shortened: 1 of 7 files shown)


packages/api/src/services/license/license.service.ts
```diff
@@ -26,21 +28,24 @@ export class LicenseService implements Service {
...
    * @returns {boolean} True if the license is a free license, false otherwise.
    */
-  isFreeLicense(): boolean {
+  isTrackableLicense(): boolean {
     const license = ServiceProvider.get(LicenseContextService).getLicense();
     const licenseTypeOfUse = license?.typeOfUse?.toLowerCase() || '';
-    return license?.productType === 'free'
-      || licenseTypeOfUse === 'evaluation'
-      || licenseTypeOfUse === 'this is an evaluation license';
+    const productType = license?.productType?.toLowerCase() || '';
```
svilenvelikov, line 46: this normalization should be in the License model. Also the model is already initialized with default values where needed.

plamen-yordanov: Typescript thinks `license` may be undefined here, because `getLicense` (`getContextPropertyValue`) may return undefined. The logic is written in a way, that license should always be defined, but TS doesn't know that

packages/shared-components/src/components/onto-footer/onto-footer.tsx
```diff
@@ -78,14 +80,26 @@ export class OntoFooter {
...
   }
+
+  private setCookieConsentVisibility(user?: AuthenticatedUser) {
+    this.shouldShowCookieConsent = this.isTrackingAllowed() && !new CookieConsent(user?.appSettings?.COOKIE_CONSENT).policyAccepted;
+  }
+
+  private subscribeToLicenseChange() {
+    this.subscriptions.add(
+      this.licenseContextService.onLicenseChanged(() => {
+        this.setCookieConsentVisibility();
```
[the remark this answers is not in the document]
plamen-yordanov: I was expecting a conversation here :)

PR #2083, opened 2025-06-06
fix(GDB-12503) WBM: Display logout button for external users based on externalAuthUser flag

teodossidossev:
## WHAT:
Fixed an issue where the logout button was not shown for users authenticated via OpenLDAP, despite the expectation that it should appear

## WHY:
The existing logic to determine whether to display the logout button was incorrectly based solely on the user.external property. However, for externally authenticated users (e.g., via OpenLDAP), a separate config property in jwt-auth.service.js (hasExternalAuthUser) is the reliable source of truth indicating such authentication. Without this fix, users under OpenLDAP couldn't properly log out from the UI.

## HOW:
- Introduced a new hasExternalAuthUser property to SecurityConfig and populated it in jwt-auth.service.js.
- Updated OntoUserMenu to receive securityConfig as a prop.
- Adjusted the logout button condition to check securityConfig.hasExternalAuthUser instead of user.external.
- Passed securityConfig to OntoUserMenu in OntoHeader
- Several Sonar issues fixed

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [ ] Tests

### DIFF
--- a/packages/shared-components/src/components/onto-user-menu/onto-user-menu.tsx
+++ b/packages/shared-components/src/components/onto-user-menu/onto-user-menu.tsx
@@ -1,5 +1,11 @@
 import {Component, h, Prop, State, Element} from '@stencil/core';
-import {AuthenticatedUser, AuthenticationService, navigateTo, ServiceProvider} from '@ontotext/workbench-api';
+import {
+  AuthenticatedUser,
+  AuthenticationService,
+  navigateTo,
+  SecurityConfig,
+  ServiceProvider
+} from '@ontotext/workbench-api';
 
 /**
  * This component displays the current user's name and provides options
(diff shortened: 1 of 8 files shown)


packages/shared-components/src/components/onto-user-menu/onto-user-menu.tsx
```diff
@@ -40,7 +48,7 @@ export class OntoUserMenu {
             <section class='onto-user-menu-dropdown'>
               <translate-label onClick={navigateTo('/settings')}
                                labelKey={'user_menu.my_settings'}></translate-label>
-              {!this.user.external ?
```
plamen-yordanov, line 43: `user.external` returns false, when LDAP?

plamen-yordanov (approved):

PR #2086, opened 2025-06-09
GDB-12505 Fix enabled request on repository deletion

plamen-yordanov:
## What
Fix unnecessary autocomplete `/enabled` check, when the selected repository is deleted

## Why
The backend throws an error, since the repository does not exist anymore

## How
Made `updateSelectedRepository` to return the resolved promise of `validatePropertyChange`. Previously, when a repository was deleted, `updateSelectedRepository` was called, but the validations from `validatePropertyChange` happen in a promise, in a separate thread tick. Then `repositoryIsSet` was broadcasted with wrong store data. This caused the application to make a request for autocomplete enabled

## Testing
n/a

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/api/src/models/repositories/repository.ts
+++ b/packages/api/src/models/repositories/repository.ts
@@ -19,6 +19,7 @@ export class Repository extends Model<Repository> implements RepositoryReference
   readable: boolean | undefined;
   writable: boolean | undefined;
   unsupported: boolean | undefined;
+  isNew?: boolean;
 
   constructor(data?: Partial<Repository>) {
     super();
@@ -34,8 +35,9 @@ export class Repository extends Model<Repository> implements RepositoryReference
     this.readable = data?.readable;
     this.writable = data?.writable;
     this.unsupported = data?.unsupported;
+    this.isNew = data?.isNew;
   }
-  
+
   toRepositoryReference(): RepositoryReference {
     return {
       id: this.id,
--- a/packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
+++ b/packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
@@ -396,21 +396,24 @@ repositories.service('$repositories', ['toastr', '$rootScope', '$timeout', '$loc
             eventEmitterService.emit('repositoryWillChangeEvent', eventData, (eventData) => {
                 if (!eventData.cancel) {
                     const repositoryContextService = ServiceProvider.get(RepositoryContextService);
-                    // this will update the values in the local storage and trigger context change for other opened tabs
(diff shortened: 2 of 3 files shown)


plamen-yordanov:
Re-created the MR, because branch name was wrong

boyan-tonchev (dismissed):

packages/api/src/models/repositories/repository.ts
```diff
@@ -19,6 +19,7 @@ export class Repository extends Model<Repository> implements RepositoryReference
   readable: boolean | undefined;
   writable: boolean | undefined;
   unsupported: boolean | undefined;
+  isNew?: boolean;
```
svilenvelikov, line 22: The repository has an `id`. Did you check when the id is assigned and if you can say that the repo is new given the id is still missing?

packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
```diff
@@ -396,22 +396,25 @@ repositories.service('$repositories', ['toastr', '$rootScope', '$timeout', '$loc
...
                 if (!eventData.cancel) {
                     const repositoryContextService = ServiceProvider.get(RepositoryContextService);
-                    // this will update the values in the local storage and trigger context change for other opened tabs
-                    repositoryContextService.updateSelectedRepository(repo);
-                    this.setRepositoryHeaders(repo);
-                    $rootScope.$broadcast('repositoryIsSet', {newRepo: true});
-
                     // if the current repo is unreadable by the currently logged in user (or free access user)
                     // we unset the repository
                     if (repo && !$jwtAuth.canReadRepo(repo) && !$jwtAuth.hasGraphqlReadRights(repo)) {
```
yordanalexandrov, line 401: repo is set as an empty string alot
does it not crash when you try to set `repo.isNew` ?

plamen-yordanov: It does, yes. I saw it today, but haven't pushed the fix yet

packages/api/src/models/repositories/repository.ts
svilenvelikov: Well. Right. We are inheriting the legacy where the repository name is used as an id and when a repo is created, the name is provided by the user through the UI

packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
plamen-yordanov: Added check for empty string. In this case, the repository is reset, so the context will emit `undefined`

PR #2089, opened 2025-06-09
GDB-12484: Interactive guides don't work on the migration branch

boyan-tonchev:
## What
- Extended the dropdown component with the ability to configure whether it should automatically close when the user clicks outside of it.
- Fixed the "View all <count> instances in SPARQL" link in the right-hand Class Hierarchy info sidebar, which wasn't working.
- Updated all guide steps to function correctly in the migration branch.

## Why
- When the guide is on a step that requires selecting a repository, the dropdown must remain open even when the user clicks elsewhere. Otherwise, the dropdown closes and breaks the guide flow.
- The link used a href attribute for navigation, which did not work correctly with the Single-SPA setup.

## How
- When the "Select repository" guide step is active, the dropdown auto-close behavior is disabled. Once the step is completed, auto-close is re-enabled.
- Refactored the link to use ng-click. When the user clicks the link, the bound function uses $location to navigate to the resource view.

## Testing
Fixed and re-enabled all previously skipped tests.

## Screenshots
N/A

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/guides/steps/complex/select-repository/plugin.js
+++ b/packages/legacy-workbench/src/js/angular/guides/steps/complex/select-repository/plugin.js
@@ -3,8 +3,17 @@ const getRepositoryName = (services, options) => {
 };
 
 const getRepositoryElementSelector = (services, options) => {
-    return services.GuideUtils.getGuideElementSelector(`repository-${getRepositoryName(services, options)}-button`);
+    return services.GuideUtils.getGuideElementSelector(`repository-id-${getRepositoryName(services, options)}`);
 };
+
+const setRepositorySelectorAutoClose = (autoClose) => {
+    const component = document.querySelector('.onto-repository-selector');
+    if (component) {
+        // Enable auto-close when the guide step is closed.
+        component.autoClose = autoClose;
+    }
+}
+
 
 PluginRegistry.add('guide.step', [
     {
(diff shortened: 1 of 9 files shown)


packages/legacy-workbench/src/js/angular/guides/steps/complex/select-repository/plugin.js
```diff
@@ -38,7 +47,8 @@ PluginRegistry.add('guide.step', [
                             throw (error);
                         }),
                     show: (guide) => () => {
-                        $('#repositorySelectDropdown').addClass('autoCloseOff');
+                        setRepositorySelectorAutoClose(false);
+                        $('.onto-repository-selector').addClass('auto-close-off');
```
svilenvelikov, line 51: wasn't the goal of the task to be removed usage of angular utility functions and jquery?

packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
svilenvelikov, line 1: One thing I don't like for applications that are organized as a monorepo is that it's quite easy for a developer to forget about the project boundaries and mix changes for different projects in single commit and MR. :(

boyan-tonchev: I will make MR for this changes in different MR

packages/legacy-workbench/src/js/angular/guides/steps/complex/select-repository/plugin.js
boyan-tonchev: Removed

packages/shared-components/src/components/onto-dropdown/onto-dropdown.tsx
boyan-tonchev: done

svilenvelikov (approved):

PR #2110, opened 2025-06-16
GDB-12409 Fix different user icons

plamen-yordanov:
## What
Fix different user icons

## Why
The user menu has a font awesome user icon, whereas the rest of the app uses the old icon

## How
- Changed the old user icon with the font awesome solid version.
- Replaced `em` with `i` tags
- Added styling

## Testing
n/a

## Screenshots
Screenshot from 2025-06-16 13-22-19
Screenshot from 2025-06-16 13-22-37
Screenshot from 2025-06-16 13-22-43

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/root-config/src/styles/css/icons.css
svilenvelikov, line 1: I'm thinking if we should also remove the user icon from the icon set here https://github.com/Ontotext-AD/ontotext-ui-guidelines/tree/master/ontotext-icon-pack

packages/legacy-workbench/src/css/login.css
```diff
@@ -0,0 +1,5 @@
+#wb-login .fa-user {
+    font-size: 2rem;
```
yordanalexandrov, line 2: Aren't there styles for the icon size in the FA stylesheet?

yordanalexandrov: `fa-2xl` sets the font-size to 2em

packages/root-config/src/styles/css/icons.css
svilenvelikov: OK. Leave the icons in the legacy icon set.

packages/legacy-workbench/src/css/login.css
svilenvelikov: No. You shouldn't use it now. These changes should happen during the next phase where we are supposed to apply a new theme based on a design system. Eventually.

svilenvelikov (approved):

yordanalexandrov (approved):

PR #2117, opened 2025-06-18
fix(GDB-11410): WBM Fix import-user-data-file-upload tests

teodossidossev:
## WHAT:
Fixed and re-enabled the flaky `import-user-data-file-upload.spec.js` end-to-end test suite.

## WHY:
The test suite was disabled due to migration caused issues

## HOW:
- in `import-user-data-file-upload.spec.js`, the `createFile()` call was decoupled from the `selectFile()` call. The file object is now created once and stored in a `const` variable. This constant is then passed to the `selectFile` step.
- re-enabled all tests
- removed commented and obsolete code

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [X] Tests

### DIFF
--- a/e2e-tests/steps/import/import-steps.js
+++ b/e2e-tests/steps/import/import-steps.js
@@ -67,13 +67,7 @@ class ImportSteps {
         }
 
         cy.visit('/import#' + type);
-
-        // cy.get('.ot-splash').should('not.be.visible');
-
         cy.get('#import-' + type).should('be.visible');
-
-        // cy.get('.ot-loader').should('not.be.visible');
-
         return this;
     }
 
@@ -98,7 +92,6 @@ class ImportSteps {
     }
 
     static openUserDataTab() {
-        // cy.get('.ot-loader').should('not.be.visible');
         return this.getTabs().eq(0).click();
     }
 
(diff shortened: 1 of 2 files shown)


e2e-tests/steps/import/import-steps.js
```diff
@@ -272,7 +262,7 @@ class ImportSteps {
     }
 
     static importFile(index) {
-        this.getResource(index).find('.import-resource-action-import-btn').click();
+        this.getResource(index).find('.import-resource-action-import-btn').click({force: true});
```
plamen-yordanov, line 265: why is force needed?

svilenvelikov, line 265: Why with the force?

teodossidossev: changed to .invoke('show').trogger('click') to simulate hover+click

teodossidossev: changed to .invoke('show').trigger('click') to simulate hover+click

plamen-yordanov (approved):

svilenvelikov (approved):

PR #2118, opened 2025-06-18
GDB-12553 add repository filtering by rights

plamen-yordanov:
## What
Add filtering to the repository dropdown, based on user rights

## Why
It currently shows repositories, to which the user has no access to

## How
Implemented a `canReadRepo` method in `authentication.service`, which contains the existing logic from `jwt-auth.service`. Added a filter by that function in `onto-header`

## Testing
jest

## Screenshots
What repositories, the user sees
image
What repositories does the user have access to
image
Available repositories at the moment
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/services/security/authentication.service.ts
+++ b/packages/api/src/services/security/authentication.service.ts
@@ -5,11 +5,14 @@ import {EventService} from '../event-service';
 import {Logout} from '../../models/events';
 import {SecurityContextService} from './security-context.service';
 import {Repository} from '../../models/repositories';
+import {RepositoryService} from '../repository';
 
 /**
  * Service responsible for handling authentication-related operations.
  */
 export class AuthenticationService implements Service {
+  private readonly repositoryService = ServiceProvider.get(RepositoryService);
+
   login(): string {
     return 'Authentication.login from the API';
   }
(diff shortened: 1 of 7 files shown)


packages/api/src/services/security/authentication.service.ts
```diff
@@ -58,14 +60,55 @@ export class AuthenticationService implements Service {
...
   /**
-   * Checks if security is enabled.
+   * Checks if security is enabled
    * @returns {boolean} True if security is enabled, false otherwise.
    */
   isSecurityEnabled(): boolean {
     return !!this.getSecurityConfig()?.enabled;
   }
 
+  canReadRepo(repository?: Repository): boolean {
```
svilenvelikov, line 70: I'm not convinced this method should be here in the authentication service. Why did you decide to couple this service with some repository stuff?

packages/api/src/services/security/authentication.service.ts
```diff
@@ -58,14 +60,55 @@ export class AuthenticationService implements Service {
...
+    }
+
+    const config = this.getSecurityConfig();
+    const user = this.getAuthenticatedUser();
+
+    if (config?.enabled) {
+      if (!user) {
+        return false;
+      }
+      if (this.hasRole(Authority.ROLE_ADMIN) || this.hasRole(Authority.ROLE_REPO_MANAGER)) {
```
svilenvelikov, line 82: This can be extracted in a method.

packages/api/src/services/security/authentication.service.ts
```diff
@@ -58,14 +60,55 @@ export class AuthenticationService implements Service {
...
+        return true;
+      }
+      if (repository.id === 'SYSTEM') {
```
svilenvelikov, line 85: This can be a method as well. Maybe in the repo service?

packages/api/src/services/security/authentication.service.ts
```diff
@@ -58,14 +60,55 @@ export class AuthenticationService implements Service {
...
   }
+
+  private getAuthenticatedUser(): AuthenticatedUser | undefined {
+    return ServiceProvider.get(SecurityContextService).getAuthenticatedUser();
+  }
+
+  private hasBaseRights(action: string, repo: Repository): boolean {
+    const repoId = repo.location ? `${repo.id}@${repo.location}` : repo.id;
+    const overCurrentRepo = `${action}_REPO_${repoId}`;
+    const overAllRepos = `${action}_REPO_*`;
```
svilenvelikov, line 105: These probably also should go into the repo service

packages/api/src/services/security/authentication.service.ts
plamen-yordanov: done

PR #2126, opened 2025-06-19
fix(GDB-11410): WBM Enable tests broken due to migration part 4

teodossidossev:
## WHAT:
Fixed when needed and re-enabled skipped test suites due changes in the main menu.

## WHY:
The test suites were disabled due to migration caused issues

## HOW:
- re-enabled tests

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [X] Tests


plamen-yordanov (dismissed):

e2e-tests/e2e-legacy/resource/resource.spec.js
```diff
@@ -370,7 +366,8 @@ describe.skip('Resource view', () => {
         });
     });
 
-    context('Download as', () => {
+    // TODO: Fix me. Broken NOT due to migration (Error: CORS violation when downloading from https://w3c.github.io/json-ld-api/tests/compact/0007-context.jsonld)
```
svilenvelikov, line 369: Interesting. This should not be the case.

svilenvelikov: See this comment https://github.com/Ontotext-AD/graphdb-workbench/blob/GDB-10373-workbench-with-single-spa/e2e-tests/e2e-flaky/import/import-server-files-operations.spec.js#L64
I'm pretty sure that this download test relies on graphdb to be started with the whitelist for this resource

PR #2175, opened 2025-07-01
GDB-12336: Version static assets to prevent browser caching issues

boyan-tonchev:
## What
- Added versioning to assets files.
- Previously, some CSS files were being copied to the final distribution even though they were already imported via vendor.js.

## Why
Currently, the root project’s files are served without versioning. This can lead to browser caching issues — especially after deployments where styles change but the file names remain the same. As a result, clients may continue to use outdated styles until the cache expires or is manually cleared.

## How
- Added a v (version) parameter based on the current GraphDB version.
- Created a new source folder that holds all asset files imported via vendor.js. – The styles/css folder now includes only the files that need to be copied to the final build.

## Screenshots
N/A

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [X] Tests


svilenvelikov:
Fix the MRname and description pls

yordanalexandrov (approved):

PR #2176, opened 2025-07-02
GDB-12626 Fix GQL repository dropdown filtering

plamen-yordanov:
## What
Fix repository dropdown filtering with GQL

## What
Currently GQL access is not taken into account

## How
Migrated the `hasGraphQlAuthority` functionality from the legacy workbench and added the filter in the repository dropdown items

## Testing
n/a

## Screenshots

Screenshot from 2025-07-02 14-41-21
Screenshot from 2025-07-02 14-41-34

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/api/src/services/security/authentication.service.ts
+++ b/packages/api/src/services/security/authentication.service.ts
@@ -12,6 +12,7 @@ import {RepositoryService} from '../repository';
  */
 export class AuthenticationService implements Service {
   private readonly repositoryService = ServiceProvider.get(RepositoryService);
+  private readonly securityContextService = ServiceProvider.get(SecurityContextService);
 
   login(): string {
     return 'Authentication.login from the API';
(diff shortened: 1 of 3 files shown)


packages/api/src/services/security/authentication.service.ts
```diff
@@ -94,6 +95,30 @@ export class AuthenticationService implements Service {
     return true;
   }
 
+  canReadGqlRepo(repository: Repository): boolean {
+    if (!repository || repository.id === '') {
```
yordanalexandrov, line 99: Is it possible for `repository.id` to be `undefined | null` ?
If so this will allow it to proceed.

plamen-yordanov: It is a mandatory field, so it shouldn't be possible. Come to think of it, the same should be true for an empty string

PR #2191, opened 2025-07-03
GDB-12654 Fix broken toasts without ssl

plamen-yordanov:
## What
Fix missing toast messages, when using GDB without ssl

## Why
They are not being displayed due to an error

## How
Replaced `window.crypto.randomUUID` with a custom generator for ids. When GDB is deployed without SSL and not on a local machine, there is no `window.crypto`, which results in an error, when opening migration toasts

## Testing
jest

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


boyan-tonchev:
We have tests that stub window.crypto.randomUUID, and they should be updated accordingly.

svilenvelikov (approved):

PR #2193, opened 2025-07-03
GDB-12346 - Replace legacy endpoints and update mappers  and payloads

DesiBorisova:
## TODO
- Still clarifying response and payload objects with BE for some endpoints.
- Stubs and fake backend will be updated after that.

## What
Some code removed and updated in accordance with the changes to the back-end.

## Why
BE optimizing and refactoring code for better maintenance. Changes require FE updates as well.

## How
Mappers and payload objects updated.

## Testing
Will be updated.

## Screenshots
N/A

## Checklist
- [ ] Branch name
- [ ] Target branch
- [ ] Commit messages
- [ ] Squash commits
- [ ] MR name
- [ ] MR Description
- [ ] Tests

### DIFF
--- a/src/i18n/locale-en.json
+++ b/src/i18n/locale-en.json
@@ -500,7 +500,7 @@
                         "tooltip": "Limit the maximum number of triples returned by Full-text search. The default value is automatic and determined at runtime. If not sure, leave the default value.",
                         "placeholder": "Automatic value"
                     },
-                    "sparql_search": {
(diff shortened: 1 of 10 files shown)


svilenvelikov:
@DesiBorisova we discussed these changes which our backend colleagues intended to do and managed to convince them that the current approach is better. So, the changes in tool models won't be needed if they stick to what we agreed.

PR #2198, opened 2025-07-07
feat(GDB-12651) Add code coverage support for legacy Workbench E2E tests

teodossidossev:
## What:
- Introduce @bahmutov/cypress-code-coverage plugin and nyc instrumentation across the repository
- Update e2e-tests/cypress-legacy.config.js to conditionally load the coverage plugin when COVERAGE=true
- Add coverage-related ignores (coverage/, .nyc_output/) to e2e-tests/.gitignore
- Ensure plugins return config and support file includes coverage support
- Extend top-level and e2e-tests package.json scripts and dependencies for nyc, cross-env, and coverage commands
- Configure Babel loader with Istanbul plugin in webpack.config.common.js
- Provide a new Docker Compose file (docker-compose-coverage.yaml) for coverage runs
- Supply a dedicated coverage.Jenkinsfile to orchestrate install, instrument, build, restore, and E2E coverage stages in CI

## Why:
- We need to measure and report client-side code coverage for the legacy Workbench UI

## How:
- Leverage nyc instrument to rewrite source in place for coverage collection.
- Use Babel’s Istanbul plugin for instrumentation in Webpack.
- Wrap Cypress runs with @bahmutov/cypress-code-coverage to capture and upload coverage data.
- Automate the full pipeline in Jenkins with a new coverage.Jenkinsfile

## Screenshots
image

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [X] Tests


e2e-tests/package.json
```diff
@@ -29,7 +30,10 @@
     "url": "git+https://github.com/Ontotext-AD/graphdb-workbench.git"
   },
   "devDependencies": {
-    "cypress": "^14.5.0",
+    "@bahmutov/cypress-code-coverage": "^2.7.2",
+    "babel-plugin-istanbul": "^7.0.0",
+    "cross-env": "^7.0.3",
```
svilenvelikov, line 35: Can you consider another library here, because cross-env is in maintanance mode since 2020.

e2e-tests/package.json
```diff
@@ -40,5 +44,16 @@
     "del": "^8.0.0",
     "minimist": "^1.2.8",
     "mocha-junit-reporter": "^2.2.1"
+  },
+  "nyc": {
+    "all": true,
```
svilenvelikov, line 49: This is probably not needed, because running `nyc instrument` by default should be instrumenting all matched source code. 
Additionally, if you narrow down the scope with the include `packages/legacy-workbench/src/**/*.js`, then you probably don't need the exclude configuration.

e2e-tests/package.json
```diff
@@ -40,5 +44,16 @@
     "del": "^8.0.0",
     "minimist": "^1.2.8",
     "mocha-junit-reporter": "^2.2.1"
+  },
+  "nyc": {
```
svilenvelikov, line 48: I wonder if it isn't better to put the nyc config in a separate nyc configuration file for better visibility

docker-compose-coverage.yaml
```diff
@@ -0,0 +1,63 @@
...
+    healthcheck:
+      test: ["CMD-SHELL", "curl -f -H 'Accept: text/html' ${GRAPHDB_URL}/index.html"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+
+  workbench:
+    build:
+      context: .
+      dockerfile: root-config.Dockerfile
```
svilenvelikov, line 29: I now see that this Dockerfile is not renamed, but I remember a discussion that we must rename it to workbench.Dockerfie or application.Dockerfile or similar, because `root-config` is not appropriate and doesn't make sence in this context

coverage.Jenkinsfile
```diff
@@ -0,0 +1,169 @@
...
+                        secrets: [
+                            [
+                                destination: 'file',
+                                filePath: 'graphdb.license',
+                                notation: 'keeper://AByA4tIDmeN7RmqnQYGY0A/file/graphdb.license'
+                            ]
+                        ]
+                    ]]) {
+                        sh 'cp graphdb.license ./e2e-tests/fixtures/'
+                        archiveArtifacts allowEmptyArchive: true, artifacts: 'graphdb.license'
```
svilenvelikov, line 106: I think we don't need this here.

coverage.Jenkinsfile
```diff
@@ -0,0 +1,169 @@
...
+                                    options: ["--abort-on-container-exit", "--exit-code-from cypress"]
+                                )
+                            } catch (e) {
+                                caughtError = true
+                            }
+                        }
+
+                        if (caughtError) {
+                            echo "Tests failed — archiving all Cypress artifacts."
+                            archiveArtifacts allowEmptyArchive: true, artifacts: 'e2e-tests/coverage/**, e2e-tests/report/screenshots/**/*.png, e2e-tests/report/videos/**/*.mp4, e2e-tests/cypress/logs/*.log'
```
svilenvelikov, line 128: not sure if we need these too?

e2e-tests/package.json
teodossidossev: replaced

e2e-tests/package.json
teodossidossev: extracted

docker-compose-coverage.yaml
teodossidossev: renamed

coverage.Jenkinsfile
teodossidossev: removed

svilenvelikov (approved):

PR #2206, opened 2025-07-10
fix(GDB-12618) Sanitize trailing slash from pathname before SPA bootstrap

teodossidossev:
## WHAT
Added an inline <script> that strips a trailing slash from window.location.pathname before any <base> tag or import-map is applied.

## WHY
To ensure the SPA always loads under a canonical URL without a trailing slash

## HOW
The script runs immediately on page load as an IIFE, checks pathname.endsWith('/') and pathname.length > 1, then uses history.replaceState to rewrite the URL (preserving search and hash) before the rest of the application bootstraps.

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [ ] Tests

### DIFF
--- a/packages/root-config/src/index.ejs
+++ b/packages/root-config/src/index.ejs
@@ -2,6 +2,25 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
+    <!--
+      SANITIZE TRAILING SLASH FROM PATHNAME BEFORE SPA BOOTSTRAP
+      This inline script executes immediately upon page load,
+      before any <base> tag or import-map is processed. It detects
+      if the URL’s pathname is longer than 1 character and ends
+      with a “/”, and if so, removes that trailing slash while
+      preserving query strings and hash fragments via
+      history.replaceState.
+    -->
+    <script>
+      (function() {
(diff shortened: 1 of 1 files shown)


packages/root-config/src/index.ejs
```diff
@@ -2,6 +2,25 @@
...
+      This inline script executes immediately upon page load,
+      before any <base> tag or import-map is processed. It detects
+      if the URL’s pathname is longer than 1 character and ends
+      with a “/”, and if so, removes that trailing slash while
+      preserving query strings and hash fragments via
+      history.replaceState.
+    -->
+    <script>
+      (function() {
+        const { pathname, search, hash } = window.location;
```
plamen-yordanov, line 16: will this affect [isHomePage()](https://github.com/Ontotext-AD/graphdb-workbench/blob/master/packages/api/src/services/utils/routing-utils.ts#L39)?

teodossidossev: no, it checks for trailing slash when pathname.length > 1

plamen-yordanov (approved):

PR #2209, opened 2025-07-11
GDB-12667 fix wrong permission banner to GQL user

plamen-yordanov:
## What
Fix wrong permission banner to gql user

## Why
The application should show a different banner, whenever a user has no permissions

## How
- Added `experimentalSlotFixes` to `stencil.config` This fixes a stencil issue, where `slot` elements are missing in the DOM
- Added `routing.service`, the logic for which is taken from angularjs $routerProvider. The purpose is to have ability to determine, which is the current active route and get the plugin associated with it. This is needed to calculate user authorities, based on the route and decide, whether to display the red banner or the authorities banner
- Added additional check in `onto-layout#setPermissions` to show red/authority banner based on present authorities
- Added condition to not load autocomplete, if user has no read repo rights. This protects from autocomplete requests on certain pages, which returned 403 and the banner was wrongfully shown
- Moved the restricted pages reset inside a repository change subscription. This solves a racing condition, where pages were cleared after a 403 is received and caused a white screen

## Testing
n/a

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/api/src/services/routing/routing.service.ts
```diff
@@ -0,0 +1,65 @@
...
+ * It works with the application's route configuration obtained from the PluginRegistry.
+ */
+import {RouteRegexParamPair} from '../../models/routing/route-regex-param-pair';
+import {RouteItemModel} from '../../models/routing/route-item-model';
+import {getPathName} from '../utils';
+import {ModelList} from '../../models/common';
+import {Service} from '../../providers/service/service';
+
+export class RoutingService implements Service {
+  private readonly routeConfig = window.PluginRegistry.get('route');
```
svilenvelikov, line 15: Didn't we have a service for abstracting the global `window` object access? If not, then we better have one

packages/api/src/services/routing/routing.service.ts
svilenvelikov, line 1: No unit tests for the stuff in the API? We spoke that you borrowed some code form angularjs, but still you can throw a bunch of tests here

packages/api/src/services/routing/routing.service.ts
plamen-yordanov: we don't have a service for abstracting `window`, but we have some utilities, which encapsulate certain methods of it, for example [routing-utils](https://github.com/Ontotext-AD/graphdb-workbench/blob/master/packages/api/src/services/utils/routing-utils.ts#L22)

packages/api/src/services/security/authentication.service.ts
```diff
@@ -142,6 +143,78 @@ export class AuthenticationService implements Service {
...
+
+    return true;
+  }
+
+  private resolveAuthorities(authoritiesList: string[]) {
+    // If no authorities list is provided, return undefined.
+    if (!authoritiesList) {
+      return [];
+    }
+
```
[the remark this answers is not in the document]
plamen-yordanov: Marked the parameter as `optional`

svilenvelikov (dismissed):

packages/api/src/services/window/window.service.ts
```diff
@@ -0,0 +1,21 @@
...
+  /**
+   * Retrieves the route plugins from the global window object.
+   *
+   * This function accesses the application's plugin registry and returns
+   * the registered route extensions. It provides access to all routes
+   * that have been registered by various plugins in the application.
+   *
+   * @returns {RouteModel} The route plugin registry containing all registered routes.
+   */
+  static getRoutePlugins(): RouteModel {
```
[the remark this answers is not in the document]
plamen-yordanov: I will yes

plamen-yordanov:
Replaced `location.pathname` with `getPathName`, because it was breaking permission resolving, when behind context

packages/api/src/services/routing/routing.service.ts
```diff
@@ -0,0 +1,66 @@
+/**
+ * Service responsible for handling application routing functionality.
+ *
+ * This service provides methods for matching URL paths against registered routes,
+ * converting route patterns to regular expressions, and retrieving active routes.
+ * It works with the application's route configuration obtained from the PluginRegistry.
+ */
```
svilenvelikov, line 7: nitpick: This jsdoc should be before the class declaration and not above imports

svilenvelikov (approved):

packages/api/src/services/routing/routing.service.ts
plamen-yordanov: moved

svilenvelikov (approved):

PR #2219, opened 2025-07-16
GDB-12721 Fix application crash with multiple tabs open

plamen-yordanov:
## What
Fix application crash with multiple tabs open

## Why
There was an error, when enabling security. Also, it showed the login screen, when we disabled the security

## How
- Introduced a new `authenticated` property for the local store. The main issue was, that the other tabs were relying on in-memory data for resolving conditions. Other tabs have no way of knowing in-memory changes. The new storage property is used to determine, whether a user is authenticated
- Added setting of the new property, whenever we set a token (explicit authentication, internal or external), or whenever security config is broadcasted (this covers logout and disabled security)
- Added additional `broadcastSecurityInit`, whenever security changes and is enabled. There was a missing broadcast in this case

## Testing
n/a

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


svilenvelikov:
With this now, I can't load the application
image
And in the backend I see
```
[WARN ] 2025-07-16 18:47:40,915 [http-nio-7200-exec-3 | c.o.g.s.TokenManager] Auth token hash mismatch.
```

plamen-yordanov:
It isn't introduced with this MR, but with the bootstrap refactor. We could mitigate it by adding a catch clause in the initial `/security/all` request. If we get 401 => redirect to login. It should solve this, until we migrate the interceptor. What do you think?

PR #2222, opened 2025-07-17
GDB-12738 - Skip test with new API key property name

DesiBorisova:
## What
Both tests in ttyg-initial-state-with-selected-repository.spec.js skipped. API key in test was updated.

## Why
The BE has updated the API key name and the Workbench still expects the old one. This fails the BE builds. Hence, the test is skipped until the BE releases successfully.

## How
I skipped the test and updated the property.

## Testing
N/A

## Screenshots
N/A

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


DesiBorisova:
New MR created, due to Jenkins rollback. https://github.com/Ontotext-AD/graphdb-workbench/pull/2227

PR #2228, opened 2025-07-18
GDB-12677: Wrong Redirection to Login Page in Free Access

boyan-tonchev:
GDB-12677: Wrong Redirection to Login Page in Free Access

## What
The user is redirected to the login page even when they have the appropriate rights to see the page.

Steps to reproduce:
- Enable security;
- Enable Free Access;
- In the settings dialog, grant GraphQL rights only to a specific repository;
- Select the repository with GraphQL only rights (IMPORTANT);
- Log out.

Actual behavior: The login page is loaded instead.
Expected behavior: The Import view should be shown.

## Why
The issue arises because authority formats are mismatched.

When the Free Access configuration dialog is closed, the application stores the user's authorities in the wrong format. Because of this:
- When the user clicks the Import menu, the frontend believes the user has permission and tries to load the view;
- However, the backend evaluates the request using a different authority format and returns a 401 Unauthorized error;
- As a result, the user is redirected to the login page, even though they technically have the correct rights.

## How
Add backend call to fetch updated free access data and refresh workbench state accordingly.

## Screenshots
N/A

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [] Tests

### DIFF
--- a/e2e-tests/e2e-security/setup/users-and-access/user-and-access.spec.js
+++ b/e2e-tests/e2e-security/setup/users-and-access/user-and-access.spec.js
@@ -1,0 +1,63 @@
+import {RepositoriesStubs} from "../../../stubs/repositories/repositories-stubs";
+import {UserAndAccessSteps} from "../../../steps/setup/user-and-access-steps";
+import {LoginSteps} from "../../../steps/login-steps";
+import {ModalDialogSteps} from "../../../steps/modal-dialog-steps";
+import ImportSteps from "../../../steps/import/import-steps";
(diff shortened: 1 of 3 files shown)


svilenvelikov:
Discussed about changing this so that mappers from backend to ui authority model to be moved in the api module and applying the transformation only on input/output and not somewhere in between. Internally the UI application should work only with the UI model.

svilenvelikov (approved):

PR #2237, opened 2025-07-23
GDB-12753 Update tracking when accepting cookie consent

yordanalexandrov:
## What
When the user clicks ok to accept the cookie consent, the consent is updated in hte principal and tracking is applied

## Why
Since the implementation of the cookie consent banner was moved to the shared components, there was no event to notify the legacy workbench about the user's action

How
- added an event `COOKIE_CONSENT_CHANGED_EVENT` which is fired when user accepts the cookies from the banner
- added handling of the event in the `mainCtrl` to update the cookie consent using the `TrackingService`
- set correct `CookieConsent` when there is none set in the `appSettings`

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/legacy-workbench/src/js/angular/controllers.js
```diff
@@ -284,6 +286,22 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, toastr, $location, $repos
         }
     });
 
+    const subscribeToCookieConsentChanged = () => {
+        document.body.addEventListener(COOKIE_CONSENT_CHANGED_EVENT, updateCookieConsentHandler)
+        return () => document.body.removeEventListener(COOKIE_CONSENT_CHANGED_EVENT, updateCookieConsentHandler);
+    }
+
+    function updateCookieConsentHandler(consentChangedEvent) {
```
svilenvelikov, line 294: This function would be hoisted as it is now, but better move it before the subscription above.

svilenvelikov (approved):

PR #2241, opened 2025-07-24
GDB-12757 - Introduce new Context size field in TTYG

DesiBorisova:
## What
When a user uses an API other than `openai-assistants`, a new field (Context size) will appear in the Agent create/edit modal. When a user uses `openai-assistants` API, the Context size field will not be visible.

## Why
Previously, the context window size was set via a static property, which made it impossible to configure different sizes for different agents. This posed a risk where ChatMemory would not be evicted correctly in multi-agent setups, potentially breaking conversations.

The fix introduces a required UI field for setting the context window size per agent, ensuring that each agent can be configured independently and ChatMemory eviction behaves correctly.

## How
I extended the models for the new field and restore to default functionality. I edited the templates to retain the original bootstrap grid size for the fields when the Context size field is not rendered.

## Testing
Tests added.

## Screenshots
Field visible when API is not `openai-assistants`
Screenshot from 2025-07-24 12-52-38

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/legacy-workbench/src/js/angular/ttyg/services/agents.mapper.js
```diff
@@ -172,6 +174,7 @@ export const agentModelMapper = (data, localRepositoryIds) => {
         // Sets the isRepositoryDeleted flag based on whether the repository ID is not in the list of repositories.
         isRepositoryDeleted: !localRepositoryIds.includes(data.repositoryId),
         model: data.model,
+        contextSize: data.api === 'openai-assistants' ? null : data.contextSize,
```
boyan-tonchev, line 177: IMO, it would be better to add an "api" field to the model and set it here. Then, instead of using:
`$scope.showContextSize = ($scope.agentFormModel.contextSize !== null);` which isn’t very descriptive, consider using something like:
$scope.showContextSize = $scope.agentFormModel.api === ApiName.OPENAI_ASSISTANTS;
The names are just examples; they can be something like agentApiName, etc.

e2e-tests/e2e-legacy/ttyg/edit-agent.spec.js
```diff
@@ -142,4 +142,50 @@ describe('TTYG edit an agent', () => {
...
+        // When I type a value
+        TtygAgentSettingsModalSteps.enterContextSize('120000');
+        // Then the error disappears
+        TtygAgentSettingsModalSteps.getContextSizeError().should('not.exist');
+        // I should be allowed to reset the value of the Context size
+        TtygAgentSettingsModalSteps.resetContextSizeValue();
+    });
+
+    it('should NOT show Context size if openai-assistants API', () => {
+        TTYGStubs.stubAgentListGet('/ttyg/agent/get-agent-list-autocomplete-query.json');
```
svilenvelikov, line 176: So, why are you repeating this over and over and not just add a specific method in the stub? Don't be afraid to do such minor refactorings.

packages/legacy-workbench/src/css/ttyg/agent-settings-modal.css
```diff
@@ -38,12 +38,53 @@
...
-    padding-left: 5px;
+.agent-settings-modal .agent-settings-form .context-size .input-with-unit {
+    position: relative;
+}
+
+.agent-settings-modal .agent-settings-form .context-size .input-with-unit input {
+    padding-right: 40px;
+}
+
+.agent-settings-modal .agent-settings-form .context-size .input-with-unit .token-unit-label {
```
svilenvelikov, line 49: Adding to the above comment, this is not the best approach for styling such field with text inside. Also I think that  using this class `input-group-addon` from bootstrap breaks some stuff here.

packages/legacy-workbench/src/js/angular/models/ttyg/agent-form.js
```diff
@@ -108,6 +128,7 @@ export class AgentFormModel {
             name: this._name,
             repositoryId: this._repositoryId,
             model: this._model,
+            contextSize: this._contextSize,
```
svilenvelikov, line 131: Shouldn't api field be added to the payload as well or it's  not part of the requets model?

packages/legacy-workbench/src/js/angular/ttyg/templates/modal/agent-settings-modal.html
```diff
@@ -58,7 +58,35 @@ <h4 class="modal-title">{{(
...
             </div>
-            <div class="form-group temperature col-md-4">
+            <div ng-if="showContextSize" class="form-group context-size col-md-3">
+                <div class="context-size-wrapper">
+                    <label for="contextSize">
+                        <span uib-popover="{{'ttyg.agent.create_agent_modal.form.context_size.tooltip' | translate}}"
+                              popover-trigger="mouseenter">{{'ttyg.agent.create_agent_modal.form.context_size.label' | translate}}</span>
+                    </label>
+
+                    <button type="button"
```
svilenvelikov, line 68: Why this button is rendered only when I hove over the field?

svilenvelikov (changes requested):

e2e-tests/e2e-legacy/ttyg/edit-agent.spec.js
DesiBorisova: Refactored the tests

packages/legacy-workbench/src/css/ttyg/agent-settings-modal.css
```diff
@@ -38,12 +38,53 @@
...
 }
 
-.agent-settings-modal .agent-settings-form .provider-name {
-    padding-left: 5px;
+.agent-settings-modal .agent-settings-form .context-size .input-with-unit {
+    position: relative;
+}
+
+.agent-settings-modal .agent-settings-form .context-size .input-with-unit input {
+    padding-right: 40px;
```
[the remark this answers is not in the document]
DesiBorisova: I edited the styling to fix this overlap

packages/legacy-workbench/src/css/ttyg/agent-settings-modal.css
```diff
@@ -38,12 +38,53 @@
...
 }
 
-.agent-settings-modal .agent-settings-form .provider-name {
-    padding-left: 5px;
+.agent-settings-modal .agent-settings-form .context-size .input-with-unit {
+    position: relative;
+}
+
+.agent-settings-modal .agent-settings-form .context-size .input-with-unit input {
+    padding-right: 3.5em;
```
svilenvelikov, line 46: This is OK now, because `tokens` in french is `jetons` and the space is enough for the word to fit in. The better approach would be to lay-out the content in a flex-ible way so that the engine to be able to adjust block dimensions appropriately.
You can leave it now, but pay attention to such things in future

svilenvelikov (approved):

PR #2254, opened 2025-07-31
GDB-11874 Fix page permissions

yordanalexandrov:
## What
Centralize page permissions and base them off of menu declarations

## Why
Page permissions were set based on 403 responses which were not always available on pages where the user does not have permission to view.

How
- added a function to set permissions based on menu item role for each defined route
- removed unauthorized interceptor, because it is no longer needed
- removed `$rootScope.deniedPermissions` and used `SecurityContextService` as single point of truth about permissions
- changed conditions in legacy template to allow the application to show `coreError` directive
- based showing the permission banner in layout on the `SecurityContextService` permissions

## Screenshots
* Admin user with no restrictions
image

* Repo manager with some restrictions
image

* User with more restrictions
image

* Free access user
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
+++ b/packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
@@ -421,9 +421,6 @@ repositories.service('$repositories', ['toastr', '$rootScope', '$timeout', '$loc
         this.onRepositorySet = function (newRepository) {
             this.setRepositoryHeaders(newRepository);
             $rootScope.$broadcast('repositoryIsSet', {newRepo: newRepository?.isNew});
-            ServiceProvider.get(SecurityContextService).updateRestrictedPages(new RestrictedPages());
-            // reset denied permissions (different repo, different rights)
-            $rootScope.deniedPermissions = {};
         }
 
         this.getDefaultRepository = function () {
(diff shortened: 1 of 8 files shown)


packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
```diff
@@ -421,9 +419,6 @@ repositories.service('$repositories', ['toastr', '$rootScope', '$timeout', '$loc
         this.onRepositorySet = function (newRepository) {
             this.setRepositoryHeaders(newRepository);
             $rootScope.$broadcast('repositoryIsSet', {newRepo: newRepository?.isNew});
-            ServiceProvider.get(SecurityContextService).updateRestrictedPages(new RestrictedPages());
```
plamen-yordanov, line 424: Pages are also reset upon [logout](https://github.com/Ontotext-AD/graphdb-workbench/blob/master/packages/shared-components/src/components/onto-layout/onto-layout.tsx#L191). That's not needed anymore and can be removed right?

plamen-yordanov (approved):

PR #2263, opened 2025-08-05
GDB-12378: All repositories are displayed by default

boyan-tonchev:
## What
All repositories are displayed by default even when "Show remote locations" is not clicked

## Why
After the introduction of GraphQL functionality, the template responsible for displaying repositories in the core error directive was changed. As a result, it started showing all repositories without filtering by location (local or remote).

## How
The directive was updated to use a method that filters the displayed repositories based on whether "Show remote locations" is selected.

## Screenshots
Before
imageimage
After
image image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
+++ b/packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
@@ -277,18 +277,12 @@ repositories.service('$repositories', ['toastr', '$rootScope', '$timeout', '$loc
             return this.getRepositories().find((repository) => repository.id === repositoryId);
         };
 
-        this.getReadableRepositories = function () {
-            return _.filter(this.getRepositories(), function (repo) {
-                return $jwtAuth.canReadRepo(repo);
-            });
-        };
-
         this.getReadableGraphdbRepositories = function () {
             return this.getReadableRepositories()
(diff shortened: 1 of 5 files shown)


packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
```diff
@@ -277,14 +277,8 @@ repositories.service('$repositories', ['toastr', '$rootScope', '$timeout', '$loc
...
 
-        this.getReadableRepositories = function () {
-            return _.filter(this.getRepositories(), function (repo) {
-                return $jwtAuth.canReadRepo(repo);
-            });
-        };
-
         this.getReadableGraphdbRepositories = function () {
-            return this.getReadableRepositories()
+            return this.getAllAccessibleRepositories()
```
yordanalexandrov, line 281: Although it is not done in this MR, the naming of the methods `getAllAccessibleRepositories` and `getAccessibleRepositories` is somewhat confusig.

By the name I would thing they do the same thing.

yordanalexandrov (approved):

PR #2265, opened 2025-08-06
GDB-12845: Add repository selection step in Repositories view

boyan-tonchev:
## What
Added a repository selection step in the Repositories view.

## Why
This enhances the guides with opportunity to teaching users that the active repository can be switched directly from the Repositories view.

## How
Updated the existing step, which was previously incomplete.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/legacy-workbench/src/pages/repositories.html
```diff
@@ -67,17 +67,21 @@ <h1>
...
+                                    ng-class="{'table-info text-secondary': isRepoActive(repository)}"
+                                    guide-selector="repository-{{repository.id}}">
+                                    <td class="select-repository-button" guide-selector="select-repository-button-wrapper">
                                         <span role="button" class="icon-connection-off icon-lg text-secondary m-0"
                                               gdb-tooltip="{{'connect.repo' | translate}}" tooltip-placement="top"
                                               ng-click="setRepository(repository)"
                                               ng-if="!isRepoActive(repository)"
                                               data-repository-id="{{repository}}"
-                                              guide-selector="repository-{{repository.id}}-plug"></span>
+                                              guide-selector="select-repository-button">
```
svilenvelikov, line 78: The parent element has a class `select-repository-button` and here you have again attribute with the same value. This is confusing.

boyan-tonchev: I can rename the parent class to match the guide selector `select-repository-button-wrapper.` What do you think?

yordanalexandrov (approved):

PR #2273, opened 2025-08-07
GDB-12852 Add steps to TTYG for SPARQL CONSTRUCT in IG

yordanalexandrov:
## What
Implement step for configuring agent using SPARQL extraction method and providing the ontology via a SPARQL CONSTRUCT query

## Why
Expand the interactive guides

How
- added a new core step `copy-text-element` which allows user to fill in an input by clicking a button
- added new steps for configuring agent using SPARQL extraction method and providing the ontology via a SPARQL CONSTRUCT query

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/legacy-workbench/src/js/angular/guides/steps/core/plugin.js
```diff
@@ -103,6 +110,51 @@ PluginRegistry.add('guide.step', [
...
+            let copyToEditorListener;
+
+            return [
+                {
+                    guideBlockName: 'input-element',
+                    options: {
+                        ...options,
+                        content: 'guide.step_plugin.core-steps.copy-text-element.content',
+                        textAsHtmlCodeElement: '<div class="shepherd-code">' + code.outerHTML + copy.outerHTML + '</div>',
+                        extraContent: options.queryExtraContent,
```
boyan-tonchev, line 139: IMO, `queryExtraContent` is not meaningful for the core step — I think it should just be `extraContent`. For more complex steps like `sparql-editor`, using a specific name like `queryExtraContent` makes sense. If we refactor that step to use the core step, we can map the option like this: `extraContent: options.queryExtraContent`. For the core step, we should keep it generic — `content` and `extraContent` — so they can be easily overridden.

packages/legacy-workbench/src/js/angular/guides/steps/core/plugin.js
```diff
@@ -103,6 +110,51 @@ PluginRegistry.add('guide.step', [
...
+            const GuideUtils = services.GuideUtils;
+
+            const text = options.text;
+            const code = document.createElement('code');
+            code.innerText = text;
+
+            const copy = document.createElement('button');
+            const copyToEditorButtonClass = 'guide-copy-to-editor-query-button';
+            copy.className = `btn btn-sm btn-secondary ${copyToEditorButtonClass}`;
+            copy.innerText = $translate.instant('guide.step_plugin.execute-sparql-query.copy-to-editor.button');
```
boyan-tonchev, line 126: 'guide.step_plugin.execute-sparql-query.copy-to-editor.button' is translated as "Copy to editor", but I think it should just be "Copy" or "Copy text", with support for overriding when necessary

packages/legacy-workbench/src/js/angular/guides/steps/core/plugin.js
```diff
@@ -103,6 +110,51 @@ PluginRegistry.add('guide.step', [
...
+            const $translate = services.$translate;
+            const GuideUtils = services.GuideUtils;
+
+            const text = options.text;
+            const code = document.createElement('code');
+            code.innerText = text;
+
+            const copy = document.createElement('button');
+            const copyToEditorButtonClass = 'guide-copy-to-editor-query-button';
+            copy.className = `btn btn-sm btn-secondary ${copyToEditorButtonClass}`;
```
boyan-tonchev, line 125: The class name should not be the same as the one used in the sparql-editor step.

packages/legacy-workbench/src/js/angular/guides/steps/core/plugin.js
```diff
@@ -103,6 +110,51 @@ PluginRegistry.add('guide.step', [
...
+                    options: {
+                        ...options,
+                        content: 'guide.step_plugin.core-steps.copy-text-element.content',
+                        textAsHtmlCodeElement: '<div class="shepherd-code">' + code.outerHTML + copy.outerHTML + '</div>',
+                        extraContent: options.queryExtraContent,
+                        show: (_guide) => () => {
+                            const elementSelector = GuideUtils.getElementSelector(options.elementSelector);
+                            const inputElement = document.querySelector(elementSelector);
+
+                            copyToEditorListener = createCopyToInputListener(inputElement, text);
```
boyan-tonchev, line 144: The listener should be initialized in the same scope where let copyToEditorListener; is declared. This avoids recreating the listener every time the step shows.

packages/legacy-workbench/src/js/angular/guides/steps/core/plugin.js
```diff
@@ -103,6 +110,51 @@ PluginRegistry.add('guide.step', [
...
+
+
+            let stepHTMLElement;
+            let copyToEditorListener;
+
+            return [
+                {
+                    guideBlockName: 'input-element',
+                    options: {
+                        ...options,
```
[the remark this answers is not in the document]
yordanalexandrov: moved it

packages/legacy-workbench/src/js/angular/guides/steps/core/plugin.js
yordanalexandrov: Changed to `Copy to input` as we discussed the behaviour.

I don't think this should be exposed for now.

PR #2275, opened 2025-08-07
GDB-12876 - Remove guide dialog class suffix

DesiBorisova:
## What
The steps class won't contain a hard-coded `-guide-dialog` suffix. The suffix will be added dynamically by the service.

## Why
As we are going to allow other users to define steps, the concern is that there might be class collisions.

## How
I removed the suffix from the class in the steps definition and automatically appended it in the service when the guide is built.

## Testing
N/A

## Screenshots
N/A

## Checklist
- [ ] Branch name
- [ ] Target branch
- [ ] Commit messages
- [ ] Squash commits
- [ ] MR name
- [ ] MR Description
- [ ] Tests


packages/legacy-workbench/src/js/angular/guides/steps/complex/select-repository/plugin.js
```diff
@@ -30,7 +30,7 @@ PluginRegistry.add('guide.step', [
                     skipPoint: true,
                     content: 'guide.step_plugin.choose-repository.content',
                     elementSelector: '.onto-repository-selector',
-                    class: 'repositories-group-button-guide-dialog',
+                    class: 'repositories-group',
```
boyan-tonchev, line 33: Was the  '-button' removed by mistake?

PR #2296, opened 2025-08-15
GDB-12885 - Extract table-graph agent steps

DesiBorisova:
## What
The `table-graph` core steps have been extracted.

## Why
Extracting the steps allows developers to reuse them.

## How
Steps extracted in the same file.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/legacy-workbench/src/js/angular/guides/steps/complex/resources/plugin.js
```diff
@@ -27,16 +146,15 @@ PluginRegistry.add('guide.step', [
...
+                {
+                    guideBlockName: 'resource-results-explain',
+                    options: {
                         content: 'guide.step_plugin.table-graph-overview',
-                        scrollToHandler: GuideUtils.scrollToTop,
-                        elementSelector: GuideUtils.CSS_SELECTORS.SPARQL_RESULTS_ROWS_SELECTOR,
-                        class: 'table-graph-overview',
-                        placement: 'top',
+                        skipUrl: true,
                         beforeShowPromise: () => GuideUtils.waitFor(`.resource-info a.source-link[href="${options.iri}"]`, 3)
```
yordanalexandrov, line 158: Why is this not extracted with the step?

boyan-tonchev: This can’t be moved to the extracted step because it is a specific case for the complex step. It is not relevant for other usages.

PR #2297, opened 2025-08-15
GDB-12936 workbench configurations

svilenvelikov:
## What
This PR introduces a configuration system that allows the workbench application to accept configurations via JSON files in the public assets folder, making the application more flexible and customizable.

## Why
Unlocks the possibility to configure certain asspects of the application through external configuration file.

## How
Adds configuration services to load and manage application settings from JSON files.
Implements a two-tier configuration system with environment files and configuration files.
Refactors plugin loading to use configurable manifest paths instead of hardcoded values.

## Testing
Unit tests added.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/error/error-base.ts
+++ b/packages/api/src/error/error-base.ts
@@ -1,0 +1,9 @@
+export abstract class ErrorBase extends Error {
+  public constructor(private text?: string) {
+    super();
+  }
+
+  public get message(): string {
+    return `${this.constructor.name}: ${this.text}`;
+  }
+}
--- a/packages/api/src/services/configuration/configuration-rest.service.ts
+++ b/packages/api/src/services/configuration/configuration-rest.service.ts
@@ -1,0 +1,32 @@
+import {HttpService} from '../http/http.service';
+import {Configuration, Environment} from '../../models/configuration';
+import envDefault from './env.default.json';
+import {MissingApplicationConfigurationError} from './missing-application-configuration-error';
+
+const ENV_JSON_URL = 'assets/env.json';
+
+export class ConfigurationRestService extends HttpService {
+  async getConfiguration(): Promise<Configuration | undefined> {
+    let environment: Environment = envDefault as Environment;
+    try {
+      environment = await this.get(ENV_JSON_URL) as Environment;
+    } catch {
+      console.warn('No downloadable "env.json" was found. Fallback to default:', environment);
+    }
+
+    let configResponse;
+    try {
+      configResponse = await this.get(environment.configUrl);
+      if (configResponse) {
+        console.debug(`Configuration file ${environment.configUrl} was downloaded`, configResponse);
+        return configResponse as Configuration;
+      }
+    } catch {
+      throw new MissingApplicationConfigurationError(
+        `Failed to download configuration from "${environment.configUrl}". Ensure that there is a
+        "configuration.default.json" JSON file in the assets folder or provide an "env.json" file pointing to another
+         configuration file.`
+      );
+    }
+  }
+}
--- a/packages/api/src/services/plugins/plugins-rest.service.ts
+++ b/packages/api/src/services/plugins/plugins-rest.service.ts
@@ -1,14 +1,14 @@
 import {HttpService} from '../http/http.service';
 import {PluginModule, PluginsManifest, PluginsManifestResponse} from '../../models/plugins';
 import {getOrigin} from '../utils';
-
-// TODO: move this to a configuration file or environment variable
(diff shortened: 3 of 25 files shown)


packages/api/src/services/configuration/configuration-rest.service.ts
```diff
@@ -0,0 +1,32 @@
...
+import envDefault from './env.default.json';
+import {MissingApplicationConfigurationError} from './missing-application-configuration-error';
+
+const ENV_JSON_URL = 'assets/env.json';
+
+export class ConfigurationRestService extends HttpService {
+  async getConfiguration(): Promise<Configuration | undefined> {
+    let environment: Environment = envDefault as Environment;
+    try {
+      environment = await this.get(ENV_JSON_URL) as Environment;
```
teodossidossev, line 12: what if the json is presengt but corrupt?

packages/api/src/services/plugins/plugins-rest.service.ts
```diff
@@ -17,7 +16,7 @@ export class PluginsRestService extends HttpService {
    * information about available plugins.
    */
   getPluginsManifest(): Promise<PluginsManifestResponse> {
-    return this.get(MANIFEST_URL);
+    return this.get(ServiceProvider.get(ConfigurationContextService).getApplicationConfiguration().pluginsManifestPath);
```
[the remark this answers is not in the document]
yordanalexandrov: Maybe we can expose an `inject` function like in Angular which will do `ServiceProvider.get(service)`, but with 13 less chars.

packages/api/src/services/plugins/plugins.service.ts
```diff
@@ -24,7 +24,7 @@ export class PluginsService implements Service {
    * @returns A promise that resolves to the plugins manifest containing information about available plugins.
    */
   async getPluginsManifest(): Promise<PluginsManifest> {
```
yordanalexandrov, line 26: no need for the async either

packages/api/src/error/error-base.ts
```diff
@@ -0,0 +1,9 @@
+export abstract class ErrorBase extends Error {
+  public constructor(private text?: string) {
+    super();
```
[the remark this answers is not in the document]
svilenvelikov: I intentionally didn't do it. Overriding the message getter does kind of the same but shortens the stacktrace

packages/api/src/services/configuration/configuration-rest.service.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    let configResponse;
+    try {
+      configResponse = await this.get(environment.configUrl);
+      if (configResponse) {
+        console.log(`Configuration file ${environment.configUrl} was downloaded`, configResponse);
+        return configResponse as Configuration;
+      }
+    } catch {
+      throw new MissingApplicationConfigurationError(
```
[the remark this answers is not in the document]
svilenvelikov: image

packages/api/src/services/plugins/plugins-rest.service.ts
boyan-tonchev, line 19: IMO, this method should not be here because it does not call the BE.
I think it would be better if we map the configuration response in `ConfigurationService.getConfiguration` and update the `applicationConfiguration` of `ConfigurationContextService` with the UI model.

The configuration mapper function can use a concrete mapper for a specific configuration (for now: `pluginsManifestMapper.mapToModel`).

Then, in the code, we can subscribe to or synchronously get the configuration through the context. In that case, the `getPluginsManifest` method in `PluginsService` can either be removed or re-implemented to use `ConfigurationContextService` with mapped data, instead of mapping it every time it is called.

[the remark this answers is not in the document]
svilenvelikov: > Maybe we can expose an inject function like in Angular which will do ServiceProvider.get(service), but with 13 less chars.

Great idea. I'll introduce such function, and use it here, but I'll not touch other places to refactor the rest of the code.

svilenvelikov: This comment is invalid. The getPluginsManifest method retrieves the `pluginsManifestPath` from the configuration context and then makes a GET request to fetch the plugins manifest JSON file from that location.

boyan-tonchev: My, fault. I missed `this.get`, which is the REST call. Ignore the comment..

PR #2299, opened 2025-08-18
GDB-12500 - Save selected agent to local store and select first in dropdown

DesiBorisova:
## What
The first agent in the TTYG agent select dropdown will be selected on page load. If the user selects another agent, it will be remembered the next time the page is loaded.

## Why
The improvement reduces steps needed to be taken in order to ask a question.

## How
I save the user selected agent to the local store, otherwise if local store is empty, I load the first option from the dropdown. When a chat is loaded, the last agent used in the chat will be selected, as per the original functionality, regardless what is in local store.

## Testing
Test expectations edited to fit the default first option selected.

## Screenshots
N/A

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/ttyg/directives/agent-select-menu.directive.js
+++ b/packages/legacy-workbench/src/js/angular/ttyg/directives/agent-select-menu.directive.js
@@ -8,9 +8,9 @@ angular
     .module('graphdb.framework.ttyg.directives.agent-select-menu', modules)
     .directive('agentSelectMenu', AgentSelectMenuComponent);
 
-AgentSelectMenuComponent.$inject = ['TTYGContextService', '$translate', '$sce', 'ModalService'];
+AgentSelectMenuComponent.$inject = ['TTYGContextService', '$translate', '$sce', 'ModalService', 'TTYGStorageService'];
 
-function AgentSelectMenuComponent(TTYGContextService, $translate, $sce, ModalService) {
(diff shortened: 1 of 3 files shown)


svilenvelikov:
Please, change the target branch to workbench-3.2

packages/legacy-workbench/src/js/angular/ttyg/directives/agent-select-menu.directive.js
```diff
@@ -83,8 +84,33 @@ function AgentSelectMenuComponent(TTYGContextService, $translate, $sce, ModalSer
...
+             * if no recent agent is available.
+             *
+             * @function
+             */
+            const selectLastUsedOrDefaultAgent = () => {
+                let selectedAgent;
+
+                // If saved in the LocalStore, the user has selected it, so select it here
+                $scope.agentOptionsList.forEach((agent) => {
+                    if (agent.data.agent.id === TTYGStorageService.getAgentId()) {
```
svilenvelikov, line 101: Get the agent id once, outside of the for loop

yordanalexandrov: Took it out and added a check if defined.

svilenvelikov (approved):

yordanalexandrov (approved):

PR #2301, opened 2025-08-20
GDB-12891 Allow user to scroll and read TTYG explanation

plamen-yordanov:
## What
Allow the user to scroll and see the explanation result in TTYG, while on the guide

## Why
It's currently not possible, as the chat panel is readonly. This results in the user not being able to see the explanation

## How
- Removed the readonly option for `hold-and-wait-until-hidden` steps
- Added events listeners for click and double-click on `show` and `hide`, which prevent user interaction. This way the user can scroll, but can't trigger anything else (click other buttons) and possibly break the guide.
 
## Testing
n/a

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/guides/steps/core/plugin.js
+++ b/packages/legacy-workbench/src/js/angular/guides/steps/core/plugin.js
@@ -42,6 +42,37 @@ const createCopyToInputListener = (elementSelector, text) => {
     }
 }
 
+/**
+ * Configures an element interactability, by consuming events and preventing them from propagating.
+ * This allows to keep scrolling, while disallowing interaction with other elements
+ * (such as clicking buttons).
+ * @param interactable - true to make the element interactable, false to make it non interactable
+ * @param elementSelector - the elementSelector
+ * @param services - The services object
+ */
+const _configureInteractions = (interactable, elementSelector, services) => () => {
+    if (!elementSelector) {
+        return;
+    }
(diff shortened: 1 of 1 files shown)


packages/legacy-workbench/src/js/angular/guides/steps/core/plugin.js
```diff
@@ -42,6 +42,34 @@ const createCopyToInputListener = (elementSelector, text) => {
...
 
+/**
+ * Makes an element readonly, by consuming events and preventing them from propagating.
+ * This allows to keep scrolling, while disallowing interaction with other elements
+ * (such as clicking buttons).
+ * @param enable - true to make the element readonly, false to make it interactable
+ * @param options - the options object containing elementSelector
+ * @param services - The services object
+ */
+const configureReadonly = (enable, options, services) => () => {
```
yordanalexandrov, line 53: I think a more suitable name would be `preventInteractions` and pass the element selector as an argument insted of the whole `options` object

plamen-yordanov: Added `enableInteractions` and `disableInteractions`. Renamed `configureReadonly` to `_configureInterations`, so it can be reused. Passed `elementSelector` instead of the whole `options` object

yordanalexandrov (approved):

PR #2319, opened 2025-08-26
GDB-12980: Change default scrollTo strategy to prevent screen flashing

boyan-tonchev:
## What
Changed the default scrollTo configuration to use a smoother and less disruptive alignment.

## Why
The previous strategy scrolled elements to the top of the viewport, which often caused the screen to flash or jump abruptly. This degraded the user experience during guides.

## How
- Updated `scrollTo.behavior` to smooth for natural animated scrolling;
- Updated block and inline to nearest, ensuring minimal movement while keeping the element visible.

## Before
image

#### After click "Next" button

image

## After
image

#### After click "Next" button

image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/guides/tour-lib-services/shepherd.service.js
+++ b/packages/legacy-workbench/src/js/angular/guides/tour-lib-services/shepherd.service.js
@@ -5,6 +5,7 @@ import {
     navigate,
     getPathName
 } from '@ontotext/workbench-api';
+import {NEAREST, SMOOTH} from "../model/guides";
 
 export const GUIDE_ID = 'shepherd.guide_id';
 export const GUIDE_CURRENT_STEP_ID = 'shepherd.guide.current.step.id';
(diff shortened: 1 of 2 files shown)


packages/legacy-workbench/src/js/angular/guides/tour-lib-services/shepherd.service.js
```diff
@@ -212,7 +215,11 @@ function ShepherdService($translate, LocalStorageAdapter, LSKeys, $interpolate,
             confirmCancel: this._confirmGuideCancel,
             defaultStepOptions: {
                 classes: 'shadow-md bg-purple-dark',
-                scrollTo: true
+                scrollTo: {
```
svilenvelikov, line 218: I know that we can't use the typescript types in the legacy, but still I don't like having these 3 essentially duplicated types for these constants. I'd used just constant strings instead

svilenvelikov (dismissed):

packages/legacy-workbench/src/js/angular/guides/tour-lib-services/shepherd.service.js
boyan-tonchev: Done

svilenvelikov (approved):

PR #2332, opened 2025-08-28
feat(GDB-12742) Respect return URL on repo create

teodossidossev:
## WHAT
- Make repo create flow return to the actual originating route via previous query param.

## WHY
Users were sent to generic paths after creating/canceling, instead of going back to where they started—hurting UX.

## HOW
- goToAddRepo - capture the current $location.url() and pass it as previous when navigating to repository/create
- goBackToPreviousLocation - read previous, remove it, navigate to it, else default to /repository
- Add trailing commas in arrays/objects/imports, ensure semicolons, normalize function() spacing, replace let with const where applicable and remove stray blank lines across touched files.

## Checklist
- [X] Branch name
- [X] Target branch
- [X] Commit messages
- [X] Squash commits
- [X] MR name
- [X] MR Description
- [ ] Tests

### DIFF
--- a/e2e-tests/cypress-legacy.config.js
+++ b/e2e-tests/cypress-legacy.config.js
@@ -1,8 +1,15 @@
-const {defineConfig} = require('cypress');
+import { defineConfig } from 'cypress';
+import setupPlugins from './plugins/index.js';
 
 const isCoverage = process.env.COVERAGE === 'true';
 
-module.exports = defineConfig({
+const loadCodeCoverage = async (on, config) => {
+    const mod = await import('@bahmutov/cypress-code-coverage/plugin');
+    const plugin = ('default' in mod) ? mod.default : mod;
+    plugin(on, config);
+};
+
+export default defineConfig({
     projectId: 'v35btb',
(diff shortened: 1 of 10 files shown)


plamen-yordanov (approved):

e2e-tests/cypress-legacy.config.js
```diff
@@ -1,8 +1,17 @@
-const {defineConfig} = require('cypress');
+import { defineConfig } from 'cypress';
+import setupPlugins from './plugins/index.js';
 
+/* eslint-disable no-undef */
```
plamen-yordanov, line 4: instead of `/*eslint-disable no-undef*/`,  `process` can be added to the `languageOptions.globals` eslint.config

e2e-tests/cypress-legacy.config.js
```diff
@@ -1,8 +1,17 @@
...
 const isCoverage = process.env.COVERAGE === 'true';
+/* eslint-enable no-undef */
```
plamen-yordanov, line 6: I'm assuming this won't be needed as well

PR #2365, opened 2025-09-17
GDB-11430 Replace console with logger service

plamen-yordanov:
## What
Replace all instances of `console` logging with the centralized `LoggingService`

## Why
To use the central logging service across the entire application

## How
Replaced console usage with the new `LoggerService`

## Testing
n/a

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/shared-components/src/components/onto-navbar/navbar-service.ts
+++ b/packages/shared-components/src/components/onto-navbar/navbar-service.ts
@@ -1,6 +1,9 @@
 import {ExternalMenuItemModel, ExternalMenuModel} from './external-menu-model';
 import {NavbarItemModel, NavbarModel} from './navbar-model';
 import {ProductInfo, UriUtil} from '@ontotext/workbench-api';
+import {LoggerProvider} from '../../services/logger-provider';
+
+const logger = LoggerProvider.logger;
 
 export class NavbarService {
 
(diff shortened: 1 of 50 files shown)


packages/shared-components/src/components/onto-navbar/navbar-service.ts
```diff
@@ -17,7 +19,7 @@ export class NavbarService {
         .filter((item) => !item.parent && item.shouldShow)
         .forEach((item) => {
           if (navbarModel.hasParent(item.label)) {
-            console.warn('Doubled parent definition: ', item);
+            NavbarService.logger.warn('Doubled parent definition: ', item);
```
svilenvelikov, line 22: umm what is this NavbarService?

packages/shared-components/src/components/onto-navbar/navbar-service.ts
```diff
@@ -1,8 +1,10 @@
 import {ExternalMenuItemModel, ExternalMenuModel} from './external-menu-model';
 import {NavbarItemModel, NavbarModel} from './navbar-model';
 import {ProductInfo, UriUtil} from '@ontotext/workbench-api';
+import {LoggingService} from '../../services/logging.service';
 
 export class NavbarService {
+  private static readonly logger = LoggingService.logger;
```
svilenvelikov, line 7: I'm not sure if we should tie the logger to such utilities as a static property.

svilenvelikov: As far as our loggers are singletons, maybe it is OK.

plamen-yordanov: I was playing around with copilot agent and it wrote it like this. Replaced with a file const

svilenvelikov (approved):

PR #2380, opened 2025-09-25
GDB-13062: Decouple guide translations from Workbench language bundles

boyan-tonchev:
## What
Expose a new utility function translate to be used in guide steps.

## Why
Guide step translation bundles should be decoupled from the Workbench. Each step will now provide its own bundle with translations.

## How
A new utility function has been added. It can be used in guide steps to translate their content. When called, the function will translate the given key with parameters using the step’s bundle. The locale used will be the one currently selected in the Workbench.

### Additional work
The Welcome guide step translation bundles have been removed, since they are now managed within the step itself.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/legacy-workbench/src/js/angular/guides/guides.service.js
```diff
@@ -181,17 +183,84 @@ function GuidesService(
...
+     * @param {Object<string, string>} parameters - Optional parameters to interpolate into the translation string
+     *   (e.g., `{ username: "John" }`).
+     * @returns {string} The translated string, or the key if no translation is available.
+     */
+    this.translate = (translationBundle = {}, key, parameters = {}) => {
+        let translation = translationBundle[this.currentLanguage]?.[key];
+
+        if (!translation) {
+            // Fallback to the default language
+            translation = translationBundle['en']?.[key];
```
plamen-yordanov, line 224: It would be better to have a config file, instead of hardcoding the default

packages/legacy-workbench/src/js/angular/guides/guides.service.js
```diff
@@ -533,4 +603,11 @@ function GuidesService(
     this._subscribeToGuidePause = () => {
         ShepherdService.subscribeToGuidePause(() => $rootScope.$broadcast('guidePaused'));
     };
+
+    this._subscribeToLanguageChange = () => {
+        ServiceProvider.get(LanguageContextService)
```
plamen-yordanov, line 608: you can use `service` here instead of `ServiceProvider`

plamen-yordanov (approved):

svilenvelikov (approved):

PR #2385, opened 2025-09-26
GDB-13146: Can't return to dataset download step if skipped, requiring restart

boyan-tonchev:
## What
Add a Previous button to the import-upload-rdf-file step.

## Why
In this step, the user must upload a file that was downloaded in the previous step. If they skip that step, they currently have to stop the guide and restart it. With the Previous button, the user can go back to the previous step to download the file without restarting the guide.

## How
Update the import-upload-rdf-file step to display a Previous button.

## Additional work
Added a check to ensure that a step has a defined url before navigating to it in the previous flow, since missing URLs caused the guide to crash.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/guides/steps/complex/import-rdf-file/plugin.js
+++ b/packages/legacy-workbench/src/js/angular/guides/steps/complex/import-rdf-file/plugin.js
@@ -18,7 +18,7 @@ PluginRegistry.add('guide.step', [
                         content: 'guide.step_plugin.import_rdf_file.content',
                         class: 'upload-rdf-file-button',
                         // If mainAction is set the title will be set automatically
-                        ...(options.mainAction ? {} : { title: IMPORT_FILE_DEFAULT_STEP_TITLE }),
+                        ...(options.mainAction ? {} : {title: IMPORT_FILE_DEFAULT_STEP_TITLE}),
                         ...options,
                         url: 'import',
                         elementSelector: GuideUtils.getGuideElementSelector('uploadRdfFileButton'),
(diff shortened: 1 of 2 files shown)


packages/legacy-workbench/src/js/angular/guides/steps/complex/import-rdf-file/plugin.js
```diff
@@ -78,14 +78,15 @@ PluginRegistry.add('guide.step', [
                                 })
                                 .catch((error) => {
                                     // This shouldn't be happening.
+                                    // eslint-disable-next-line no-console
```
plamen-yordanov, line 81: this should be `console.error`, which is allowed by lint

packages/legacy-workbench/src/js/angular/guides/steps/complex/import-rdf-file/plugin.js
```diff
@@ -174,11 +175,13 @@ PluginRegistry.add('guide.step', [
                           return GuideUtils.waitFor('.import-resource-message', 10);
                       },
                       onPreviousClick: () => GuideUtils.getOrWaitFor(GuideUtils.getGuideElementSelector('import-file-' + options.resourceFile), 10)
-                          .then((element) => { element.click() })
-                  }
-              }
-          ]
-      }
+                          .then((element) => {
+ element.click();
```
plamen-yordanov, line 179: could you check this indentation?

PR #2404, opened 2025-10-02
GDB-12871: extend the plugin registry to support overriding

boyan-tonchev:
## What
Modify the plugin registry to allow plugins to override previously registered plugins with the same unique ID.

## Why
Overriding is necessary to allow updates or replacements of plugins dynamically. This improves flexibility and prevents duplicate plugin registrations in the same extension point.

## How
Add logic in `registerPlugin` to remove any existing plugin with the same ID before registering a new one.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/legacy-workbench/src/js/angular/plugin-registry.js
```diff
@@ -52,37 +51,66 @@ window.PluginRegistry = (function() {
         return typeof variable !== 'undefined';
     }
 
-    function registerPlugin(extensionPoint, pluginDefinition) {
-        if (pluginDefinition.disabled) {
-            return;
-        }
+    function getExtensionPointIdFieldName(extensionPointName) {
+      if ('guide.step' === extensionPointName) {
```
svilenvelikov, line 55: umm what? The plugin registry should be unaware of the concrete extension points

boyan-tonchev: Maybe I didn’t understand you, but we decided to make the unique identifier abstract and dependent on the extension point. In the context of guides, it is `guideBlockName`. I’ll think of another way to implement it.

packages/api/src/services/plugins/plugin-registry.ts
```diff
@@ -0,0 +1,132 @@
...
+  private readonly extensionPoints: Record<string, ExtensionPoint<Plugin> | ExtensionPoint<OrderedPlugin>> = {};
+
+  /**
+   * Initializes all predefined extension points.
+   */
+  constructor() {
+    this.extensionPoints[ExtensionPointName.MAIN_MENU] = new MainMenuExtensionPoint();
+    this.extensionPoints[ExtensionPointName.ROUTE] = new RouteExtensionPoint();
+    this.extensionPoints[ExtensionPointName.INTERACTIVE_GUIDE] = new InteractiveGuideExtensionPoint();
+    this.extensionPoints[ExtensionPointName.THEMES] = new ThemesExtensionPoint();
```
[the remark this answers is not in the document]
boyan-tonchev: Done.

PR #2407, opened 2025-10-03
GDB-12766 support dark and light modes

svilenvelikov:
## What
Adds support for dark and light modes to the styleguide.

## Why
The workbench is currently supporting light and dark modes but the dark mode is emulated artificially, so we need a proper dark mode theme.

## How
Updating the build process to generate css files for light and dark mode.

Replaced single optimized CSS variables with separate light and dark mode CSS files
Created a new purge script to generate both light-mode.css and dark-mode.css from upstream styleguide files
Updated the main styleguide entry point to import both theme files

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/root-config/src/vendor/variables.css
svilenvelikov, line 1: Temporarily commented these out. The dark theme doesn't look good with the new dark mode stylesheet but the problems will be resolved with a separate MR.

PR #2414, opened 2025-10-08
GDB-13194 fix guide breaking, when api key is not set

plamen-yordanov:
## What
Fix the guide breaking, when api key is not set

## Why
Because it was breaking unexpectedly

## How
- Replaced `text()` with `textContent`, as the element is no longer angular wrapped
- Added the new property for api key to the validation. The `graphdb.openai.api-key` is deprecated and now the error message contains `graphdb.llm.api-key` instead. Left the old value as well just in case
- Also fixed a `console.error` message, when trying to get saved chats.

## Testing
cypress

## Checklist
- [ ] Branch name
- [ ] Target branch
- [ ] Commit messages
- [ ] Squash commits
- [ ] MR name
- [ ] MR Description
- [ ] Tests


boyan-tonchev (approved):

packages/legacy-workbench/src/js/angular/guides/steps/complex/ttyg/common/end-on-api-key-error/plugin.js
```diff
@@ -1,10 +1,19 @@
+const GRAPHDB_API_KEY_PROPERTIES = [
+    'graphdb.llm.api-key',
+    // deprecated
+    'graphdb.openai.api-key',
+];
+
+const isApiKeyError = (text) => {
+    return GRAPHDB_API_KEY_PROPERTIES.some((key) => text.includes(key));
```
svilenvelikov, line 8: This is not cool, but I can't think of another way ATM.

PR #2522, opened 2025-11-27
GDB-13418 sync repository deletion/creation from another tab

plamen-yordanov:
## What
Implement synchronization of repository deletion/creation across different tabs

## Why
Because currently there is no such synchronization. When creating/deleting repositories, the changes are not reflected in other tabs from the same origin.

## How
- Added a `BroadcastService` to send `BroadcastChannel` messages to other browser contexts
- Added `broadcast-message` as a unified model of the message to be sent
- Added a subscription in the main `controller.js`, which listens for `MessageType.REPOSITORIES_UPDATED` and triggers `$repositories.init()`
- Added event dispatch on create/delete repository actions

## Testing
n/a

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/api/src/services/broadcast/broadcast.service.ts
+++ b/packages/api/src/services/broadcast/broadcast.service.ts
@@ -1,0 +1,40 @@
+import {BroadcastMessage, MessageHandler} from '../../models/broadcast';
+import {Subscription} from '../../models/common';
+
+/**
+ * A service that facilitates communication between different browser contexts
+ * (such as tabs, windows, or iframes) of the same origin.
+ * It uses the `BroadcastChannel` API to send and receive messages.
+ */
+export class BroadcastService {
+  private readonly CHANNEL_NAME = 'ontotext-workbench-broadcast';
+  private readonly MESSAGE_EVENT = 'message';
+
+  private channel?: BroadcastChannel;
+
+  /**
+   * Sends a message to all listening browser contexts on the same origin.
+   * @param {BroadcastMessage} message The message to be sent.
+   */
+  sendMessage(message: BroadcastMessage): void {
+    this.getChannel().postMessage(message);
+  }
+
+  /**
+   * Subscribes to incoming messages.
+   * @param {MessageHandler} callback The function to be executed when a message is received.
+   * @returns {Subscription} An unsubscribe function.
+   */
+  subscribeToMessages(callback: MessageHandler): Subscription {
+    const channel = this.getChannel();
+    const listener = (event: MessageEvent) => callback(event.data);
+
+    channel.addEventListener(this.MESSAGE_EVENT, listener);
+    return () => channel.removeEventListener(this.MESSAGE_EVENT, listener);
+  }
+
+  private getChannel(): BroadcastChannel {
+    this.channel ??= new BroadcastChannel(this.CHANNEL_NAME);
+    return this.channel;
+  }
+}
--- a/packages/legacy-workbench/src/js/angular/controllers.js
+++ b/packages/legacy-workbench/src/js/angular/controllers.js
@@ -43,6 +43,8 @@ import {
     service,
     ShepherdService,
     WindowService,
+    BroadcastService,
+    MessageType,
 } from '@ontotext/workbench-api';
 import {EventConstants} from './utils/event-constants';
 import {CookieConsent} from './models/cookie-policy/cookie-consent';
@@ -1137,6 +1139,12 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, $location, $repositories,
     const licenseUpdatedSubscription = service(LicenseContextService).onLicenseChanged(onLicenseUpdated);
     const cookieConsentChangedSubscription = subscribeToCookieConsentChanged();
 
+    const broadcastUnsubscribe = service(BroadcastService).subscribeToMessages((message) => {
+        if (message.type === MessageType.REPOSITORIES_UPDATED) {
+            $repositories.init();
+        }
+    });
+
     // TODO: The $destroy hook is not called in the workbench unmounting process, so these subscriptions are not cleaned up.
     $scope.$on('$destroy', () => {
         onSelectedRepositoryChangedSubscription?.();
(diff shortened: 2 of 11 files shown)


svilenvelikov:
@plamen-yordanov Wait with this change until Monday please. It turns out to be a feature implementation instead of bugfix.

packages/api/src/services/broadcast/broadcast.service.ts
```diff
@@ -0,0 +1,43 @@
...
+  sendMessage(message: BroadcastMessage): void {
+    this.getChannel().postMessage(message);
+  }
+
+  /**
+   * Subscribes to incoming messages.
+   * @param {MessageHandler} callback The function to be executed when a message is received.
+   * @returns {() => void} An unsubscribe function.
+   */
+  onMessageReceived(callback: MessageHandler): () => void {
```
yordanalexandrov: We have a `Subscription` type which can be returned here

packages/api/src/services/broadcast/broadcast.service.ts
```diff
@@ -0,0 +1,43 @@
+import {BroadcastMessage} from '../../models/broadcast';
+
+type MessageHandler = (message: BroadcastMessage) => void;
```
yordanalexandrov: This should probably be with the other models

packages/api/src/services/broadcast/broadcast.service.ts
yordanalexandrov: Why not name it `subscribe` as in other services with similar methods?

packages/legacy-workbench/src/js/angular/repositories/controllers.js
```diff
@@ -708,7 +711,10 @@ function AddRepositoryCtrl($rootScope, $scope, toastr, $repositories, $location,
                 toastr.success($translate.instant('created.repo.success.msg', {repoId: $scope.repositoryInfo.id}));
                 return $repositories.init();
             })
-            .then(() => $scope.goBackToPreviousLocation())
+            .then(() => {
+                $scope.goBackToPreviousLocation();
+                service(BroadcastService).sendMessage(new BroadcastMessage(MessageType.REPOSITORIES_UPDATED));
```
yordanalexandrov: Edit is also affected, as you can rename a graphdb repository

packages/legacy-workbench/src/js/angular/controllers.js
```diff
@@ -1137,6 +1139,12 @@ function mainCtrl($scope, $menuItems, $jwtAuth, $http, $location, $repositories,
     const licenseUpdatedSubscription = service(LicenseContextService).onLicenseChanged(onLicenseUpdated);
     const cookieConsentChangedSubscription = subscribeToCookieConsentChanged();
 
+    const broadcastUnsubscribe = service(BroadcastService).onMessageReceived((message) => {
+        if (message.type === MessageType.REPOSITORIES_UPDATED) {
```
yordanalexandrov, line 1143: Will this trigger `init` in the current context or it will skip it as it is the emitter?

packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
```diff
@@ -437,6 +442,7 @@ repositories.service('$repositories', ['toastr', '$rootScope', '$timeout', '$loc
             return RepositoriesRestService.deleteRepository(repo)
                 .success(function() {
                     that.init();
+                    service(BroadcastService).sendMessage(new BroadcastMessage(MessageType.REPOSITORIES_UPDATED));
```
yordanalexandrov: I am wondering whether it will be better to have the message emitted in the `init` method.
Instead of having it at multiple places.

Perhaps after some sort of diff.

packages/legacy-workbench/src/js/angular/controllers.js
plamen-yordanov: The emitting context does not get the message it sent. If we want that for some reason, we need to add some additional handling

packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
plamen-yordanov: Too risky IMO. Having an emit there can lead to infinite loop between tabs emitting and `init`-ing.

packages/api/src/services/broadcast/broadcast.service.ts
plamen-yordanov: replaced

packages/api/src/services/broadcast/broadcast.service.ts
plamen-yordanov: extracted

packages/api/src/services/broadcast/broadcast.service.ts
plamen-yordanov: I don't remember, why I chose this name, since it was a while ago, but I renamed it

packages/legacy-workbench/src/js/angular/repositories/controllers.js
plamen-yordanov: Right. The improvement only mentions deletion, but syncing only deletions and not creation and edit seems like work half done

yordanalexandrov: Will you do it then?

packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
```diff
@@ -427,6 +430,8 @@ repositories.service('$repositories', ['toastr', '$rootScope', '$timeout', '$loc
                         that.setRepository('');
                     }
                     that.init();
+                    // Deleting a location removes its repositories from the list
+                    service(BroadcastService).sendMessage(new BroadcastMessage(MessageType.REPOSITORIES_UPDATED));
```
yordanalexandrov: Extract this `service(BroadcastService)` at the top with the other services

packages/legacy-workbench/src/js/angular/repositories/controllers.js
plamen-yordanov: I was under the false impression, that this scenario was covered, but I was on the repositories page. Added emit on edit as well.

packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
plamen-yordanov: done

packages/legacy-workbench/src/js/angular/core/services/repositories.service.js
yordanalexandrov: ok, leave it like that and we will think about it when rewriting the repositories

yordanalexandrov (approved):

PR #2873, opened 2026-04-06
GDB-14399 migrate yasgui facade directive

svilenvelikov:
## What
Migrate the yasgui-component.directive.js to a new facade component in the new workbench application.

## Why
This is part of the sparql view migration to the new workbench application.

## How
* Implemented loadSavedQueries event handler and configuration.
* Fixed the default for the getCellContent configuration.
* Implemented save query event handler.
* Fixed issue in http service that doesn't propagate the http params for the PUT, PATCH and DELETE request methods.
* Implemented updateQuery method in the sparql service.
* Implemented update query event handler in yasgui component facade.
* Implemented delete saved query event handler in yasgui component facade.
* Implemented share query event handler in yasgui component facade.
* Implemented share saved query event handler in yasgui component facade.
* Implemented query copied event handler in yasgui component facade.
* Implemented the generic Output event handler. Refactor the generic output channel wiring and typing to be a bit simpler and typesafe.
* Migrated countQueryRequestHandler

* implement save query opened event handler
* Output handlers
* Refactor the generic output channel wiring and typing to be a bit simpler and typesafe.
* Migrate countQueryRequestHandler.
* Implemented countQueryResponseHandler.
* Download as handlers implementation.
* Extend HttpService to support blob response type.
* Implemented downloadResultsAsFile in the rdf4j-repository service. This is used for download from server operations.
* Handle downloadAs event to support download operations in the sparql view.
* Handle notificationMessageEvent to support custom notification messages triggered directly from the yasgui component.
* Fix the specific sparql editor page css rules not being properly applied to yasgui component due to view encapsulation.
* Remove the fixed 100vh height from the main container applied in the new workbench. This now appears to make the container too high forcing unnecessary vertical scroll at times.
* Extend the http service with the ability to pass an abort signal to allow request aborting.
* Extend the autocomplete.service.ts to provide optional abort signal to the rest client. This is needed for the yasgui facade component where the namespaces autocomplete is using it to cancel previous ongoing requests.
* Implemented namespaces autocomplete handler.
* Applied min-height to codemirror similar to how it is done in legacy in order to have the keyboard shortcuts button visible again when sparql view is opened directly and the legacy's bootstrap-graphdb-theme stylesheet where the override is, is not present.
* Extend the init function with onAbort handler and keyboard shortcuts config.
* Yasgui persistence migration is migrated.
* Implemented addKnownPrefixes method in the sparql service plus tests. This would be used to populate the prefixes in a sparql query pasted by the user where not all prefixes are present.
* Implemented the paste and keyup handlers in the yasgui.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/workbench/src/app/components/yasgui-component-facade/models/event/ontotext-yasgui-event.ts
```diff
@@ -0,0 +1,31 @@
...
+import {QueryExecutedEvent} from './query-executed-event';
+import {SaveQueryOpened} from './save-query-opened';
+import {RequestAbortedEvent} from './request-aborted-event';
+
+/**
+ * This interface represents a custom event emitted by the Yasgui component in the Ontotext Workbench application.
+ */
+export interface OntotextYasguiEvent extends CustomEvent {
+  type: EventDataType;
+  payload: EventPayload;
```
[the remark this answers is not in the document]
svilenvelikov: fixed

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
```diff
@@ -50,14 +78,315 @@ export class YasguiComponentFacadeComponent {
...
+    url.searchParams.set('query', queryData.body);
+    // TODO: pass these from the component as well,
+    url.searchParams.set('infer', String(queryData.inference ?? true));
+    url.searchParams.set('sameAs', String(queryData.sameAs ?? true));
+    return url.toString();
+  }
+
+  private toEventData(event: OntotextYasguiEvent) {
+    return new EventData(event.detail.type, event.detail.payload);
+  }
```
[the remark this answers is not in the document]
svilenvelikov: removed

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
```diff
@@ -50,14 +78,315 @@ export class YasguiComponentFacadeComponent {
...
+  /**
+   * A computed property that returns the handlers for the different events emitted by the yasgui component via the
+   * generic output event channel.
+   * The handlers are defined in this component and are passed to the yasgui component via the config. This allows to
+   * handle events emitted from the yasgui component in a centralized way.
+   */
+  private readonly outputHandlers = computed((): OutputHandlers => {
+    const config = this.yasguiConfig();
+    return config?.outputHandlers ? {...config.outputHandlers} : {};
+  });
```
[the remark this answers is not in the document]
svilenvelikov: fixed

packages/workbench/src/app/components/yasgui-component-facade/models/query-response.ts
```diff
@@ -0,0 +1,10 @@
+export interface QueryResponse {
+  body: number | QueryResponseData
+}
+
+export interface QueryResponseData {
+  results: {
+    bindings: never[]
+  }
+  totalElements: {}
```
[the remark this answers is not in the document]
svilenvelikov: fixed

packages/workbench/src/app/components/yasgui-component-facade/models/query/save-query-event.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Event emitted by the ontotext-yasgui component when a query should be saved.
+ * The event payload contains the query data from which a saved query object should be created.
+ */
+export interface SaveQueryEvent extends CustomEvent {
+  detail: {
+    queryName: string;
+    query: string;
+    isPublic: boolean;
+    isNew?: boolean;
+    messages?: string[]
```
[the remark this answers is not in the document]
svilenvelikov: fixed

packages/api/src/services/domain/rdf4j/rdf4j-rest.service.ts
```diff
@@ -35,4 +35,45 @@ export class Rdf4jRestService extends HttpService {
...
+   * @param linkHeader - The value for the 'Link' header, which may contain additional information about the request or
+   * response, such as pagination details or related resources.
+   * @returns A Promise that resolves to the HTTP response containing the file data as a Blob, along with the
+   * appropriate headers for file download. The response will include the file data and the filename extracted from the
+   * 'Content-Disposition' header if available.
+   * @throws An error if the HTTP request fails or if the response does not contain the expected file data.
+   */
+  downloadResultsAsFile(repositoryId: string, data: object, acceptHeader: string, linkHeader: string) {
+    const properties = Object.entries(data)
+      .filter(([value]) => value !== undefined)
```
plamen-yordanov: `value` will point to the key of the entry here

packages/api/src/services/http/http.service.ts
```diff
@@ -364,7 +409,8 @@ export class HttpService {
         return fetch(request.url, {
           method: request.method,
           headers: request.headers as HeadersInit,
-          body
+          body,
+          signal: request.signal,
         });
       })
       .then((response) => this.interceptorService.postProcess(response));
```
yordanalexandrov, line 416: Wondering, will the aborted go through the post processors? Should they go or through them or bypass them.

Right now there are `UnauthenticatedInterceptor` and `UnauthorizedInterceptor`, which won't do much but still.

packages/workbench/src/app/components/yasgui-component-facade/mappers/autocomplete-result-to-yasgui-model.mapper.ts
```diff
@@ -0,0 +1,20 @@
...
+import {AutocompleteYasguiModel} from '../models/autocomplete-yasgui-model';
+
+/**
+ * Mapper for AutocompleteSearchResult objects to the expected internal YASGUI autocomplete model.
+ */
+export const mapAutocompleteSearchResultToYasguiModel: MapperFn<AutocompleteSearchResult, AutocompleteYasguiModel> = (data) => {
+  const result: AutocompleteYasguiModel = {
+    suggestions: []
+  };
+  result.suggestions = data?.suggestions.getItems().map((suggestion) => ({
```
yordanalexandrov, line 14: Here you set initial value which you then override with `[] | undefined`.

Maybe add a `?? []` at the end of the asignment

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-request-event.ts
```diff
@@ -0,0 +1,40 @@
...
+  constructor(eventData: OntotextYasguiEvent) {
+    this.query = eventData.detail.payload.query;
+    this.queryMode = eventData.detail.payload.queryMode;
+    this.queryType = eventData.detail.payload.queryType;
+    this.pageSize = eventData.detail.payload.pageSize;
+    this.request = eventData.detail.payload.request;
+  }
+
+  setPageSize(pageSize?: string) {
+    this.request._data['pageSize'] = pageSize;
```
yordanalexandrov, line 30: shoudn't the class members also be set?

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
```diff
@@ -0,0 +1,96 @@
...
+   * Constructs the model for {@link EventDataType.COUNT_QUERY_RESPONSE} event.
+   *
+   * @param eventData - event emitted by "ontotext-yasgui-web-component".
+   */
+  constructor(eventData: OntotextYasguiEvent) {
+    this.response = eventData.detail.payload.response ?? {} as QueryResponse;
+  }
+
+  hasResponse() {
+    return this.response?.body;
```
yordanalexandrov: the `?` is not needed based on the types

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  getResponseBody() {
+    if (this.hasResponse()) {
```
yordanalexandrov: can be simplifies to
```
return this.response.body ?? null
```

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
```diff
@@ -0,0 +1,96 @@
...
+      return this.response.body;
+    }
+    return null;
+  }
+
+  isResponseNumber() {
+    if (this.hasResponse()) {
```
yordanalexandrov, line 30: So if `body` is a number and that number is `0` there is no body.

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
```diff
@@ -0,0 +1,96 @@
...
+      return typeof this.response.body === 'number';
+    }
+    return false;
+  }
+
+  isResponseArray() {
+    if (this.hasResponse()) {
+      return Array.isArray(this.response.body);
```
yordanalexandrov: can be simplifies to `return Array.isArray(this.response.body)`

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+    return false;
+  }
+
+  getBindings() {
+    if (this.hasResponse() && typeof this.response.body !== 'number') {
+      return this.response.body.results.bindings;
```
yordanalexandrov, line 42: This would assume that if body is not a number it is an object.

But it could be an Array as we have the `isResponseArray` method.

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+    return null;
+  }
+
+  hasBindingResponse() {
+    if (this.hasResponse() && typeof this.response.body !== 'number') {
+      return this.response.body.results.bindings[0];
```
yordanalexandrov, line 49: Same as above, but maybe the check should be if `bindings.length > 0`

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
```diff
@@ -31,32 +80,774 @@ defineCustomElements();
...
+    let queryString: string;
+    const ontotextYasguiElement = this.getOntotextYasguiComponent();
+    ontotextYasguiElement.getQuery()
+      .then((query) => {
+        queryString = query;
+        return queryString;
+      })
+      .then((queryString: string) => this.sparqlService.addKnownPrefixes(queryString))
+      .then((queryString: string) => {
+        return ontotextYasguiElement.setQuery(queryString);
```
plamen-yordanov, line 598: the old implementation re-assigns `queryString` here. Currently `emitQueryChanged` emits a query without prefixes. We should assign here or add a parameter to the last promise, since the query is returned

plamen-yordanov (changes requested):

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  private resolveTotalElements() {
+    let totalElements = undefined;
+    if (this.isResponseNumber()) {
+      totalElements = this.getResponseBody();
+    }
+
+    if (this.isResponseArray()) {
+      const body = this.getResponseBody() as TextArrayResponse;
```
yordanalexandrov, line 72: This is some strange magic. `TextArrayResponse` does not define an Array

plamen-yordanov:
the title of the MR points to a different jira issue

packages/workbench/src/app/components/yasgui-component-facade/service/yasgui-persistence-migration.service.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  /**
+   * Checks if migration is needed by checking if the old yasgui persistence is present in local storage.
+   * @returns true if migration is needed, false otherwise
+   */
+  isMigrationNeeded(): boolean {
+    const yasguiData = this.getPersistedData<SparqlViewData>(NEW_YASGUI_PERSISTENCE_KEY);
+    let isMigrated = false;
+    if (yasguiData) {
+      isMigrated = !!(yasguiData.val.migration &&
```
yordanalexandrov, line 84: Perhaps have some value for the `migration` property, in case we do another migration in the future?

I see there is a date but  it is the date of actual migration

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
```diff
@@ -31,32 +80,774 @@ defineCustomElements();
...
+   * @param notificationMessageEvent - the "ontotext-yasgui-web-component" event.
+   */
+  private readonly notificationMessageHandler = (notificationMessageEvent: NotificationMessageEvent) => {
+    const { messageType, message } = notificationMessageEvent;
+    if (!messageType || !message) {
+      return;
+    }
+    switch (messageType) {
+    case 'success':
+      this.toastrService.success(message);
```
yordanalexandrov, line 441: maybe move to the `NotificationService`. There is an exported `notify` function.

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
```diff
@@ -31,32 +80,774 @@ defineCustomElements();
...
+      el.removeEventListener('paste', this.delegatedPasteHandler);
+      el.removeEventListener('keyup', this.delegatedKeyupHandler);
+    });
+  }
+
+  private async init(yasguiConfig: OntotextYasguiConfig | undefined) {
+    if (!yasguiConfig) {
+      return;
+    }
+    console.info('%cinit config', 'background: orange', yasguiConfig);
```
yordanalexandrov, line 470: Remove at some point

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
```diff
@@ -31,32 +80,774 @@ defineCustomElements();
...
+    );
+    this.config.set(actualConfig);
+
+    this.addDirtyCheckHandlers();
+
+    this.setInitialQueryState().then(() => {
+      this.afterInit()?.();
+    });
+
+    console.info('%cconfig updated', 'background: tan', {runtimeConfig, actualConfig, config: this.config()});
```
yordanalexandrov, line 537: remove at some point

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
```diff
@@ -31,32 +80,774 @@ defineCustomElements();
...
+      .then((queryString: string) => this.sparqlService.addKnownPrefixes(queryString))
+      .then((queryString: string) => {
+        return ontotextYasguiElement.setQuery(queryString);
+      })
+      .then(() => {
+        this.emitQueryChanged(JSON.stringify(queryString));
+      })
+      .catch((error) => {
+        const errorMessage = error instanceof Error ? error.message : String(error);
+        this.toastrService.error(
```
yordanalexandrov, line 605: NotificationService

packages/workbench/src/app/pages/sparql-editor/sparql-editor-page.component.scss
```diff
@@ -30,31 +30,22 @@
   margin-top: calc(47px + 0.5em);
 }
 
-.sparql-editor-page #query-editor .yasgui-host-element {
+:host ::ng-deep #query-editor .yasgui-host-element {
```
yordanalexandrov: Perhaps make this a bit more specific, as other pages might contain the same id class and the style might leak.

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
```diff
@@ -31,32 +80,774 @@ defineCustomElements();
...
+      isVirtualRepository,
+      infer: isVirtualRepository || yasguiConfig.infer,
+      immutableInfer: isVirtualRepository,
+      sameAs: isVirtualRepository || yasguiConfig.sameAs,
+      immutableSameAs: isVirtualRepository,
+      yasqeAutocomplete: {
+        LocalNamesAutocompleter: (term: string) => {
+          this.autocompleteAbortController?.abort();
+          this.autocompleteAbortController = new AbortController();
+          return this.autocompleteLocalNames(term, this.autocompleteAbortController?.signal);
```
boyan-tonchev: ?. is unnecessary here because` this.autocompleteAbortController` is guaranteed to be initialized at this point.

packages/api/src/services/http/http.service.ts
svilenvelikov: No. Aborted requests skip the postprocessing entirely. When a request is aborted it is intentional. For the aborted by the abort controller requests, the fetcher rejects and the flow continues with the catch, if any. And this the expected behavior.

packages/workbench/src/app/components/yasgui-component-facade/mappers/autocomplete-result-to-yasgui-model.mapper.ts
svilenvelikov: Good point. 
Fixed

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-request-event.ts
svilenvelikov: TBH I'm not sure. The legacy event is designed the same way, but I think it is just underthought. As far as I can see, the only class member that is used is `request`. I was planning to redesign those events with the task for the `output` event channel removal.

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
svilenvelikov: removed

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
```diff
@@ -31,32 +80,774 @@ defineCustomElements();
...
+      const yasguiConfig = this.yasguiConfig();
+      if (!yasguiConfig) {
+        return;
+      }
+      this.init(yasguiConfig);
+    });
+  }
+
+  ngOnDestroy(): void {
+    this.removeDirtyCheckHandlers();
```
boyan-tonchev, line 157: We have to abort the existing autocompleteAbortController, if one exists.

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
svilenvelikov: That's a bug. Will fix the #hasResponse() method to be explicit in the check.

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
svilenvelikov: You are right. There should be a guard for the probable array response.

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
svilenvelikov: It does. Check again
`'http://www.ontotext.com/'?: { value: string | number }[]`
And yes. I agree that it is strange, but this is the actual response and I had to type it somehow. I could've taken more generic approach here, but I didn't oversee a reason or application for this, because this is the only place where this type will be used.

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
svilenvelikov: You are right. I should've deleted this by mistake maybe.

packages/workbench/src/app/components/yasgui-component-facade/service/yasgui-persistence-migration.service.ts
svilenvelikov: I didn't get what you suggest exactly.
I initially thought about removing this already because it was initially introduced with GDB 10.4 I believe and now we are heading to GDB 12 already. 
But then I decided to leave it because with the page migration, we would most likely do some other changes.

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
svilenvelikov: I know I should, but I'll do it with a separate MR.
The problem is that we need a strategy/convention for the codes and I don't want to deal with those here/now.
I'll create a task for it.
https://graphwise.atlassian.net/browse/GDB-14637

packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
svilenvelikov: You are right. Although, the abort controller is used to terminate ongoing namespaces autocomplete while the user types in the query editor and it's very unlikely that this flow would be hit easily.
Anyway. I'll add the abort here as well.

plamen-yordanov (dismissed):

packages/workbench/src/app/components/yasgui-component-facade/models/event/count-query-response-event.ts
svilenvelikov: but maybe you've meant that the name of the type is bad

boyan-tonchev (approved):

PR #2900, opened 2026-04-23
GDB-14488 display markdown for markdown typed literals in the node info sidebar

svilenvelikov:
## What
Support rendering markdown for markdown typed literals in the node info sidebar.

## Why
This is extending the support of different literal types.

## How
* Included the markdown-content directive in the graphs visualization view and using it for rendering literals with datatypes
```
http://www.ontotext.com/proton/protonsys#Markdown
http://ns.ontowiki.net/SysOnt/Markdown
https://www.w3.org/ns/iana/media-types/text/markdown#Resource
http://www.w3.org/ns/iana/media-types/text/markdown#Resource
```

* Optimized the markdown-content directive to prevent extra reinitialization when the content didn't change.
Also using one-time binding for the content and ng-if in the markdown-content directive declaration because the data used is static and doesn't change.
* Eliminate flickering of markdown-content directive in graph node info panel. The node info side panel was causing the markdown-content directive to rerender 4 times when a graph node was selected resulting in visible flickering.

## Testing
Insert data with query like this
```
PREFIX ex: <http://example.org/>
PREFIX sysont: <http://www.ontotext.com/proton/protonsys#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>

INSERT DATA {
    ex:MyNode a ex:Thing ;
        rdfs:label "My Node" ;
        ex:description "# My Node\n\nThis is a **Markdown** literal.\n\n- item 1\n- item 2"^^sysont:Markdown .
}
```
and browse the visual graph for the `MyNode`

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/core/directives/markdown-content/markdown-content.js
+++ b/packages/legacy-workbench/src/js/angular/core/directives/markdown-content/markdown-content.js
@@ -1,7 +1,7 @@
 import 'angular/core/services/markdown/markdown.service';
 
 const modules = [
-    'graphdb.framework.core.services.markdown-service'
+    'graphdb.framework.core.services.markdown-service',
 ];
 
 /**
@@ -34,10 +34,10 @@ function markdownContentDirective($compile, MarkdownService) {
         restrict: 'E',
         scope: {
             content: '@',
-            options: '='
+            options: '=',
+            trusted: '@',
         },
-        link: function ($scope, element) {
-
+        link: function($scope, element) {
             // =========================
             // Public variables
             // =========================
@@ -45,7 +45,7 @@ function markdownContentDirective($compile, MarkdownService) {
             /**
              * The rendered HTML content generated from the provided Markdown.
              *
-             * @type {string} markdownContent
+             * @type {string|undefined} markdownContent
              */
             $scope.markdownContent = undefined;
 
@@ -54,8 +54,16 @@ function markdownContentDirective($compile, MarkdownService) {
             // =========================
 
             const init = () => {
-                $scope.markdownContent = MarkdownService.renderMarkdown($scope.content, $scope.options);
-                compileMarkdownAnswerContents();
+                if (!$scope.content) {
+                    $scope.markdownContent = undefined;
+                    return;
+                }
+                if ($scope.trusted === 'true') {
+                    $scope.markdownContent = MarkdownService.renderMarkdown($scope.content, $scope.options);
+                    compileMarkdownAnswerContents();
+                } else {
+                    $scope.markdownContent = MarkdownService.renderMarkdown($scope.content, $scope.options, false);
+                }
             };
 
             /**
@@ -71,7 +79,13 @@ function markdownContentDirective($compile, MarkdownService) {
                 });
             };
 
+            $scope.$watch('content', (newVal, oldVal) => {
+                if (newVal !== oldVal) {
+                    init();
+                }
+            });
+
             init();
-        }
+        },
     };
 }
--- a/packages/legacy-workbench/src/js/angular/core/services/markdown/markdown.service.js
+++ b/packages/legacy-workbench/src/js/angular/core/services/markdown/markdown.service.js
@@ -15,12 +15,12 @@ const logger = LoggerProvider.logger;
  * @param {$sce} $sce - AngularJS service for Strict Contextual Escaping.
  */
 angular
-    .module('graphdb.framework.core.services.markdown-service', [])
+    .module('graphdb.framework.core.services.markdown-service', ['ngSanitize'])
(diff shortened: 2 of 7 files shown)


packages/legacy-workbench/src/js/angular/core/directives/markdown-content/markdown-content.js
```diff
@@ -54,6 +53,9 @@ function markdownContentDirective($compile, MarkdownService) {
             // =========================
 
             const init = () => {
+                if (!$scope.content) {
+                    return;
+                }
                 $scope.markdownContent = MarkdownService.renderMarkdown($scope.content, $scope.options);
                 compileMarkdownAnswerContents();
```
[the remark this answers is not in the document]
svilenvelikov: Cleared the markdownContent

packages/legacy-workbench/src/pages/graphs-visualizations.html
```diff
@@ -541,7 +541,8 @@ <h3 class="hovered-parent">
                         <div class="item clearfix break-word-alt small">{{item.key}}</div>
                         <div class="data-value">
                             <a ng-if="item.value[0].t === 'i'" href="#" ng-click="openIRI(item.value[0].v, $event)" ng-bind="item.value[0].v"></a>
-                            <span ng-if="item.value[0].t !== 'i'" ng-bind="item.value[0].v"></span>
+                            <span ng-if="item.value[0].t !== 'i' && !isMarkdownLiteral(item.value[0].t)" ng-bind="item.value[0].v"></span>
+                            <markdown-content ng-if="::isMarkdownLiteral(item.value[0].t)" content="{{::item.value[0].v}}"></markdown-content>
```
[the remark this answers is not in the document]
svilenvelikov: made the markdown-content directive safe to render content with potential xss infected strings

packages/legacy-workbench/src/js/angular/graphexplore/controllers/graphs-visualizations.controller.js
```diff
@@ -2581,25 +2590,33 @@ function GraphsVisualizationsCtrl(
...
                 return o.key === 'image';
             });
             $scope.propertiesNotFiltered = $scope.propertiesObj.items;
+
+            $scope.showInfoPanel = true;
+            $scope.showNodeInfo = true;
+            if (panelWasAlreadyOpen) {
+                // ui-scroll doesn't re-init automatically when panel stays open
+                $timeout(() => $scope.adapterContainer.adapter.reload(), 500);
+            }
```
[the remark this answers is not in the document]
svilenvelikov: fixed the issue

packages/legacy-workbench/src/js/angular/core/services/markdown/markdown.service.js
```diff
@@ -44,19 +44,40 @@ function MarkdownService($sce) {
...
         }
     };
 
+    /**
+     * Renders Markdown text into HTML without trusting the output, ensuring that any potentially unsafe content is
+     * properly escaped.
+     * @param text - The Markdown text to render.
+     * @returns {string} The rendered HTML, or the original text in case of an error.
+     */
+    const renderMarkdownUntrusted = (text) => {
```
yordanalexandrov: Why are you not using `const html = getMarkdown(config).render(text);` like the other method?

yordanalexandrov: I am a bit confused. There is a function `renderMarkdown` with argument `trusted` and a function `renderMarkdownUntrusted` which does something else.

yordanalexandrov: Isn't it better to name it `...sanitized...` or something?

svilenvelikov: You are right it's confusing. I'll remove the new one and update the existing to also sanitize the content which is what I wanted.

svilenvelikov: Actually, the difference was needed because we plugin some extensions in the markdown in some cases and those plugins like markdown-code-copy are producing html that must not be sanitized while in the new method I wanted explicitly sanitization strategy to be applied to prevent xss.
Anyway. I think I can still merge both methods

svilenvelikov: Both methods were merged so this is now obsolete

yordanalexandrov (approved):

PR #2904, opened 2026-04-24
GDB-14502 fix class relations not loading

plamen-yordanov:
## What
Fix class relations not loading

## Why
There might be a lot of data in the repository in which case it takes graphdb a lot of time to load/reload the diagram. Currently we retry after 2 seconds, but that is nowhere near enough as this might take 20/30 seconds and even more, depending on the data

## How
Replaced the timeout with an interval. Now the workbench will continuously poll for the status until a `READY` status is received. Then the polling will be stopped

## Testing
n/a

## Screenshots
while polling:
Screenshot from 2026-04-24 14-56-01

polling stops after GDB is `READY`
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/graphexplore/controllers/dependencies-chord.controller.js
+++ b/packages/legacy-workbench/src/js/angular/graphexplore/controllers/dependencies-chord.controller.js
@@ -1,10 +1,10 @@
 import {NumberUtils} from "../../utils/number-utils";
-import {service, RepositoryContextService} from '@ontotext/workbench-api'
+import {service, RepositoryContextService} from '@ontotext/workbench-api';
 
 import {select} from 'd3';
 
 const d3 = {
-  select
+  select,
 };
 
 const STATUS = {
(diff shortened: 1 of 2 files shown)


packages/legacy-workbench/src/js/angular/graphexplore/controllers/dependencies-chord.controller.js
```diff
@@ -156,34 +158,38 @@ function DependenciesChordCtrl($scope, $rootScope, $repositories, toastr, $timeo
...
-                        $timeout.cancel(timer);
+                        $interval.cancel(timer);
                         timer = null;
                     }
                     getRelationshipsClasses();
                 }
                 if ($scope.status.indexOf('ERROR;') === 0) {
                     $scope.status = STATUS.ERROR;
+                    if (timer !== null) {
+                        $interval.cancel(timer);
```
boyan-tonchev, line 186: This can be extracted into a method and reused

boyan-tonchev (approved):

yordanalexandrov (approved):

PR #2915, opened 2026-04-30
GDB-14271: Amend YASR fullscreen configuration

yordanalexandrov:
## What
Updated the YASR fullscreen configuration to comply with YASGUI changes

## Why
It ensures that the fullscreen mode is enabled when the component is embedded.

## How
- Removed the YasrFullscreenConfig object and replaced it with a boolean value.
- Updated the yasguiConfig to directly set fullscreen behavior based on the embedded state.
- Adjusted the HTML to conditionally display error messages based on the embedded state.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/legacy-workbench/src/pages/sparql-editor.html
```diff
@@ -1,12 +1,12 @@
 <link href="css/sparql-editor.css?v=[AIV]{version}[/AIV]" rel="stylesheet"/>
 
 <div class="sparql-editor-view">
-    <h1 id="title-container">
+    <h1 id="title-container" ng-if="!embedded">
```
svilenvelikov, line 4: Are these applied to the new sparql view page?

yordanalexandrov: Amended `PageLayoutComponent` to consider `embedded` query parameter.

packages/workbench/src/app/pages/sparql-editor/sparql-editor-page.component.ts
```diff
@@ -737,6 +737,8 @@ export class SparqlEditorPageComponent implements OnInit, OnDestroy {
         this.logger.error('Error fetching namespaces for the active repository: ', error);
         // Even if fetching namespaces fails, we should still initialize the editor with an empty prefix list
         this.prefixes = new NamespaceMap({});
+      })
+      .finally(() => {
```
boyan-tonchev: Is this correct? On success we call `this.initViewFromUrlParams(true),` and then in finally we call it again with `this.initViewFromUrlParams(clearYasguiState). `The same applies to `this.setInferAndSameAs(authenticatedUser).`

yordanalexandrov: I will double check

yordanalexandrov: I reverted this change.
The problem was that the state and results were cleared on each refresh and I could not test the embedded flag.

This probably needs more thinking and is actually irrelevant to the PR.

boyan-tonchev (dismissed):

svilenvelikov (dismissed):

boyan-tonchev (approved):

PR #2928, opened 2026-05-07
GDB-14530 improve the security of remote location setup

svilenvelikov:
## What
Add an option to encrypt the basic auth password for a remote location.

## Why
Fixing a security issue.

## How
When basic auth is chosen as auth type when creating/editing a remote location, an option to encrypt the password is added for the user to choose. 
The new option is stored in a flag inside the remote location model and is sent to the server.
Also, a warning message is added to notify the user that by default the password is stored unencrypted. As well as a warning appearing when the user selects the option to encrypt the password saying that this will harm the compatibility with older GDB instances.

## Testing
Test added.

## Screenshots
image

image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/legacy-workbench/src/js/angular/repositories/controllers/manage-remote-location-dialog.controller.js
```diff
@@ -25,15 +25,18 @@ ManageRemoteLocationDialogController.$inject = ['$scope', '$uibModalInstance', '
...
      */
     $scope.remoteLocation = data.remoteLocation ? new RemoteLocationModel({...data.remoteLocation}) : new RemoteLocationModel();
+
+    $scope.RemoteLocationType = RemoteLocationType;
+    $scope.RemoteLocationAuthType = RemoteLocationAuthType;
+    $scope.isEdit = !!data.remoteLocation;
+    // Flag showing if the encryptPassword checkbox should be disabled.
+    // TODO: TBD: The checkbox can be enabled only when creating a new location or if editing existing location for which the
+    // encryptPassword=false. The reason is that, once password is encrypted, it cannot be decrypted.
+    $scope.passwordEncryptionFieldDisabled = $scope.isEdit && $scope.remoteLocation.encryptPassword;
```
yordanalexandrov: When editing the location with encryptPassword set, why can't we replace the encrypted password with another? Why do we need to decrypt it?

svilenvelikov: Ignore the TODO comment. It's obsolete and the idea was not to decript the passowrd but to either replace the password what is the current behavior or to disable the checkbox to prevent disabling it which we rejected as an approach.

yordanalexandrov (dismissed):

plamen-yordanov (dismissed):

yordanalexandrov (dismissed):

PR #2936, opened 2026-05-10
GDB-14531 prevent disabling of security

plamen-yordanov:
## What
Prevent disabling of security through the workbench whenever GDB is configured with `graphdb.auth.security.enabled=true`.

## Why
Users should not be able to disable security in that case.

## How
- Added `allowSecurityToggle` property to the `SecurityConfig` class.
- Added `isSecurityToggleAllowed()` method to check if toggling is permitted.
- Updated the `users.html` to disable the security toggle if not allowed.
- Added conditional tooltips for both cases

## Testing
n/a

## Screenshots

Screenshot from 2026-05-10 18-32-41
Screenshot from 2026-05-10 18-32-54
Screenshot from 2026-05-11 14-59-26

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/legacy-workbench/src/js/angular/security/controllers.js
```diff
@@ -47,6 +48,16 @@ securityModule.controller('UsersCtrl', ['$scope', '$uibModal', 'toastr', '$windo
...
             return authenticationService.isSecurityEnabled();
         };
+        $scope.isSecurityToggleAllowed = function() {
+            return securityContextService.getSecurityConfig()?.isSecurityToggleAllowed() ?? true;
+        };
+        $scope.getSecurityToggleTooltip = function() {
+            if (!$scope.isSecurityToggleAllowed()) {
+                return $translate.instant('security.toggle.not.allowed');
+            }
+            const action = $scope.securityEnabled() ? $translate.instant('disable') : $translate.instant('enable');
```
svilenvelikov: I don't like these granular labels. They are error-prone, hard for refactong and reason about when they are used more than once in different contexts. I'd suggest just creating a new label for this case where you can pass the disabled status as a parameter.

packages/legacy-workbench/src/js/angular/security/controllers.js
```diff
@@ -47,6 +48,16 @@ securityModule.controller('UsersCtrl', ['$scope', '$uibModal', 'toastr', '$windo
         $scope.securityEnabled = function() {
             return authenticationService.isSecurityEnabled();
         };
+        $scope.isSecurityToggleAllowed = function() {
+            return securityContextService.getSecurityConfig()?.isSecurityToggleAllowed() ?? true;
+        };
+        $scope.getSecurityToggleTooltip = function() {
```
svilenvelikov: This is too complex and heavy for a simple tooltip resolution. I'd suggest replacing the method binding with a scoped variable that is resolved on a security config change.

packages/legacy-workbench/src/js/angular/security/controllers.js
```diff
@@ -47,6 +48,16 @@ securityModule.controller('UsersCtrl', ['$scope', '$uibModal', 'toastr', '$windo
         $scope.securityEnabled = function() {
             return authenticationService.isSecurityEnabled();
         };
+        $scope.isSecurityToggleAllowed = function() {
+            return securityContextService.getSecurityConfig()?.isSecurityToggleAllowed() ?? true;
```
svilenvelikov: Same here. Define a scope variable and toggle it on security config changes

packages/legacy-workbench/src/i18n/locale-en.json
```diff
@@ -1702,6 +1702,7 @@
     "security.any.data.repo": "Any data repository",
     "security.data.repos": "Data repositories are all repositories that aren't the SYSTEM repository.",
     "security.is": "Security is",
+    "security.toggle.not.allowed": "Toggling security is not allowed due to GDB configuration",
```
yordanalexandrov: I think it will be more infomative if we change this to "Security is set through GraphDB configuration"

svilenvelikov: This message will be extended with a documentation link and must be displayed with a popover. Consult with Donka about the wording!

packages/legacy-workbench/src/js/angular/security/controllers.js
plamen-yordanov: created a new label and added action as param

packages/legacy-workbench/src/js/angular/security/controllers.js
plamen-yordanov: created a scoped variable. Security config change is not needed as it may not change runtime without page reload. Added language change subscription however, since we need to update the labels whenever language is changed

packages/legacy-workbench/src/js/angular/security/controllers.js
plamen-yordanov: changed to a scope variable. It doesn't need anything else as config doesn't change at runtime

packages/legacy-workbench/src/i18n/locale-en.json
plamen-yordanov: Added popover template with link and updated labels as per Donka's suggestions

packages/legacy-workbench/src/js/angular/security/templates/users.html
```diff
@@ -7,9 +7,21 @@ <h1>
...
 				<span class="security-switch-label">{{'security.is' | translate}} <span class="tag {{securityEnabled() ? 'tag-primary' : 'tag-default'}}">{{securityEnabled() ? 'common.on.btn' : 'common.off.btn' | translate}}</span></span>
-				<span gdb-tooltip="{{'click.to' | translate}} {{securityEnabled() ? 'disable' : 'enable' | translate}} {{'security.label' | translate}}" tooltip-placement="top" ng-click="toggleSecurity()" class="switch">
-					<input type="checkbox" class="switch" ng-checked="securityEnabled()"/>
-					<label for="toggle-security"></label>
+				<span class="switch">
+					<input type="checkbox" id="securityToggle" class="switch" ng-disabled="!isSecurityToggleAllowed" ng-checked="securityEnabled()"/>
+					<label ng-if="!isSecurityToggleAllowed"
+                           for="securityToggle"
+						   uib-popover-template="'js/angular/security/templates/security-toggle-not-allowed-popover.html'"
+						   popover-popup-close-delay="800"
```
svilenvelikov: These popovers are a bit unfriendly implemented this way, because you can't actually easily copy anything from them and the reason is that they just disappear too fast when you get out of the target element. This is the same approach you applied for the encryption property in the repository view. It'd be better if we keep the popover visible while the pointer is over the target or inside the popover.

packages/legacy-workbench/src/js/angular/security/controllers.js
```diff
@@ -31,22 +33,46 @@ const modules = [
...
         const usersService = service(UsersService);
         const repositoryAuthorityService = service(RepositoryAuthorityService);
+        const securityContextService = service(SecurityContextService);
+        const languageContextService = service(LanguageContextService);
 
         // UI view-model
         $scope.usersData = [];
 
         $scope.loader = true;
+        $scope.isSecurityToggleAllowed = securityContextService.getSecurityConfig()?.isSecurityToggleAllowed() ?? true;
```
svilenvelikov: I'd put the default in the security config model

packages/legacy-workbench/src/js/angular/security/templates/users.html
plamen-yordanov: I know, but I couldn't find anything in the bootstrap config, which achieves that. I have to come up with something custom. That is why I took the same approach as in the agent configuration popover

plamen-yordanov: I found something relatively light. Added the popover inside the parent element, instead of appending it to the body. Remove the triggers and manually set the hovered state on mouseleave and mouseenter on the parent. It has to be an object with a hovered property, otherwise there is some issue with the scope and it doesn't work

packages/legacy-workbench/src/js/angular/security/controllers.js
plamen-yordanov: I removed the default altogether. There shouldn't be a case, where it is not defined or null

packages/legacy-workbench/src/js/angular/security/templates/users.html
svilenvelikov: looks good now

svilenvelikov (dismissed):

svilenvelikov (approved):

PR #2937, opened 2026-05-11
GDB-13827: Implement generic dialog service in workbench

yordanalexandrov:
## What
Added a new DialogProviderService for managing dialog components and their configurations. Introduced an DialogOpenOptions interface to standardize dialog options.

## Why
This change enhances the dialog management capabilities within the application, allowing for more flexible and reusable dialog components.

## How
- Created `DialogOpenOptions` interface to define dialog configuration options.
- Implemented `DialogProviderService` to handle dialog opening and configuration.
- Updated `app.config.ts` to provide the PrimeNG `DialogService`.
- Added unit tests for `DialogProviderService` to ensure functionality and reliability.

## Testing
added

## Screenshots
n/a

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/workbench/src/app/app.config.ts
+++ b/packages/workbench/src/app/app.config.ts
@@ -7,7 +7,8 @@ import {routes} from './app.routes';
 import {bootstrapProviders} from './bootstrap/bootstrap';
 import {APP_BASE_HREF} from '@angular/common';
 import {getSingleSpaExtraProviders} from 'single-spa-angular';
-import { providePrimeNG } from 'primeng/config';
+import {providePrimeNG} from 'primeng/config';
+import {DialogService} from 'primeng/dynamicdialog';
 
 export const appConfig: ApplicationConfig = {
   providers: [
@@ -24,6 +25,7 @@ export const appConfig: ApplicationConfig = {
           prefix: 'gw',
         }
       }
-    })
+    }),
+    DialogService
   ]
 };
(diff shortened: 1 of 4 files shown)


packages/workbench/src/app/services/onto-dialog.service.ts
```diff
@@ -0,0 +1,44 @@
...
+import {map, Observable} from 'rxjs';
+import {OntoDialogConfig} from '../models/onto-dialog-config';
+
+/**
+ * Service for opening dialogs in the application. It provides methods to open dialogs with a specified configuration.
+ */
+@Injectable({
+  providedIn: 'root'
+})
+export class OntoDialogService {
```
svilenvelikov: I don't think we need the `Onto` prefix.

yordanalexandrov: We need some kind of prefix or suffix or another name, since `PrimeNG` has decided to name their dialog service - `DialogService`

svilenvelikov: Yeah. I can see that, but it'd be great if we don't stick with such prefixes again. I'd suggest a name like `DialogProviderService` or similar

yordanalexandrov: Changed name.

svilenvelikov (approved):

svilenvelikov:
You can update the PR description because it's outdated now

PR #2940, opened 2026-05-12
GDB-14600 make popovers visible while hovered

plamen-yordanov:
## What
Make popovers for storage encryption and similarity extraction methods visible while hovered.

## Why
They closed after a set timeout of 800ms before. Now they remain visible on hover, which makes it easier to click their links

## How
- Added `ng-mouseenter` and `ng-mouseleave` events to control the visibility of popovers for the storage encryption checkbox and similarity extraction method.
- Updated the `uib-popover` attributes to use `popover-is-open` for dynamic visibility based on hover state.

## Testing
n/a

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/repositories/controllers.js
+++ b/packages/legacy-workbench/src/js/angular/repositories/controllers.js
@@ -510,6 +510,7 @@ function AddRepositoryCtrl($rootScope, $scope, toastr, $repositories, $location,
     $scope.enable = true;
     $scope.entityIndexSizeMin = ENTITY_INDEX_SIZE_MIN;
     $scope.documentationUrlForStorageEncryption = DocumentationUrlResolver.getDocumentationUrl(productInfo.productShortVersion, 'creating-a-repository.html#using-workbench');
+    $scope.encryptInput = {hovered: false};
 
     $scope.loader = true;
     $scope.pageTitle = $translate.instant('view.create.repo.title');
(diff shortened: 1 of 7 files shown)


svilenvelikov (approved):

packages/legacy-workbench/src/js/angular/repositories/controllers.js
```diff
@@ -510,6 +510,7 @@ function AddRepositoryCtrl($rootScope, $scope, toastr, $repositories, $location,
     $scope.enable = true;
     $scope.entityIndexSizeMin = ENTITY_INDEX_SIZE_MIN;
     $scope.documentationUrlForStorageEncryption = DocumentationUrlResolver.getDocumentationUrl(productInfo.productShortVersion, 'creating-a-repository.html#using-workbench');
+    $scope.encryptInput = {hovered: false};
```
yordanalexandrov, line 513: Why do we need an object for this and for `similarityInput`? Is it not enough for them to be booleans?

svilenvelikov: I guess it's angular.js BS. It can't properly inherit primitive values in child scopes. At least this is in our angular version

plamen-yordanov: There is some issue with the scopes (not sure what) and it doesn't work

PR #2953, opened 2026-05-15
GDB-12154 add unsaved changes mechanism

plamen-yordanov:
## What
Add unsaved changes detection mechanism to prevent navigation when there are unsaved changes.

## Why
This feature addresses the issue of users accidentally losing unsaved changes when navigating away from a component. It enhances user experience by providing a confirmation dialog that alerts users about unsaved changes.

## How
- Introduced a new service `UnsavedChangesService` to manage unsaved changes across components.
- Created an abstract class `UnsavedChangesComponent` for components to implement unsaved changes logic.
- Updated `app.component.html` to include a confirmation dialog component.
- Modified the `application-lifecycle-context.service.ts` to handle navigation events related to unsaved changes.
- Added French and English translations for confirmation messages.

## Testing
jest

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/workbench/src/app/services/unsaved-changes/unsaved-changes.service.ts
+++ b/packages/workbench/src/app/services/unsaved-changes/unsaved-changes.service.ts
@@ -1,0 +1,81 @@
+import {inject, Injectable} from '@angular/core';
+import {
+  service,
+  EventService,
+  EventName,
+  NavigationStartPayload,
+  Subscription,
+  SubscriptionList
+} from '@ontotext/workbench-api';
+import {ConfirmationService} from 'primeng/api';
+import {TranslocoService} from '@jsverse/transloco';
+
+/**
+ * Coordinates unsaved-changes detection across the application.
+ *
+ * Callers register a callback via {@link register}, which returns a {@link Subscription}
+ * that removes the callback when invoked. When a `NAVIGATION_START` event fires and at least
+ * one registered callback returns `true`, the service intercepts navigation and shows a
+ * confirmation dialog. The user can then allow or cancel the navigation.
+ */
+@Injectable({
+  providedIn: 'root'
+})
+export class UnsavedChangesService {
+  private readonly callbacks = new SubscriptionList();
+  private readonly eventService = service(EventService);
+  private readonly translocoService = inject(TranslocoService);
+  private readonly confirmationService = inject(ConfirmationService);
+
+  constructor() {
+    this.subscribeToNavigationStart();
+  }
+
+  /**
+   * Registers `callback` to be consulted before each navigation. Returns a {@link Subscription}
+   * that, when called, removes the callback so it no longer blocks navigation.
+   */
+  public register(callback: () => boolean): Subscription {
+    this.callbacks.add(callback);
+    return () => this.callbacks.remove(callback);
+  }
+
+  private subscribeToNavigationStart() {
+    this.eventService.subscribe(EventName.NAVIGATION_START, (eventPayload: NavigationStartPayload) => this.onNavigationStart(eventPayload));
+  }
+
+  private async onNavigationStart(eventPayload: NavigationStartPayload) {
+    if (!this.callbacks.size || !this.hasAnyUnsavedChanges()) {
+      return;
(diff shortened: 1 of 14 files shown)


packages/workbench/src/app/services/unsaved-changes/unsaved-changes.service.ts
```diff
@@ -0,0 +1,74 @@
...
+   * Registers `callback` to be consulted before each navigation. Returns a {@link Subscription}
+   * that, when called, removes the callback so it no longer blocks navigation.
+   */
+  public register(callback: () => boolean): Subscription {
+    this.callbacks.add(callback);
+    return () => this.callbacks.delete(callback);
+  }
+
+  private subscribeToNavigationStart() {
+    this.eventService.subscribe(EventName.NAVIGATION_START, (eventPayload: NavigationStartPayload) => this.onNavigationStart(eventPayload));
```
yordanalexandrov, line 44: We are now supporting navigation start at two places, as Event and in the `ApplicationLifecycleContextService`.

Both are at different lifecycle points. I wonder if we need both.

WDYT as the one debugging it lately.

yordanalexandrov (dismissed):

packages/workbench/src/app/services/unsaved-changes/unsaved-changes.service.ts
plamen-yordanov: Am I missing something? I don't see  navigationStart in `ApplicationLifecycleContextService`

yordanalexandrov: There isn't.

`NavigationStart` event is emitted on `single-spa:before-routing-event`.

The `NavigationBeforeMountRouting` is updated on `single-spa:before-mount-routing-event`.

Both are routing events and are subscribed in the root config.

svilenvelikov (dismissed):

packages/workbench/src/app/services/unsaved-changes/unsaved-changes.service.ts
plamen-yordanov: ok, so both are needed due to their chronological order. We cancel routing in `single-spa:routing-event` and we show the loader in `single-spa:before-mount-routing-event`, which comes immediately after that. NavigationStart is not emitted in the second one. I made it emit ApplicationStateChange not to have conflicts

packages/workbench/src/app/services/unsaved-changes/unsaved-changes.service.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Callers register a callback via {@link register}, which returns a {@link Subscription}
+ * that removes the callback when invoked. When a `NAVIGATION_START` event fires and at least
+ * one registered callback returns `true`, the service intercepts navigation and shows a
+ * confirmation dialog. The user can then allow or cancel the navigation.
+ */
+@Injectable({
+  providedIn: 'root'
+})
+export class UnsavedChangesService {
+  private readonly callbacks = new Set<() => boolean>();
```
yordanalexandrov: Why is this a `Set`? all functions have differernt references I believe.

plamen-yordanov: replaced with `SubscriptionList`. Added a remove method to it, since it didn't exist

packages/workbench/src/assets/i18n/en.json
```diff
@@ -9,6 +9,7 @@
     "dialog": {
       "confirmation": {
         "title": "Confirm",
+        "message": "You have unsaved changes. Are you sure you want to continue?",
```
svilenvelikov, line 12: I don't think this should be here. Can you move it as a separate component, e.g. `components.dialog.confirmation.unsaved_changes.message` or even `components.dialog.unsaved_changes.confirmation.message`

svilenvelikov (approved):

PR #2954, opened 2026-05-15
GDB-14613 add warning for shown link limit

plamen-yordanov:
## What
Add a warning message for users when the displayed links reach the specified limit.

## Why
This message tells the users that there might be more links and suggests increasing the limit to view them

## How
- Introduced a new scope variable `showingLimit` to track when the limit is reached.
- Updated the HTML to conditionally display the warning message based on the new variable.

## Testing
cypress

## Screenshots
image
Screenshot from 2026-05-15 11-42-58

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/graphexplore/controllers/graphs-visualizations.controller.js
+++ b/packages/legacy-workbench/src/js/angular/graphexplore/controllers/graphs-visualizations.controller.js
@@ -141,6 +141,7 @@ function GraphsVisualizationsCtrl(
         max: MAX_LINKS_LIMIT,
         decreaseDisabled: false,
         increaseDisabled: false,
+        showingLimit: false,
     };
     $scope.invalidLimit = false;
     let linksLimitDebounceTimer = null;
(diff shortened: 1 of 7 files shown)


packages/legacy-workbench/src/js/angular/graphexplore/controllers/graphs-visualizations.controller.js
```diff
@@ -141,6 +141,7 @@ function GraphsVisualizationsCtrl(
         max: MAX_LINKS_LIMIT,
         decreaseDisabled: false,
         increaseDisabled: false,
+        showingLimit: false,
```
yordanalexandrov, line 144: The name is a bit odd.

yordanalexandrov: Perhaps `isShowingAllNodes` or `isMoreAvailable` or something

plamen-yordanov: > The name is a bit odd

The functionality is a bit odd. The message appears whenever the shown nodes number is equal to the current link limit. 

> Perhaps isShowingAllNodes or isMoreAvailable or something

We may have more, or we may not. We don't know if we're showing all nodes and we are not certain there are more

svilenvelikov (approved):

PR #2955, opened 2026-05-18
GDB-14452 fix license issue for open-location-code-typescript library

svilenvelikov:
## What
Graphdb build fails on CI because of incorrect license reported by the license-checker for the open-location-code-typescript library.

## Why
The library was added recently in the ontotext-yasgui component but it has incomplete packgae.json lacking a license property. The library has an Apache 2 license file but the missing property in the package.json tricks the license-checker to incorrectly report its license as Apache* which in result fails the graphdb build.

## How
Fixed by patching the generated report from the license checker with a correct value for the library.

## Checklist
- [ ] Branch name
- [ ] Target branch
- [ ] Commit messages
- [ ] Squash commits
- [ ] MR name
- [ ] MR Description
- [ ] Tests


svilenvelikov:
Closing this because another solution was designed

PR #2965, opened 2026-05-22
GDB-14399 introduce confirmation service provider

svilenvelikov:
## What
Introduce a wrapper for the primeng's confirmation service.

## Why
Hiding the primeng behind an abstraction layer for easier managment and minimizing the footprint that primeng has on the project.

## How
Implemented a provider service wrapper for the primeng's confirmation service.

Additionally:
Moved dialog service inside services/dialog package
Moved logger provider service inside services/logger package
Moved notification provider service inside services/notification package

## Testing
Added tests for the confirmation service.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/workbench/src/app/services/dialog/model/confirmation-config.ts
```diff
@@ -0,0 +1,17 @@
+export interface ConfirmationConfig {
```
yordanalexandrov, line 6: Maybe add some documentation here.

For example the target is not needed if showing a modal dialog confirmation and not attaching it to another element

packages/workbench/src/app/services/dialog/model/confirmation-config.ts
```diff
@@ -0,0 +1,17 @@
...
+  message: string;
+  header: string;
+  target?: EventTarget;
+  acceptButton?: {
+    label: string;
+    type?: string;
+    styleClass?: string;
```
yordanalexandrov, line 28: As this is our service with our config, do we need the `styleClass` prop as configuration?

yordanalexandrov: or type

svilenvelikov: There are places in the app where the confirmation dialog has specific marker classes on these buttons, so I wanted to allow the same here.

packages/workbench/src/app/services/dialog/model/confirmation-config.ts
svilenvelikov: I will

yordanalexandrov (dismissed):

boyan-tonchev (dismissed):

packages/workbench/src/app/pages/sparql-editor/sparql-editor-page.component.ts
```diff
@@ -87,9 +86,8 @@ interface UrlParametersAfterClear { repositoryId?: string, embedded?: boolean }
     MessageModule,
     PageLayoutComponent,
     YasguiComponentFacadeComponent,
-    ConfirmDialogModule
   ],
-  providers: [ConfirmationService],
+  providers: [ConfirmationProviderService],
```
plamen-yordanov: the ConfirmationProviderService service is provided in root from the class decorator, so this isn't needed. It will create an additional service instance as well

packages/workbench/src/app/services/dialog/confirmation-provider.service.ts
```diff
@@ -0,0 +1,40 @@
...
+   */
+  confirm(config: ConfirmationConfig) {
+    this.confirmationService.confirm({
+      target: config.target,
+      message: config.message,
+      header: config.header,
+      acceptButtonStyleClass: config.acceptButton?.styleClass || 'confirm-btn',
+      rejectButtonStyleClass: config.rejectButton?.styleClass || 'cancel-btn',
+      acceptButtonProps: {
+        label: config.acceptButton?.label || this.translocoService.translate('components.dialog.confirmation.confirm_btn'),
```
plamen-yordanov, line 29: according to the model `label` is required

yordanalexandrov: Hm.. an empty string would satisfy the requirement but maybe we should still have a default label.

plamen-yordanov: I was thinking just mark it as optional. I like the ability to override it

svilenvelikov: Changed to optional

packages/workbench/src/app/pages/sparql-editor/sparql-editor-page.component.ts
svilenvelikov: Indeed. Removing it

plamen-yordanov (approved):

PR #2976, opened 2026-05-27
GDB-14672 allow HTML formatting in guide content

plamen-yordanov:
## What
Allow HTML formatting in guide content.

## Why
Previously, HTML content was sanitized, preventing users from using formatting options

## How
- Removed the `sanitize` method from the `guideApi` and replaced it with `HtmlUtil.sanitize`
- Introduced a new `HtmlUtil` class that implements a whitelist-based HTML sanitizer, allowing specific tags and attributes while stripping out unsafe content.

## Testing
jest

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/services/utils/test/html-util.spec.ts
+++ b/packages/api/src/services/utils/test/html-util.spec.ts
@@ -1,0 +1,48 @@
+import {HtmlUtil} from '../html-util';
+
+describe('HtmlUtil', () => {
+  describe('sanitize', () => {
+    test('should return empty string for empty input', () => {
+      expect(HtmlUtil.sanitize('')).toBe('');
+    });
+
+    test('should pass plain text through unchanged', () => {
+      expect(HtmlUtil.sanitize('Hello World')).toBe('Hello World');
+    });
+
+    test('should strip script tags and their content', () => {
+      expect(HtmlUtil.sanitize('<script>alert(1)</script>')).toBe('');
+    });
+
+    test('should remove on* event handler attributes', () => {
+      expect(HtmlUtil.sanitize('<b onclick="evil()">text</b>')).toBe('<b>text</b>');
+    });
+
+    test('should remove javascript: hrefs', () => {
+      expect(HtmlUtil.sanitize('<a href="javascript:alert(1)">link</a>')).not.toContain('javascript:');
+    });
+
+    test('should not add target or rel to links without a target attribute', () => {
+      const result = HtmlUtil.sanitize('<a href="https://example.com">link</a>');
+      expect(result).not.toContain('target=');
+      expect(result).not.toContain('rel=');
+    });
+
+    test('should add rel="noopener noreferrer" to target="_blank" links', () => {
+      const result = HtmlUtil.sanitize('<a href="https://example.com" target="_blank">link</a>');
+      expect(result).toContain('target="_blank"');
+      expect(result).toContain('rel="noopener noreferrer"');
+    });
+
+    test('should preserve target value other than _blank without adding rel', () => {
+      const result = HtmlUtil.sanitize('<a href="https://example.com" target="_self">link</a>');
+      expect(result).toContain('target="_self"');
+      expect(result).not.toContain('rel=');
+    });
+
+    test('should accept a custom config', () => {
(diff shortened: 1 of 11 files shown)


packages/api/src/services/utils/test/html-util.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    test('should preserve allowed block tags', () => {
+      expect(HtmlUtil.sanitize('<p>paragraph</p>')).toBe('<p>paragraph</p>');
+      expect(HtmlUtil.sanitize('<ul><li>item</li></ul>')).toBe('<ul><li>item</li></ul>');
+      expect(HtmlUtil.sanitize('<ol><li>item</li></ol>')).toBe('<ol><li>item</li></ol>');
+      expect(HtmlUtil.sanitize('<blockquote>quote</blockquote>')).toBe('<blockquote>quote</blockquote>');
+    });
+
+    test('should strip tags, which are not explicitly allowed', () => {
+      expect(HtmlUtil.sanitize('<div>content</div>')).toBe('content');
```
boyan-tonchev: Why do we sanitize elements like "div" or "h1" ... They are not dangerous by themselves.

packages/api/src/services/utils/test/html-util.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      expect(HtmlUtil.sanitize('<h1>heading</h1>')).toBe('heading');
+      expect(HtmlUtil.sanitize('<section>text</section>')).toBe('text');
+    });
+
+    test('should strip <script> and preserve its text content', () => {
+      expect(HtmlUtil.sanitize('<script>alert(1)</script>')).toBe('alert(1)');
```
boyan-tonchev: IMO Dangerous elements like <script> should be removed entirely instead of unwrapped,
because their content is executable code and should not be rendered as plain text.

packages/api/src/services/utils/test/html-util.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    test('should remove javascript: hrefs', () => {
+      expect(HtmlUtil.sanitize('<a href="javascript:alert(1)">link</a>')).toBe('<a>link</a>');
+    });
+
+    test('should remove javascript: hrefs with surrounding whitespace', () => {
+      expect(HtmlUtil.sanitize('<a href="  javascript:alert(1)">link</a>')).toBe('<a>link</a>');
+    });
+
+    test('should remove disallowed attributes from allowed tags', () => {
+      expect(HtmlUtil.sanitize('<strong style="color:red">text</strong>')).toBe('<strong>text</strong>');
```
boyan-tonchev: Why do we sanitize the `style` attribute?

boyan-tonchev:
IMO, we should use a library like DOMPurify.

packages/api/src/services/utils/test/html-util.spec.ts
plamen-yordanov: Since this is used to sanitize guides only for now, I don't think we need complex html in the json files. That being said, it is configurable, so we can allow whatever we need, if need be

packages/api/src/services/utils/test/html-util.spec.ts
plamen-yordanov: agree. I was thinking about this. I'll change it

packages/api/src/services/utils/test/html-util.spec.ts
plamen-yordanov: I don't think styling should be coming from the guide json file, but again, this is configurable and can be changed if needed

plamen-yordanov:
> IMO, we should use a library like DOMPurify.

I was avoiding adding additional dependencies, but if @svilenvelikov is ok, I'll use that instead

packages/api/src/services/utils/test/html-util.spec.ts
boyan-tonchev: It is a utility, so it can be used anywhere.

boyan-tonchev (approved):

yordanalexandrov (approved):

PR #2977, opened 2026-05-28
GDB-14579: Migrate JSON-LD download settings dialog

yordanalexandrov:
## What
Migrated JSON-LD download settings dialog for configuring JSON-LD export settings, including form selection and context links.

## Why
This feature enhances user experience by allowing users to customize their JSON-LD export settings directly from the application, improving usability and flexibility.

## How
- Created `DownloadSettingsDialogComponent`.
- Implemented form handling for JSON-LD modes and context links.
- Integrated local storage service to persist user settings.
- Added translations for dialog labels and tooltips in `en.json` and `fr.json`.
- Updated application settings model to include JSON-LD export settings.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/api/src/services/application-settings/application-settings-storage.service.ts
+++ b/packages/api/src/services/application-settings/application-settings-storage.service.ts
@@ -1,6 +1,7 @@
 import {LocalStorageService} from '../storage';
 import {ApplicationSettings, LegacySettings, ThemeMode} from '../../models/application-settings/application-settings';
 import {LoggerProvider} from '../logging/logger-provider';
+import {JsonldExportSettings} from '../../models/application-settings/jsonld-export-settings';
 
 /**
  * Service to manage the storage of application settings in the local storage.
@@ -101,5 +102,16 @@ export class ApplicationSettingsStorageService extends LocalStorageService {
       return settingsData.themeMode !== undefined;
     }
     return false;
+  }
+
+  getJsonLDExportSettings(): JsonldExportSettings | undefined {
+    const settings = this.getApplicationSettings();
+    return settings.jsonLdExportSettings;
+  }
+
+  setJsonLDExportSettings(jsonLdExportSettings: JsonldExportSettings): void {
+    const settings = this.getApplicationSettings();
+    settings.jsonLdExportSettings = jsonLdExportSettings;
+    this.setApplicationSettings(settings);
   }
 }
--- a/packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
+++ b/packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
@@ -1,0 +1,63 @@
+<div class="dialog-content">
+  <form class="export-settings-form">
+    <div class="form-row">
+      <label for="jsonldMode">
+        {{ 'components.dialog.download_settings.form.label' | transloco }}
+        <span class="info-icon"
+              [pTooltip]="'components.dialog.download_settings.form.info' | transloco"
+              tooltipEvent="both"
+              tooltipPosition="right">
+            <em class="icon ri-information-2-line"></em>
+          </span>
+      </label>
+      <div class="field-input">
+        <p-select
+          inputId="jsonldMode"
+          name="jsonldMode"
+          [(ngModel)]="currentMode"
+          [options]="JSONLDModeOptions"
+          styleClass="input"
+          appendTo="body"
+          (onChange)="clearLinkInput($event)"/>
+      </div>
+    </div>
+
+    @if (isFramedMode()) {
+      <div class="form-row">
+        <label for="jsonldFrame">
+          {{ 'components.dialog.download_settings.frame.link' | transloco }}
+          <span class="info-icon"
+                [pTooltip]="'components.dialog.download_settings.frame.info' | transloco"
+                tooltipEvent="both"
+                tooltipPosition="right">
+              <em class="icon ri-information-2-line"></em>
+            </span>
+        </label>
+        <div class="field-input">
+          <input pInputText id="jsonldFrame" name="jsonldFrame" type="text"
+                 [(ngModel)]="link"
+                 [placeholder]="'components.dialog.download_settings.frame.uri_placeholder' | transloco"/>
+        </div>
+      </div>
+    }
+
+    @if (isContextMode()) {
+      <div class="form-row">
+        <label for="jsonldContext">
+          {{ 'components.dialog.download_settings.context.link' | transloco }}
+          <span class="info-icon"
+                [pTooltip]="'components.dialog.download_settings.context.info' | transloco"
+                tooltipEvent="both"
+                tooltipPosition="right">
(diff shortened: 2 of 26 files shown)


packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
```diff
@@ -0,0 +1,80 @@
...
+        </div>
+      </div>
+    }
+  </form>
+
+  <div class="actions">
+    <p-button
+      class="restore-defaults"
+      [label]="'components.dialog.download_settings.actions.restore_defaults' | transloco"
+      [link]="true"
```
svilenvelikov: This link button doesn't look very impressive ATM. I think we need to style it additionally.

packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
```diff
@@ -0,0 +1,80 @@
...
+        <span class="info-icon"
+              [pTooltip]="'components.dialog.download_settings.form.info' | transloco"
+              tooltipEvent="both"
+              tooltipPosition="right">
+            <em class="icon ri-information-2-line"></em>
+          </span>
+      </label>
+      <div class="field-input">
+        <p-select
+          inputId="wb-JSONLD-mode"
```
svilenvelikov: These are strangely formatted ID's. Would you change them to camelcase please.
Also, it is a good practice to put same name for the id and the name attributes.

packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
```diff
@@ -0,0 +1,80 @@
+<div class="dialog-content">
+  <form class="export-settings-form">
+    <div class="form-row">
+      <label for="wb-JSONLD-mode">
+        {{ 'components.dialog.download_settings.form.label' | transloco }}
+        <span class="info-icon"
+              [pTooltip]="'components.dialog.download_settings.form.info' | transloco"
```
svilenvelikov, line 7: This is another topic for discussion. Are we going to use the primeng's tooltip or our onto-tooltip?
If we decided to use the primeng's one, then we need to style both to look one and the same.

packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
```diff
@@ -0,0 +1,80 @@
...
+          name="jsonldMode"
+          [(ngModel)]="currentMode"
+          [options]="JSONLDModeOptions"
+          styleClass="input"
+          appendTo="body"
+          (onChange)="clearLinkInput()"/>
+      </div>
+    </div>
+
+    @if (currentMode && JSONLDFramedModes.indexOf(currentMode) > -1) {
```
svilenvelikov: I would prefer a more declarative approach here: eg. either use declarative form or having a UI model with proper flags to control the visibility and other field state where needed instead of having such conditions in the template

packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
```diff
@@ -0,0 +1,80 @@
...
+        <div class="field-input">
+          <input pInputText id="wb-JSONLD-context" name="jsonldContext" type="text"
+                 [(ngModel)]="link"
+                 [placeholder]="'components.dialog.download_settings.context.uri_placeholder' | transloco"/>
+        </div>
+      </div>
+    }
+  </form>
+
+  <div class="actions">
```
svilenvelikov: I wonder, why you put the buttons inside the dialog body? It would be better if we use their template system and always put these buttons in the dialog #footer

packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.scss
```diff
@@ -0,0 +1,56 @@
...
+  display: grid;
+  grid-template-rows: auto;
+  gap: 1rem;
+
+  .export-settings-form {
+    display: grid;
+    grid-template-rows: auto;
+    gap: 1rem;
+
+    .form-row {
```
svilenvelikov, line 11: As we spoke, we will introduce some utilities (a library) and have these coming from there.
Leave this as is for now. We will refactor it later

svilenvelikov (changes requested):

packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
yordanalexandrov: Added some additional styling for the link button.

packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
yordanalexandrov: Changed to camelcase.

packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
yordanalexandrov: PrimeNG and Onto tooltips seems almost the same. There are some DS vars that need to be change to improve the PrimeNG one.

packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
yordanalexandrov: Changed

packages/workbench/src/app/components/download-settings-dialog/download-settings-dialog.component.html
yordanalexandrov: Because the way `DynamicDialog` works if no very nice. It does not use templates and you you must make a separate component for the footer which has no idea about the state of the dialog. There must be a mechanism to bridge them.

I was wondering about this when implementing.

I have made an implementation for this dialog only. If we approve it I think we can make a dynamic footer component for a default footer with declaration of the buttons that will be available. It should be a separate task though.

packages/api/src/services/application-settings/application-settings-storage.service.ts
```diff
@@ -102,4 +103,15 @@ export class ApplicationSettingsStorageService extends LocalStorageService {
     }
     return false;
   }
+
+  getJsonLDExportSettings(): JsonldExportSettings | undefined {
+    const settings = this.get(this.SETTINGS).getAsJson() as Partial<ApplicationSettings>;
```
svilenvelikov: Why didn't you use the exisitng getApplicationSettings method?

yordanalexandrov: I guess it was autosuggested and I overlooked it. I will change it

svilenvelikov (approved):

PR #2978, opened 2026-05-28
GDB-14677 add target to loader component

plamen-yordanov:
## What
Add target selector to `onto-loader` component.

## Why
To decouple the loader from the layout and simplify the layout altogether

## How
- Introduced a new `targetSelector` property in `onto-loader` to specify the DOM element for attachment.
- Attached the loader to the `main` element
- Removed the main slot wrapper, the tooltip and the dialog from the layout to simplify them.
- Updated styles

## Testing
n/a

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [ ] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/shared-components/src/components/onto-layout/onto-layout.scss
+++ b/packages/shared-components/src/components/onto-layout/onto-layout.scss
@@ -41,20 +41,17 @@
     margin-left: 0;
   }
 
-  .main-slot-wrapper {
-    .loader-wrapper {
-      flex-grow: 1;
-      display: flex;
-      align-items: center;
-      justify-content: center;
-    }
+  main {
+    grid-area: main;
+  }
 
-    &.loading {
-      display: flex;
-      flex-direction: column;
+  &.loading {
+    main {
+      position: relative;
+      visibility: hidden;
 
-      .page-content {
-        visibility: hidden;
+      onto-loader {
+        visibility: visible;
       }
     }
   }
@@ -75,10 +72,6 @@
 
   header {
     grid-area: header;
-  }
-
-  main {
-    grid-area: main;
   }
 
   footer {
--- a/packages/shared-components/src/components/onto-layout/onto-layout.scss
+++ b/packages/shared-components/src/components/onto-layout/onto-layout.scss
@@ -49,10 +49,6 @@
     main {
       position: relative;
       visibility: hidden;
-
-      onto-loader {
-        visibility: visible;
-      }
     }
   }
 
--- a/packages/shared-components/src/components/onto-layout/onto-layout.tsx
+++ b/packages/shared-components/src/components/onto-layout/onto-layout.tsx
@@ -129,10 +129,10 @@ export class OntoLayout {
           </nav>
         }
 
-        {this.loading &&
-          <onto-loader targetSelector={'main'} size={96} messageText={TranslationService.translate('common.loading')}></onto-loader>
-        }
-        <slot name="main"></slot>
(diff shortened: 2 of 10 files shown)


packages/shared-components/src/components/onto-layout/onto-layout.scss
```diff
@@ -41,20 +41,17 @@
...
-      flex-direction: column;
+  &.loading {
+    main {
+      position: relative;
+      visibility: hidden;
 
-      .page-content {
-        visibility: hidden;
+      onto-loader {
+        visibility: visible;
```
svilenvelikov: I can't  remember if we've discussed this back then when this loader component was introduced here, but IMO it's not OK to be styled this way here. Neither the new styling nor the old one is a good approach considering the fact that this component is intended to be used in different contexts as the loader in the legacy.

packages/shared-components/src/components/onto-layout/onto-layout.tsx
```diff
@@ -127,20 +128,16 @@ export class OntoLayout {
...
         }
 
-        <div class={{'main-slot-wrapper': true, 'loading': this.loading}}>
-          {this.loading && <div class="loader-wrapper">
-            <onto-loader size={96} messageText={TranslationService.translate('common.loading')}></onto-loader>
-          </div>}
-          <slot name="main"></slot>
-        </div>
+        {this.loading &&
+          <onto-loader targetSelector={'main'} size={96} messageText={TranslationService.translate('common.loading')}></onto-loader>
```
svilenvelikov: IMO the best is if the loader component accepts the loading flag as an input and manages it's  own visibility instead of controlling and styling it on each place where it is used.

packages/shared-components/src/components/onto-layout/onto-layout.scss
plamen-yordanov: moved inside the loader component

packages/shared-components/src/components/onto-layout/onto-layout.tsx
plamen-yordanov: added a `loading` prop to the component

packages/shared-components/src/components/onto-loader/onto-loader.tsx
```diff
@@ -1,32 +1,66 @@
...
+  private readonly logger = LoggerProvider.logger;
+
+  componentDidLoad() {
+    this.attachToSelector();
+  }
 
   render() {
     return (
-      <Host>
+      <Host class={{[this.ATTACHED_CLASS]: !!this.targetSelector, 'onto-loader-active': this.loading}}>
```
svilenvelikov, line 43: eventually extract a constant for `'onto-loader-active'` as well

packages/shared-components/src/components/onto-loader/onto-loader.tsx
```diff
@@ -1,32 +1,66 @@
...
     );
   }
+
+  private attachToSelector() {
+    if (!this.targetSelector) {
+      return;
+    }
+
+    const target = document.querySelector<HTMLElement>(this.targetSelector);
+    if (!target) {
```
svilenvelikov, line 58: One minor optional devex improvement would be if you also check if the target is has a position and at least print a warning.

svilenvelikov (dismissed):

svilenvelikov (approved):

PR #2986, opened 2026-06-01
GDB-14684 install and configure bootstrap

svilenvelikov:
## What
Install and configure bootstrap in the new workbench Angular app.

## Why
Primeng components that we use in the new workbench are styled by our styleguide, but outside of them, on a page level we usually need to apply some additional styles. To style the new workbench application consistently we need some utility css classes and helpers for easy creating layouts, grids, controlling spacing, etc. That's why we introduce the bootstrap library and scope it to work in the new angular app only.

## How
- Added `bootstrap` dependency in `packages/workbench`.
- Imported selected Bootstrap SCSS partials in `packages/workbench/src/styles.scss` under `app-root` scope.
- Resolved merge conflicts with latest `master` updates in `styles.scss` and kept both:
  - Bootstrap scoped imports
  - `@use "styles/primeng-overrides";`

## Testing
- Reproduced and resolved merge conflicts locally by merging latest `master` and latest remote PR branch.
- Ran targeted checks/build in `packages/workbench` to validate the merged stylesheet changes.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/workbench/src/styles.scss
+++ b/packages/workbench/src/styles.scss
@@ -1,2 +1,21 @@
 /* You can add global styles to this file, and also import other style files */
 @use "styles/primeng-overrides";
+
+app-root {
+  @import "../node_modules/bootstrap/scss/functions";
+  // Required Bootstrap imports
+  @import "../node_modules/bootstrap/scss/variables";
+  @import "../node_modules/bootstrap/scss/variables-dark";
+  @import "../node_modules/bootstrap/scss/maps";
+  @import "../node_modules/bootstrap/scss/mixins";
+  // Probably not needed and leaking properties to :root
+  //@import "../node_modules/bootstrap/scss/root";
+  // Optional components
+  @import "../node_modules/bootstrap/scss/utilities";
+  // Probably not needed as well
+  //@import "../node_modules/bootstrap/scss/reboot";
+  @import "../node_modules/bootstrap/scss/containers";
+  @import "../node_modules/bootstrap/scss/grid";
+  @import "../node_modules/bootstrap/scss/helpers";
+  @import "../node_modules/bootstrap/scss/utilities/api";
+}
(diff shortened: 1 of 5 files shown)


packages/workbench/src/styles.scss
```diff
@@ -1 +1,20 @@
 /* You can add global styles to this file, and also import other style files */
+
+app-root {
+  @import "../node_modules/bootstrap/scss/functions";
```
yordanalexandrov, line 5: Wrap those in their own file and do only one import

svilenvelikov: What's the point?

yordanalexandrov: To be cleaner and easier to remove/change. There is nothing else right now, but the `styles.scss` is the global styling file for angular, so it might get other things like PrimeNG overrides, etc.

svilenvelikov: This is the only place where  these will be imported, so I'll leave them here.

svilenvelikov:
@copilot resolve the merge conflicts in this pull request

packages/workbench/src/styles.scss
```diff
@@ -1,2 +1,22 @@
...
+  @import "../node_modules/bootstrap/scss/utilities";
+  // Probably not needed as well
+  //@import "../node_modules/bootstrap/scss/reboot";
+  @import "../node_modules/bootstrap/scss/containers";
+  @import "../node_modules/bootstrap/scss/grid";
+  @import "../node_modules/bootstrap/scss/helpers";
+  @import "../node_modules/bootstrap/scss/utilities/api";
+}
+
 @use "styles/primeng-overrides";
```
yordanalexandrov: Oh look, something else is imported here as well.

Also isn't `@import` depricated?

yordanalexandrov (dismissed):

PR #2988, opened 2026-06-02
GDB-13816: Update Angular to v21 and adjust configurations

yordanalexandrov:
## What
Updated Angular dependencies to version 21 and modified related configuration files.

## Why
This update ensures compatibility with the latest Angular features and improvements, enhancing the overall performance and maintainability of the application.

## How
- Changed Angular dependencies in package.json to version 21.x
- Updated test configuration in angular.json to include 'zoneless' option
- Modified tsconfig.spec.json to include 'node' types and updated module settings
- Enabled the main menu in plugin.js by setting 'disabled' to false
- update `single-spa-angular` to v21.0.1
- migrated changes to latest version
- Add CORS header to webpack-dev-server for ES module compatibility
  - ES modules (used since single-spa-angular v21) enforce CORS on cross-origin fetches, unlike classic scripts which do not. Without this header, browsers block workbenchApp.js when loaded by root-config from a different port.
- update transloco and jsdom

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/workbench/tsconfig.spec.json
```diff
@@ -4,12 +4,16 @@
...
+      "jest",
+      "node"
     ],
     "paths": {
       "@ontotext/workbench-api": ["../api/dist/ontotext-workbench-api.d.ts"],
       "@ontotext/shared-components": ["../shared-components/dist/types/index.d.ts"]
-    }
+    },
+    "module": "Node16",
+    "moduleResolution": "Node16",
```
svilenvelikov: Probably these should point to something more recent, but I've read that this is not mandatory.

svilenvelikov (dismissed):

packages/workbench/tsconfig.spec.json
yordanalexandrov: Updated to recommended values

svilenvelikov (approved):

PR #2994, opened 2026-06-03
GDB-12880 introduce repository picker list

svilenvelikov:
## What
Introduce repository picker list component.

## Why
This is a component which is accessible on all pages that depend on having an active repository.

## How
Implemented a repository picker list component that allows the user to select a repository, filter available local or remote repositories or quick access the repository create page.

Additionally:
* Added responsive breakpoints and updated media queries in styles
* Improved the primeng's link button styling by removing the underline from icons if any
* Removed the lead class from the alert components in the core-error directive. It enforces a larger font-size and weight which makes everything inside the component bigger and inconsistent. This was probably added by some mistake long time ago.
* Added a new stateLowercase property to Repository model. This can be used as a marker css class inside templates for example without transformation.
* Added a section with primeng message components in the ux-test page

## Testing
Implemented tests.

## Screenshots
image
image
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/api/src/models/repositories/repository-list.ts
```diff
@@ -22,6 +22,10 @@ export class RepositoryList extends ModelList<Repository> {
     super(repositories);
   }
 
+  filterWithCallback(filterCallback: (repository: Repository) => boolean): RepositoryList {
```
yordanalexandrov, line 26: Why is this called `filterCallback`? Isn't filter informative enough?

yordanalexandrov: Is there a collision with the super's names?

packages/api/src/services/domain/repository/mappers/repository-list.mapper.ts
```diff
@@ -28,6 +28,7 @@ export const mapRepositoryListResponseToModel: MapperFn<RepositoryListResponse,
           externalUrl: repositoryData.externalUrl,
           location: repositoryData.location,
           state: toEnum(RepositoryState, repositoryData.state),
+          stateLowercase: repositoryData.state ? repositoryData.state.toLowerCase() : undefined,
```
yordanalexandrov, line 31: Why set this here? Isn't it more apropriate to set it in the constructor? It is a helper prop.

packages/api/src/models/repositories/repository-list.ts
yordanalexandrov: Ok, so .. what it does different than the already inherited `filter` method is that  it returns a new List with the values. I think `filterWithCallback` should be renamed to something like `filteredList`.

packages/api/src/services/domain/security/authorization.service.ts
```diff
@@ -296,6 +299,46 @@ export class AuthorizationService implements Service {
...
+  getReadableRepositories(): RepositoryList {
+    return this.repositoryContextService.getRepositoryList().filterWithCallback(repository =>
+      this.canReadRepo(repository) || this.canReadGqlRepo(repository)
+    );
+  }
+
+  /**
+   * Retrieves a list of repositories that the user has access to, based on the specified parameters.
+   * @param includeRemote - If true, includes remote repositories in the list; otherwise, only local repositories are included.
+   * @param isRestricted - If true, returns repositories that the user has write access to; if false, returns repositories that the user has read access to.
```
yordanalexandrov: This is kind of strange. If `isRestricted` is `true` I get unrestricted access repos (write/read) and if it is false, I am restricted only to read access.

packages/workbench/src/app/components/page-layout/page-layout.component.scss
```diff
@@ -15,7 +15,7 @@
     gap: 0.5rem;
   }
 
-  .page-banners {
+  app-page-restrictions {
     display: block;
```
yordanalexandrov, line 19: Perhaps the `display: block` should be in the component itself `:host`

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.html
```diff
@@ -0,0 +1,18 @@
+<div class="page-restrictions">
+  @if (restrictions().length) {
+    <p-message [severity]="'info'" icon="ri-alert-line">
+      @for (reason of restrictions(); track reason.translationKey) {
```
yordanalexandrov, line 4: not big deal but strange

shouldn't the `restrictions()` give back a `restriction` which might/might not have a `reason`?

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.restrictionResolverService.resolve({
+      selectedRepository: this.selectedRepository(),
+      isRestricted: this.isRestricted(),
+      pageTitle: this.title() ?? 'Missing page title',
+    })
+  );
+
+  ngOnInit(): void {
+    this.selectedRepository.set(this.repositoryContextService.getSelectedRepository());
+    this.repositoryList.set(this.repositoryContextService.getRepositoryList());
```
yordanalexandrov: just wondering if any case that will change the repo list context dynamically

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.html
```diff
@@ -1 +1,81 @@
-<p>repository-picker-list works!</p>
+<div class="repository-picker-list">
+  <div class="toolbar">
+    @if (vm().showFilters) {
```
yordanalexandrov, line 8: it took me a while to realize what `vm` means

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.ts
```diff
@@ -1,10 +1,132 @@
...
+    this.subscriptions.unsubscribeAll();
+  }
+
+  /**
+   * Navigates to the repository creation page, passing the current page as a query parameter to allow returning to it
+   * after repository creation.
+   */
+  createRepository(): void {
+    const currentPage = WindowService.getLocationPathname();
+    navigate(`/repository/create?previous=${currentPage}`);
```
yordanalexandrov, line 112: Maybe remove the leading `/`, I think it won't work with context path

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.ts
```diff
@@ -1,10 +1,132 @@
...
-export class RepositoryPickerListComponent {
+export class RepositoryPickerListComponent implements OnInit, OnDestroy {
+  private readonly repositoryContextService = service(RepositoryContextService);
+  private readonly licenseContextService = service(LicenseContextService);
+  private readonly authorizationService = service(AuthorizationService);
 
+  /**
+   * If the user has permissions to manage repositories for example.
+   */
+  isRestricted = input<boolean>(true);
```
yordanalexandrov: This doesn't seem to do what it should do. It is always true. Is this correct?

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.view-model.ts
```diff
@@ -0,0 +1,91 @@
...
+   */
+  get filteredRepositoryList(): Repository[] {
+    const query = this.filterQuery.trim().toLowerCase();
+    const state = this.stateFilter;
+    return this.repositoryList.filter((repo) => {
+      const matchesQuery =
+        !query ||
+        repo.id?.toLowerCase().includes(query) ||
+        repo.title?.toLowerCase().includes(query);
+      const matchesState = state === null || repo.state === state;
```
yordanalexandrov: It will be more readable if this is extracted in another `filter` pipe.

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.html
yordanalexandrov: I saw it is actually `RestrictionReason` so it makes sense for the var to be `reason` if `restrictions` is actually `restrictionReasons`

packages/workbench/src/app/services/page-restrictions/restriction-reason.ts
```diff
@@ -0,0 +1,9 @@
+export type RestrictionSeverity = 'info' | 'warn';
+
+export interface RestrictionReason {
+  severity: RestrictionSeverity;
+  translationKey: string;
+  translationParams?: Record<string, string>;
+  ctaKey?: string;
```
yordanalexandrov, line 7: Please add a comment waht `cta` means.

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
```diff
@@ -0,0 +1,103 @@
...
+    const isSecurityEnabled = this.securityContextService.getSecurityConfig()?.isEnabled() ?? false;
+    const canWrite = this.authorizationService.canWriteRepo(repo);
+    const canManage = this.authorizationService.isRepoManager();
+    const accessibleRepositoriesCount = this.authorizationService.getAccessibleRepositories(true, ctx.isRestricted).size();
+    const hasAccessibleRepositories = accessibleRepositoriesCount > 0;
+
+    const reasons: RestrictionReason[] = [];
+
+    if (!repo && isLicenseValid) {
+      if (canManage && !hasAccessibleRepositories) {
```
yordanalexandrov, line 39: missing case `(!canManage && !hasAccessibleRepositories)`

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
```diff
@@ -0,0 +1,103 @@
...
+      } else if (!canManage && hasAccessibleRepositories) {
+        reasons.push({
+          severity: 'info',
+          translationKey: 'components.page_restrictions.no_active_repository_select_one',
+        });
+      }
+    }
+
+    if (repo) {
+      if (isSecurityEnabled && !canWrite) {
```
yordanalexandrov, line 58: I believe a user can have only read with the security off. This would be by using the override user from config. @plamen-yordanov can confirm if this is possible.

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
```diff
@@ -0,0 +1,103 @@
...
+      reasons.push({
+        severity: 'warn',
+        translationKey: 'components.page_restrictions.invalid_license',
+        ctaKey: 'components.page_restrictions.set_new_license',
+        ctaLink: '/license',
+      });
+    }
+
+    if (!hasAccessibleRepositories) {
+      if (ctx.isRestricted) {
```
yordanalexandrov, line 93: I am now totally baffled by what `isRestricted` means in all the contexts

packages/workbench/src/styles/primeng/_button.scss
```diff
@@ -1,6 +1,12 @@
 .p-button.p-button-link {
-  text-decoration-line: underline;
-  text-decoration-style: dotted;
-  text-decoration-thickness: var(--link-decoration-thickness);
-  text-underline-offset: var(--link-underline-offset);
+  color: var(--gw-button-primary-link-hover-color);
```
yordanalexandrov, line 2: This is not right. There should be a `--gw-button-primary-link-color` var

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.scss
```diff
@@ -0,0 +1,113 @@
+@use 'src/styles/variables' as *;
```
yordanalexandrov, line 1: Is this actually used here?

packages/workbench/src/styles/_variables.scss
```diff
@@ -0,0 +1,7 @@
+// Responsive breakpoints
```
yordanalexandrov, line 1: Wouln't it be better if those were actually defined in the DS?

Or at least as css vars, so they can be accessed everywhere.

packages/api/src/models/repositories/repository-list.ts
svilenvelikov: OK. The argument can be renamed to `filter` I guess. My idea was somewhat I've seen in lodash for example where they have functions like `unionWith`, `uniqueWith`, etc. where those take a callback/comparator function but made it a bit more verbose.

packages/api/src/services/domain/repository/mappers/repository-list.mapper.ts
svilenvelikov: Yes this should be in the constructor as it is a derived property. I used to change it, but I guess I've dropped it somewhere down the road

packages/api/src/services/domain/security/authorization.service.ts
svilenvelikov: yeah, this was a blind copy/paste from the core-errors. It's stupid indeed.

packages/workbench/src/app/components/page-layout/page-layout.component.scss
svilenvelikov: why? It's the component client's decision how to layout its children. I know we often put in the components stuf like `:host { display: block; width: 100% }` etc. but this kind of limits their flexibility.

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.html
svilenvelikov: its the view model

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.ts
svilenvelikov: yeah, good catch

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.ts
svilenvelikov: No. It's not. As I already mentioned the other day, this should stay hardcodded while the task for the `isRestricted` gets resolved.

packages/workbench/src/app/services/page-restrictions/restriction-reason.ts
svilenvelikov: I wasn't sure how exactly to name this. Donka usually use the term `call to action` for such elements. It needs a documentation indeed.

yordanalexandrov: I would've never thought about `call to action`. Maybe just action is ok.

packages/api/src/models/repositories/repository-list.ts
yordanalexandrov: I agree with the lodash functions, but here you already have `filter` which does the same thing, only difference is that here you return a new reference for a list.

If you need both: to get a list and to get an array, perhaps this shoud be a `.toFilteredList` or something like this.

packages/workbench/src/styles/primeng/_button.scss
svilenvelikov: yeah, but it has wrong color and I set this until the other one gets fixed. I probably shouldn't have to.

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.html
yordanalexandrov: Yes I know, but havn't seen this pattern with such naming in the WB and was caught off guard :)

packages/workbench/src/styles/_variables.scss
svilenvelikov: The design system should not declare responsive breakpoints I think. This is the application's responsibility IMO

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.scss
svilenvelikov: It was. I used to use the responsive-breakpoints in some initial versions, but then this was changed and I forgot to delete this

packages/workbench/src/app/app.routes.ts
```diff
@@ -1,14 +1,15 @@
...
 export const routes: Routes = [
   {
     path: 'sparql-new',
     data: {
       title: 'sparql_editor.title',
       helpInfo: 'sparql_editor.helpInfo',
-      documentationUrl: 'sparql-queries.html'
-    },
+      documentationUrl: 'sparql-queries.html',
+    } satisfies WorkbenchRouteData,
```
yordanalexandrov, line 12: This is not ok as you will need to add it to all routes.

Create a WB routes models.

Something like this:
```
type WbRouteData = {
  title: string;
  helpInfo: string;
  documentationUrl: string;
};

interface WbRoute extends Route {
  data?: WbRouteData;
}

type WbRoutes = WbRoute[];

export const routes: WbRoutes = [
  {
```

packages/api/src/services/domain/repository/mappers/repository-list.mapper.ts
boyan-tonchev: Moved to the constructor

packages/api/src/models/repositories/repository-list.ts
boyan-tonchev: renamed to filterAsList

packages/api/src/services/domain/security/authorization.service.ts
boyan-tonchev: I renamed the parameter to requireWriteAccess

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.ts
boyan-tonchev: Update both with subscriptions to changes.

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.view-model.ts
boyan-tonchev: Extracted

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
boyan-tonchev: Implemented

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
boyan-tonchev: This will be rearranged after the isRestricted issue is implemented.

packages/workbench/src/app/services/page-restrictions/restriction-reason.ts
boyan-tonchev: rename ctaKey/ctaLink to actionLabelKey/actionLink, and add isExternalAction to distinguish in-app routes (routerLink) from external links (href), and document the fields with comments.

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.scss
boyan-tonchev: removed

packages/workbench/src/app/app.routes.ts
boyan-tonchev: Done

packages/workbench/src/styles/primeng/_button.scss
boyan-tonchev: Removed. It is applied to --gw-button-link-color.

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.ts
boyan-tonchev: Removed. Also, fetching the current location using `WindowService.getLocationPathname() `was changed to use the `getPathName` function, which returns the URL without the context path.

yordanalexandrov:
I have a few comments. I will raise an issue about them so we can comment on them.
* The @ Local is always visible, which IMO does not need to be, when `Local only` is checked (as old picker
* The typography is not good. IMO the weight of the name has to come down and the discription and location should be muted.
* * I don't like that there is a relativly small special select button that you have to targeet to select a repository. IMO the whole line should be selectable.

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.ts
```diff
@@ -0,0 +1,144 @@
...
+  readonly title = input<string>();
+
+  private readonly restrictionResolverService = inject(RestrictionResolverService);
+  private readonly repositoryContextService = service(RepositoryContextService);
+  private readonly elementRef = inject(ElementRef<HTMLElement>);
+  private resizeObserver?: ResizeObserver;
+
+  repositoryList = signal<RepositoryList | undefined>(undefined);
+  selectedRepository = signal<Repository | undefined>(undefined);
+  isRestricted = signal<boolean>(false);
```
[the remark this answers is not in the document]
yordanalexandrov: This is implemented with anotoher issue in an upcoming PR.

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.scss
```diff
@@ -0,0 +1,129 @@
...
+  .repository-list {
+    flex: 1;
+    min-height: 200px;
+    border: 1px solid var(--gw-datatable-border-color);
+
+    .actions-column {
+      width: 7rem;
+
+      .select-repository-btn {
+        visibility: hidden;
```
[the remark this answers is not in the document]
yordanalexandrov: This is an accessability issue which should be taken into account. I will create a new issue as we might have to rethink the design.

packages/root-config/scripts/validate-translations.js
```diff
@@ -9,6 +9,7 @@ const packagesDir = path.join(baseDir, 'packages');
 // Verified identical translations - no warnings or errors about them
 const identicalTranslations = [
   "#",
+  "Select",
```
yordanalexandrov: Why is this here?
The french translation is not the same

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+      .compileComponents();
+
+    fixture = TestBed.createComponent(PageRestrictionsComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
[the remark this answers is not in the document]
yordanalexandrov: I will open an issue for adding tests

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.ts
```diff
@@ -1,10 +1,210 @@
...
+  /**
+   * Returns the repositories accessible to the current user after applying all
+   * configured restrictions.
+   *
+   * @returns The filtered collection of accessible repositories.
+   */
+  private getAllowedRepositories() {
+    return this.authorizationService.getAccessibleRepositories(true, this.requireWriteAccess())
+      .filterByType(this.filterByTypes())
+      .filter(this.createRepositoryPermissionFilter());
```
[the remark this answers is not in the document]
yordanalexandrov: I will open an issue for adding tests

packages/workbench/src/testing-utils/resize-observer-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+
+      unobserve(target: Element): void {
+        const targets = observedTargets.get(target);
+        if (targets) {
+          observedTargets.set(target, targets.filter((entry) => entry.callback !== this.callback));
+        }
+      }
+
+      disconnect(): void {
+        observedTargets.forEach((targets, target) => {
```
yordanalexandrov: The second argument should be `element` not `target`

yordanalexandrov: or `targetElement`

packages/api/src/services/domain/security/authorization.service.ts
```diff
@@ -295,6 +298,46 @@ export class AuthorizationService implements Service {
     return !!config && config.hasAdditionalAuthSource() && config.hasLocalAdditionalAuthSources();
   }
 
+  /**
+   * Retrieves a list of repositories that the user has write access to.
+   *
+   * @returns {RepositoryList} A list of repositories that the user can write to.
+   */
+  getWritableRepositories(): RepositoryList {
+    return this.repositoryContextService.getRepositoryList().filterAsList(repository => this.canWriteRepo(repository));
```
plamen-yordanov, line 307: the legacy check is `canWriteRepo(repo) && !that.isActiveRepoOntopType(repo)`. Either add it, or add a todo, since it is similar to the isRepoManager situation

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
yordanalexandrov: This should be revisited in the other MR for the restrictions

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
```diff
@@ -0,0 +1,107 @@
...
+  private readonly securityContextService = service(SecurityContextService);
+  private readonly authorizationService = service(AuthorizationService);
+
+  resolve(ctx: RestrictionContext): RestrictionReason[] {
+    const repo = ctx.selectedRepository;
+    const isLicenseValid = this.licenseContextService.getLicenseSnapshot()?.valid ?? false;
+    const isSecurityEnabled = this.securityContextService.getSecurityConfig()?.isEnabled() ?? false;
+    const canWrite = this.authorizationService.canWriteRepo(repo);
+    const canCreateRepository = this.authorizationService.isRepoManager();
+    const accessibleRepositoriesCount = this.authorizationService.getAccessibleRepositories(true, ctx.isRestricted).size();
```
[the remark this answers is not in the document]
yordanalexandrov: This should be revisited in the other MR about the restrictions

packages/workbench/src/app/components/repository-picker-list/repository-picker-list.component.html
```diff
@@ -1 +1,107 @@
...
+      class="repository-list">
+      <ng-template #body let-repository>
+        <tr>
+          <td class="type-column">
+            <em [ngClass]="['type-icon', 'icon-lg', 'icon-repo-' + repository.type]"></em>
+          </td>
+          <td class="name-column">
+            <div class="repository-header">
+              <span class="repository-state"
+                    [pTooltip]="'components.repository_picker_list.state_filter.' + repository.stateLowercase | transloco"
```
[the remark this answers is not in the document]
boyan-tonchev: Fixed

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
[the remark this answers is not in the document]
boyan-tonchev: Done

packages/workbench/src/assets/i18n/fr.json
```diff
@@ -50,8 +50,37 @@
...
+    },
+    "repository_picker_list": {
+      "hide_remote_locations_btn": {
+        "label": "Masquer les emplacements distants"
+      },
+      "create_repository_btn": {
+        "label": "Créer un dépôt"
+      },
+      "select_repository_btn": {
+        "label": "Select"
```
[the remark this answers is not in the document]
boyan-tonchev: Done

packages/root-config/scripts/validate-translations.js
boyan-tonchev: Removed

packages/api/src/services/domain/security/authorization.service.ts
```diff
@@ -295,6 +298,46 @@ export class AuthorizationService implements Service {
...
+   * Retrieves a list of repositories that the user has access to, based on the specified parameters.
+   * @param includeRemote - If true, includes remote repositories in the list; otherwise, only local repositories are included.
+   * @param requireWriteAccess - If true, returns repositories that the user has write access to; if false, returns repositories that the user has read access to.
+   * @returns A list of repositories that the user has access to, filtered based on the provided parameters.
+   */
+  getAccessibleRepositories(includeRemote = false, requireWriteAccess = true): RepositoryList {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    let remoteLocationsFilter = (_repo: Repository) => true;
+    if (!includeRemote) {
+      remoteLocationsFilter = (repo) => !!repo.local;
```
plamen-yordanov, line 332: `local` should already be a boolean. Is the conversion needed?

packages/workbench/src/testing-utils/resize-observer-utils.ts
boyan-tonchev: Renamed

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.ts
```diff
@@ -0,0 +1,144 @@
...
+   */
+  requiredRepositoryPermission  = input<RepositoryPermissionType | undefined>(undefined);
+
+  private readonly subscriptions = new SubscriptionList();
+
+  readonly restrictions = computed(() =>
+    this.restrictionResolverService.resolve({
+      selectedRepository: this.selectedRepository(),
+      isRestricted: this.isRestricted(),
+      pageTitle: this.title() ?? 'Missing page title',
```
plamen-yordanov, line 84: the default should probably be translated

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
```diff
@@ -0,0 +1,107 @@
...
+import {
+  AuthorizationService,
+  LicenseContextService,
+  Repository,
+  SecurityContextService,
+  service,
+} from '@ontotext/workbench-api';
+import {RestrictionReason} from './restriction-reason';
+
+export interface RestrictionContext {
```
plamen-yordanov, line 13: this should be in models

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
```diff
@@ -0,0 +1,107 @@
...
+  selectedRepository: Repository | undefined;
```
plamen-yordanov, line 14: we should pick one approach for the optional stuff. We have both `?:` and `| undefined`

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.html
```diff
@@ -0,0 +1,29 @@
+<div class="page-restrictions" [style.height.px]="availableHeight()">
+  @if (restrictions().length) {
```
plamen-yordanov, line 2: I think this`if` is not needed. The `@for` will not do anything with an empty array

packages/workbench/src/testing-utils/resize-observer-utils.ts
plamen-yordanov, line 1: rename to `resize-observer-testing-utils` or something which suggests it is for testing.

packages/api/src/services/domain/security/authorization.service.ts
boyan-tonchev: The model was boolean or undefined. I changed the model to be only boolean, and change the default value to be set as it is in the backend.

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.ts
boyan-tonchev: I removed the default because I don't think we can provide a meaningful default name here. It is used as the description parameter in a case when the selected repository is FedX.

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
boyan-tonchev: Moved

packages/workbench/src/app/services/page-restrictions/restriction-resolver.service.ts
boyan-tonchev: I changed to use ?

packages/workbench/src/testing-utils/resize-observer-testing-utils.ts
boyan-tonchev: Renamed

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.html
boyan-tonchev: Removed

plamen-yordanov (dismissed):

yordanalexandrov (dismissed):

packages/api/src/services/domain/security/authorization.service.ts
boyan-tonchev: You are right. I added a comment and created an [issue](https://graphwise.atlassian.net/browse/GDB-15286) to address it.

yordanalexandrov (approved):

PR #3002, opened 2026-06-17
GDB-13829: Integrate interactive guides with new SPARQL view

yordanalexandrov:
## What

Integrate interactive guides with the new Angular SPARQL view. Introduces an event-based synchronisation protocol between `ShepherdService` and the active micro-frontend, enriches `GuideApi` with frontend-specific services, simplifies `GuidesService.getServices()` to return the `GuideApi` instance directly, and adds a dedicated Cypress configuration for guide e2e tests.

## Why

Guides run inside `@ontotext/workbench-api` but depend on frontend-specific services (notifications, translations, repository data, etc.) that live in different micro-frontends depending on the current page — either the Angular workbench or the legacy AngularJS workbench. Without synchronisation, `GuideApi` can hold stale or missing service references at the moment a guide step is displayed after navigation.

## How

A two-way event protocol is introduced:

- `ShepherdService` emits a `RequestServiceUpdateEvent` before every step (via `setTimeout` to let the current call stack settle).
- Both micro-frontends listen to this event and check whether they are the active frontend using `single-spa:before-app-change` lifecycle data.
- The active frontend injects its services into `GuideApi` and emits a `ServiceUpdatedEvent`.
- `ShepherdService` blocks the step from showing until `ServiceUpdatedEvent` is received.

`GuidesService.getServices()` in the AngularJS workbench is simplified to return the `GuideApi` instance directly instead of a manually assembled plain object, since `GuideApi` now carries all required services.

## Testing

- Start a guide in the legacy AngularJS view and navigate between steps across different URLs
- Start a guide in the new Angular SPARQL view
- Walk a guide that crosses both micro-frontends (e.g. a step in the legacy view followed by a step in the SPARQL view)
- `npm run cy:run-guides` for automated e2e tests

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/api/src/models/app-lifecycle/application-names.ts
```diff
@@ -0,0 +1,4 @@
+export enum ApplicationNames {
```
svilenvelikov, line 1: I think it's not a big issue in our case, but still, can you check if we can derive the app names from the single-spa instead of hardcodding them

Image

svilenvelikov: but now seeing how you are using this enum in the ApplicationsState I think this won't work

yordanalexandrov: Yes, I wanted to be able to reference them not by string

svilenvelikov (dismissed):

packages/workbench/src/app/app.component.ts
```diff
@@ -36,6 +41,28 @@ export class AppComponent implements OnInit, OnDestroy {
...
+
+  /**
+   * Part of the micro-frontend guide-sync protocol (see ShepherdService class JSDoc).
+   *
+   * Injects Angular-specific services into {@link GuideApi} so they are ready before
+   * the pending guide step is shown. Called only when this frontend is the one being loaded.
+   */
+  private updateGuideServices() {
+    const guidesApi = service(GuideApi);
+    guidesApi.YasguiComponentDirectiveUtil = YasguiComponentUtil;
```
boyan-tonchev: This is not enough to change the reference. We also need to update the plugins because they still reference the legacy Workbench. This happens because, when the guides are compiled during startup, the steps initialize a reference to the legacy utility (for example https://github.com/Ontotext-AD/graphdb-workbench-plugins/blob/main/plugins/guides/sparql-editor/execute-sparql-query.js#L61). We changed the reference here, but the plugin project still uses the old one.

I verified this by disabling the old plugin and enabling the new one. The guides stopped working because the plugin project was not updated and continued to rely on the legacy Workbench reference.

boyan-tonchev: fixed

boyan-tonchev (dismissed):

boyan-tonchev (approved):

PR #3004, opened 2026-06-17
GDB-14762: Fix query abort handling

boyan-tonchev:
## What
- When the Logout action was triggered while a query was executing, the application started the logout process immediately. This caused the header and sidebar to disappear before the user confirmed the action. Additionally, the running query was not aborted correctly.
- Change EventService emit signature to return Promise<boolean>.

## Why
- There was no mechanism to prevent event execution once an action was triggered. As a result, events such as Logout were emitted immediately, even when additional confirmation or asynchronous processing was required before executing the action.
- The single-spa navigation event exposes a cancelNavigation method that accepts either a boolean or a Promise<boolean>, depending on the resolved value, navigation is either canceled or allowed to proceed. Previously, emit was synchronous, so this worked correctly. After adding event cancellation support, emit became asynchronous internally (it now awaits each subscriber's cancellation check) — but its signature stayed void, so single-spa had no way to wait for that check. As a result, single-spa continued navigating immediately, and any cancellation request from a subscriber was effectively ignored.

## How
- Extended the EventService subscription mechanism with optional cancellation handlers.
Before an event is emitted, all registered cancellation handlers are evaluated. If any handler requests cancellation, the event is not emitted and the associated action is not executed.
 Implemented cancellation handlers for query-related actions in the SPARQL editor. These handlers display the running query confirmation dialog, abort the running query when the user confirms the action, and prevent the action when the user cancels it.
- emit now returns a Promise<boolean> indicating whether the event was canceled. This promise is passed directly to single-spa's cancelNavigation, so single-spa waits for the cancellation check to resolve before proceeding. If the event was canceled, navigation is prevented; otherwise, it continues.

## Additional work
Fixes Operations links to open views in a new tab.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [ ] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/models/events/auth/logged-out.ts
+++ b/packages/api/src/models/events/auth/logged-out.ts
@@ -1,0 +1,13 @@
+import {Event} from '../event';
+import {EventName} from '../event-name';
+
+/**
+ * Represents an {@link EventName.LOGGED_OUT} event.
+ *
+ * This event is triggered when user logged out.
+ */
+export class LoggedOut extends Event<undefined> {
+  constructor() {
+    super(EventName.LOGGED_OUT);
+  }
+}
--- a/packages/api/src/services/domain/guides/guide-api.ts
+++ b/packages/api/src/services/domain/guides/guide-api.ts
@@ -6,7 +6,7 @@ import {LoggerProvider} from '../../logging/logger-provider';
 import {GuideStepBridge} from './guide-step-bridge';
 import {WindowService} from '../../window';
 import {ScrollLocation} from '../../../models/interactive-guide/scroll-location';
-import {getCurrentRoute, HtmlUtil, navigate, UriUtil} from '../../utils';
+import {getCurrentRoute, HtmlUtil, navigate, TranslationUtil, UriUtil} from '../../utils';
 import {ProductInfoContextService} from '../product-info';
 
 /**
@@ -56,42 +56,64 @@ export class GuideApi implements Service, GuideStepBridge {
    * @returns The translated string, or the key itself if no translation is found.
    */
   _translate(bundle: TranslationBundle | undefined, key: string | Record<string, string>, parameters: Record<string, string> = {}): string {
-    const currentLanguage = this.languageContextService.getSelectedLanguage() ?? this.languageContextService.getDefaultLanguage();
+    const defaultLanguage = this.languageContextService.getDefaultLanguage();
+    const currentLanguage = this.languageContextService.getSelectedLanguage() ?? defaultLanguage;
 
     if (typeof key === 'object') {
-      const translation = key[currentLanguage] ?? key[this.languageContextService.getDefaultLanguage()];
-      if (translation) {
-        return HtmlUtil.sanitize(this.applyParameters(translation, parameters));
+      const translation = key[currentLanguage] ?? key[defaultLanguage];
+
+      if (!translation) {
+        this.logger.warn(
+          `Missing translation for language [${currentLanguage}] in message object`,
+        );
+
+        return '';
       }
-      this.logger.warn(`Missing translation for language [${currentLanguage}] in message object`);
-      return '';
+
+      return HtmlUtil.sanitize(
+        TranslationUtil.applyParameters(translation, parameters),
+      );
     }
 
-    let languageBundle: TranslationBundle | undefined;
-    if (bundle) {
-      const byLanguage = bundle[currentLanguage];
-      languageBundle = (typeof byLanguage === 'object' ? byLanguage : undefined) ?? bundle;
-    } else {
-      languageBundle = this.languageContextService.getLanguageBundle() ?? this.languageContextService.getDefaultBundle();
+    const serviceBundle =
+      this.languageContextService.getLanguageBundle() ??
+      this.languageContextService.getDefaultBundle();
+
+    const currentBundle = bundle
+      ? this.resolveLanguageBundle(bundle, currentLanguage)
+      : serviceBundle;
+
+    if (!currentBundle) {
+      this.logger.warn(
+        `Translation bundle is not provided for translation key: ${JSON.stringify(key)}`,
+      );
+
+      return key;
     }
 
-    if (!languageBundle) {
-      this.logger.warn('Translation bundle is not provided for translation key: ' + JSON.stringify(key));
-      return key?.toString() ?? '';
-    }
+    const translation =
+      TranslationUtil.translate(currentBundle, key, parameters) ??
+      (bundle
+        ? TranslationUtil.translate(serviceBundle, key, parameters)
+        : undefined);
 
-    let translation = languageBundle[key];
-    if (!translation && bundle) {
-      // Fallback to the language service bundle when the provided bundle does not contain the key
-      translation = (this.languageContextService.getLanguageBundle() ?? this.languageContextService.getDefaultBundle())?.[key];
-    }
-
-    if (translation) {
-      return HtmlUtil.sanitize(this.applyParameters(translation as string, parameters));
+    if (translation !== undefined) {
+      return translation;
     }
 
     this.logger.warn(`Missing translation for [${key}] key`);
-    return key?.toString();
+    return key;
+  }
+
+  private resolveLanguageBundle(
+    bundle: TranslationBundle,
+    language: string,
+  ): TranslationBundle {
+    const languageBundle = bundle[language];
+
+    return typeof languageBundle === 'object'
+      ? languageBundle
+      : bundle;
   }
 
   private _resolveDocumentationUrl(endpoint: string): string {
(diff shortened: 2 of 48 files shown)


packages/api/src/models/events/auth/logged-out.ts
```diff
@@ -0,0 +1,13 @@
+import {Event} from '../event';
+import {EventName} from '../event-name';
+
+/**
+ * Represents a "loggedOut" event.
```
yordanalexandrov: Should be: represents a `EventName.LOGGED_OUT` event

as this is the proper event name

packages/api/src/models/events/auth/logged-out.ts
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * This event is triggered when users logged out.
```
yordanalexandrov: `uesr is logged out`

packages/api/src/models/events/auth/logged-out.ts
boyan-tonchev: Done

packages/api/src/models/events/auth/logged-out.ts
boyan-tonchev: Fixed

packages/api/src/services/domain/guides/guide-api.ts
```diff
@@ -56,42 +56,64 @@ export class GuideApi implements Service, GuideStepBridge {
    * @returns The translated string, or the key itself if no translation is found.
    */
   _translate(bundle: TranslationBundle | undefined, key: string | Record<string, string>, parameters: Record<string, string> = {}): string {
-    const currentLanguage = this.languageContextService.getSelectedLanguage() ?? this.languageContextService.getDefaultLanguage();
+    const defaultLanguage = this.languageContextService.getDefaultLanguage();
+    const currentLanguage = this.languageContextService.getSelectedLanguage() ?? defaultLanguage;
```
svilenvelikov: can't we just make the getSelectedlanguage return the selected or the default instead of implementing such things througout entire app

packages/api/src/services/domain/guides/guide-api.ts
```diff
@@ -56,42 +56,64 @@ export class GuideApi implements Service, GuideStepBridge {
...
+      );
     }
 
-    if (!languageBundle) {
-      this.logger.warn('Translation bundle is not provided for translation key: ' + JSON.stringify(key));
-      return key?.toString() ?? '';
-    }
+    const serviceBundle =
+      this.languageContextService.getLanguageBundle() ??
+      this.languageContextService.getDefaultBundle();
```
svilenvelikov: Same here! Both methods are in the context service, so we can just make the getlanguageBundle return it or the default

packages/api/src/services/domain/guides/guide-api.ts
boyan-tonchev: Done

packages/api/src/services/domain/guides/guide-api.ts
boyan-tonchev: Done

packages/api/src/services/language/language-context.service.ts
```diff
@@ -94,6 +94,10 @@ export class LanguageContextService extends ContextService<LanguageContextFields
     return this.getContextPropertyValue(this.LANGUAGE_BUNDLE);
   }
 
+  getLanguageBundleOrDefault(): TranslationBundle | undefined {
```
svilenvelikov: OK. I don't want to be picky, but why do we need the additional method? This way, someone might still be confused to use the `getLanguageBundle` and invent its own default somewhere in other modules. We don't need that.

packages/legacy-workbench/src/js/angular/plugin.js
```diff
@@ -47,7 +47,7 @@ PluginRegistry.add('main.menu', {
     },
 );
 PluginRegistry.add('main.menu', {
-        disabled: true,
+        disabled: false,
```
svilenvelikov: revert this

packages/api/src/services/domain/guides/guide-api.ts
svilenvelikov: are you sure its done?

packages/api/src/services/language/language-context.service.ts
boyan-tonchev: Sorry, I misunderstood you, I will change it.

packages/legacy-workbench/src/js/angular/plugin.js
boyan-tonchev: Ups :)

packages/api/src/services/domain/guides/guide-api.ts
boyan-tonchev: Done

packages/api/src/services/language/language-context.service.ts
boyan-tonchev: Removed  the getLanguageBundleOrDefault and updated the getLanguageBundle and getSelectedLanguage methods

packages/legacy-workbench/src/js/angular/plugin.js
boyan-tonchev: Reverted

packages/api/src/services/event-service/event.service.ts
```diff
@@ -19,29 +21,63 @@ export class EventService implements Service {
...
+    }
+
+    return this.shouldCancelEvent(subscribers, event)
+      .then((cancelEvent) => {
+        if (!cancelEvent) {
+          subscribers.forEach((subscriber) => {
+            subscriber.notify(ObjectUtil.deepCopy(event.payload));
+          });
+        }
+        return Promise.resolve(cancelEvent);
```
yordanalexandrov: just return the value, no need to create a new promise

boyan-tonchev: Removed

packages/api/src/services/domain/guides/guide-api.ts
```diff
@@ -115,14 +133,11 @@ export class GuideApi implements Service, GuideStepBridge {
    * @returns The translation with parameters applied.
    */
   applyParameters(translation: string, parameters: Record<string, string> = {}): string {
```
yordanalexandrov, line 135: this looks obsolete now

packages/api/src/services/domain/security/authentication.service.ts
```diff
@@ -73,9 +73,9 @@ export class AuthenticationService implements Service {
         this.securityContextService.updateIsLoggedIn(false);
         if (this.authorizationService.hasFreeAccess()) {
           this.authorizationService.initializeFreeAccess();
-          this.eventService.emit(new Login());
+          void this.eventService.emit(new Login());
```
yordanalexandrov: As you are in a promise chain, you can return the emitting of the events

yordanalexandrov: You await it in the login

packages/api/src/services/event-service/event.service.ts
```diff
@@ -19,29 +21,63 @@ export class EventService implements Service {
...
+   *
+   * Handlers are evaluated sequentially until one of them requests cancellation of the event.
+   *
+   * @param observers - The observers whose cancellation handlers should be evaluated.
+   *
+   * @returns A promise that resolves to <code>true</code> if the event should be canceled,
+   * or <code>false</code> if no handler requests cancellation.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private async shouldCancelEvent<T extends {} | undefined>(observers: EventObserver<any>[], event: Event<T>): Promise<boolean> {
```
yordanalexandrov: You can probably replace the `any` with `unknown` and remove the eslint exception. Also `event` is missing from docs

packages/shared-components/src/components/onto-operations-notification/onto-operations-notification.tsx
```diff
@@ -74,7 +74,7 @@ export class OntoOperationsNotification {
                   <li key={operation.id}>
                     <a class={`operation-status-content onto-btn no-underline ${operationStatusToWarningClass[operation.status]}`}
                       target="_blank"
-                      onClick={navigateTo(operation.href)}>
+                      href={operation.href}>
```
yordanalexandrov, line 77: Is this ok with/without proxi?

yordanalexandrov: It seems to work

yordanalexandrov (changes requested):

yordanalexandrov (dismissed):

packages/api/src/services/domain/guides/guide-api.ts
boyan-tonchev: Yes, I agree with you. The same applies to `replaceAll`, but they are public methods, and I’m not sure if they are used anywhere, so I decided to leave them.

packages/api/src/services/domain/security/authentication.service.ts
boyan-tonchev: Done

packages/api/src/services/event-service/event.service.ts
boyan-tonchev: Done

packages/api/src/services/domain/guides/guide-api.ts
yordanalexandrov: We should remove them, after we can look for them in the codebase.

boyan-tonchev: I’ll check if we use them in the guides.

PR #3005, opened 2026-06-18
GDB-14755 Create reactodia page

plamen-yordanov:
## What
Create reactodia page

## Why
To implement reactodia as a custom web component inside the workbench

## How
- Added `reactodia-page.component` which will house the new custom web component wrapper.
- Added new endpoint which allows a dynamic `accept` header. This is needed by reactodia to make SPQRQL requests using the host http client, also different sparql requests may have different accept headers

## Testing
n/a

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/workbench/src/app/pages/reactodia/reactodia-page.component.html
+++ b/packages/workbench/src/app/pages/reactodia/reactodia-page.component.html
@@ -1,0 +1,10 @@
+<app-page-layout class="sparql-editor-page">
+  <div>
+    @if (language() && currentRepository()) {
+      <app-reactodia-component-facade
+        [currentRepository]="currentRepository()"
+        [language]="language()">
+      </app-reactodia-component-facade>
+    }
+  </div>
+</app-page-layout>
--- a/packages/workbench/src/app/pages/reactodia/reactodia-page.component.ts
+++ b/packages/workbench/src/app/pages/reactodia/reactodia-page.component.ts
@@ -1,0 +1,58 @@
+import {Component, OnDestroy, OnInit, signal} from '@angular/core';
+import {LanguageContextService, RepositoryContextService, service, SubscriptionList} from '@ontotext/workbench-api';
+import {
+  ReactodiaComponentFacadeComponent
+} from '../../components/reactodia-component-facade/reactodia-component-facade.component';
+import {PageLayoutComponent} from '../../components/page-layout/page-layout.component';
+
+/**
+ * Page that hosts the Reactodia graph. It owns the context subscriptions (repository and language),
+ * gates between the "repository required" banner and the {@link ReactodiaComponentFacadeComponent}
+ * (which owns the `graphwise-reactodia` web component and its wiring) based on the active
+ * repository, and feeds the current repository/language down to the facade.
+ */
+@Component({
+  selector: 'app-reactodia-page',
+  standalone: true,
+  templateUrl: './reactodia-page.component.html',
+  imports: [
+    ReactodiaComponentFacadeComponent,
+    PageLayoutComponent,
+  ],
+  styleUrl: './reactodia-page.component.scss'
+})
+export class ReactodiaPageComponent implements OnInit, OnDestroy {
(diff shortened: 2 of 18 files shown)


packages/workbench/src/app/pages/reactodia/reactodia-page.component.html
```diff
@@ -0,0 +1,9 @@
+@if (!currentRepository()) {
+  <app-repository-required-banner></app-repository-required-banner>
+} @else if(language()) {
+  <graphwise-reactodia
```
svilenvelikov: I'd strongly suggest having a facade in front of the graphwise-reactodia web component that will contain the initialization and configuration logic, event handlers and what not. This woud be similar thos we have for the yasgui, graphql-playground, etc. See these for example:
 https://github.com/Ontotext-AD/graphdb-workbench/blob/master/packages/workbench/src/app/components/yasgui-component-facade/yasgui-component-facade.component.ts
https://github.com/Ontotext-AD/graphdb-workbench/blob/master/packages/legacy-workbench/src/js/angular/core/directives/yasgui-component/yasgui-component.directive.js
https://github.com/Ontotext-AD/graphdb-workbench/blob/master/packages/legacy-workbench/src/js/angular/core/directives/graphql-playground/graphql-playground.directive.js

packages/workbench/src/app/pages/reactodia/reactodia-page.component.ts
```diff
@@ -0,0 +1,93 @@
...
+  readonly language = signal(this.languageContextService.getSelectedLanguage());
+
+  /**
+   * Transport for Reactodia's SPARQL requests. Reactodia chooses the `Accept` per query
+   * (SPARQL-results JSON for SELECT lookups, RDF/Turtle for CONSTRUCT element info) and passes
+   * it in `params.headers`; we forward it so GraphDB returns the matching format, then hand back
+   * the raw `Response` which Reactodia expects.
+   */
+  readonly queryFunction = (params: SparqlQueryParams) =>
+    this.rdf4jRestService
```
boyan-tonchev: I think we should avoid injecting the REST service directly into the component. It would be better to introduce a service that encapsulates the REST service calls.

svilenvelikov: I agree. This should not be done this way. There should be a callback function here that is passed as a query function to the component and the callback itself should be using the dedicated domain service and not the rest controller itself as Boyan pointed out.

packages/workbench/src/app/pages/reactodia/reactodia-page.component.html
plamen-yordanov: added a facade and left the transport function there. The language and repository subscriptions are in the page component and passed as input parameters

packages/workbench/src/app/pages/reactodia/reactodia-page.component.ts
plamen-yordanov: Added a new method in the `rdf4jRepositoryService` and used that instead of the rest service directly

packages/workbench/src/app/pages/reactodia/reactodia-page.component.html
```diff
@@ -0,0 +1,8 @@
+@if (!currentRepository()) {
+  <app-repository-required-banner></app-repository-required-banner>
+} @else {
+  <app-reactodia-component-facade
+    [currentRepository]="currentRepository()"
+    [language]="language()">
+  </app-reactodia-component-facade>
```
svilenvelikov: Wrap the whole thing in the page-layout please. Look at the sparql view page.

Image

plamen-yordanov: wrapped

svilenvelikov (dismissed):

PR #3006, opened 2026-06-18
GDB-14177 update `concurrently` to version 10.0.3 and other dependencies

plamen-yordanov:
## What
Update `concurrently` to version 10.0.3 and other dependencies in `package.json` and `package-lock.json`.

## Why
To fix vulnerabilities

## How
- Updated `concurrently` from version `9.2.1` to `10.0.3`.
- Updated various other dependencies in `package-lock.json` to their latest versions, including `js-yaml`, `body-parser`, `express`, and `yargs`.

## Testing
n/a

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/workbench/package.json
+++ b/packages/workbench/package.json
@@ -50,6 +50,7 @@
     "eslint": "^10.4.1",
     "jest": "^30.4.2",
     "jest-environment-jsdom": "^30.4.1",
+    "jest-preset-angular": "^16.2.0",
     "jsdom": "^29.1.1",
     "sonarqube-scanner": "^3.3.0",
     "typescript": "^5.9.3",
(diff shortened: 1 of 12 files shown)


packages/workbench/package.json
```diff
@@ -50,6 +50,7 @@
     "eslint": "^10.4.1",
     "jest": "^30.4.2",
     "jest-environment-jsdom": "^30.4.1",
+    "jest-preset-angular": "^16.2.0",
```
yordanalexandrov, line 53: Sholdn't this be configured to be used? Is just adding it as a dependency working?

plamen-yordanov: after `npm audit fix` it ended up in `node_modules/@angular-builders/jest/node_modules/jest-preset-angular` for some reason. Adding it as a dependency works, yes

yordanalexandrov (approved):

PR #3012, opened 2026-06-25
GDB-14779: Solr deprecation banner is not sticky

boyan-tonchev:
## What
- Kept the deprecation banner sticky together with the application header.
- Added user preferences infrastructure for storing user-specific UI settings.
- Implemented persistence of the Solr deprecation banner dismissed state for authenticated users.

## Why
- The deprecation banner was rendered outside of the sticky header container, causing it to scroll out of view instead of remaining visible with the header.
- Users should not have to dismiss the Solr deprecation banner every time they visit the application when they are authenticated.
- Anonymous users cannot be identified, so their dismissed state should be preserved only for the current browser session.

## How
- Moved the deprecation banner into the sticky header container so both the header and the banner remain fixed at the top while scrolling.
- Introduced models, context, storage, and service classes for managing user-specific preferences.
- Persisted the dismissed banner state per authenticated user while keeping the preference in memory for anonymous users.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


packages/api/src/services/domain/users/user-preferences.service.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Checks whether preferences should be persisted for the current user.
+   *
+   * Preferences are persisted only when security is enabled and the user is logged in.
+   *
+   * @returns {@code true} if preferences should be persisted.
+   */
+  private shouldPersistPreferences(): boolean {
+    return this.securityContextService.getSecurityConfig()?.enabled === true
```
plamen-yordanov: you can use `config.isEnabled()`

plamen-yordanov: Without security enabled the user is an admin (or a repo manager if there is override auth). Shouldn't those be able to persist their preferences? If a user should be explicitly logged in, you can replace all of this with `authentication.service#isLoggedIn`

boyan-tonchev: I will check this. If security is disabled and getIsLoggedIn() returns false, I'll change it.

packages/api/src/services/domain/users/user-context.service.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+type UserContextFieldParams = {
+  readonly USER_PREFERENCES: UserPreferences;
+};
+
+/**
+ * Context service responsible for storing user-specific data for the current application session.
+ */
+export class UserContextService extends ContextService<UserContextFields> implements DeriveContextServiceContract<UserContextFields, UserContextFieldParams>, LifecycleHooks {
```
svilenvelikov: This must be UserPreferencesContextService

packages/api/src/services/domain/users/user-preferences.service.ts
svilenvelikov: I think we can go with logged in users only.

packages/api/src/services/user-preferences/user-preferences.service.ts
svilenvelikov, line 1: Why are these services in the domain package? They have nothing to do with the domains. Move them outside

packages/api/src/services/domain/users/user-preferences.service.ts
```diff
@@ -0,0 +1,121 @@
...
+  private loadCurrentUserPreferences(): UserPreferences {
+    const username = this.getAuthenticatedUsername();
+
+    if (!username) {
+      return new UserPreferences();
+    }
+
+    return this.userPreferencesStorageService
+      .getUsersPreferences()
+      .getUserPreferences(username) ?? new UserPreferences();
```
svilenvelikov: This should be something like this
```
this.userPreferencesStorageService
      .getPreferences()
      .getPreferenceByUsername(username)
```

packages/api/src/services/domain/users/user-context.service.ts
boyan-tonchev: Initially, the name was UserPreferencesContextService, but I thought it could be used in the future for other user-related properties.

packages/api/src/services/user-preferences/user-preferences.service.ts
boyan-tonchev: I thought of them as user-related, so I placed them next to the user service. I'll rename UserPreferencesContextService and move them back to where they were.

packages/api/src/services/domain/users/user-preferences.service.ts
boyan-tonchev: Done

boyan-tonchev: Done

packages/api/src/services/domain/users/user-context.service.ts
boyan-tonchev: Renamed and moved

packages/api/src/services/user-preferences/user-preferences.service.ts
boyan-tonchev: Moved

packages/api/src/services/domain/users/user-preferences.service.ts
boyan-tonchev: Done

packages/api/src/services/user-preferences/test/user-preferences.service.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+};
+
+describe('UserPreferencesService', () => {
+  let userPreferencesService: UserPreferencesService;
+  let userStorageService: UserStorageService;
+  let userPreferencesContextService: UserPreferencesContextService;
+  let securityContextService: jest.Mocked<SecurityContextService>;
+  let authenticationService: jest.Mocked<AuthenticationService>;
+
+  let storage: Record<string, string>;
```
svilenvelikov: We have a StorageMock. Why don't you use it?

packages/api/src/services/user-preferences/test/user-preferences.service.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+  let context: Record<string, unknown>;
```
svilenvelikov: Why do you need to mock the context?! Why not just using the actual context service and just spy on it if you need.

packages/api/src/services/user-preferences/user-preferences.service.ts
```diff
@@ -0,0 +1,113 @@
...
+  private loadCurrentUserPreferences(): UserPreferences | undefined {
+    const username = this.getAuthenticatedUsername();
+
+    if (!username) {
+      return new UserPreferences();
+    }
+
+    return this.userPreferencesStorageService
+      .getUsersPreferences()
+      .getUserPreferences(username);
```
svilenvelikov, line 75: this is still unchanged

packages/api/src/models/user-preference/user-preferences.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Stores user-specific UI preferences.
+ *
+ * These preferences are persisted for authenticated users and control personalized UI behavior.
+ */
+export class UserPreferences {
+  /**
+   * Indicates whether the user has dismissed the Solr deprecation banner.
+   * When {@code true}, the banner is not shown again for the user.
+   */
+  dismissSolrDeprecationBanner = false;
```
svilenvelikov: rename this to either `dismissedSolrDeprecationBanner` or `isSolrDeprecationBannerDismissed`

boyan-tonchev: Done

packages/api/src/services/user-preferences/user-preferences.service.ts
boyan-tonchev: Done

packages/api/src/services/user-preferences/test/user-preferences.service.spec.ts
boyan-tonchev: Reimplemented

packages/api/src/services/user-preferences/test/user-preferences.service.spec.ts
boyan-tonchev: Reimplemented

svilenvelikov (approved):

PR #3020, opened 2026-06-29
GDB-14669: Show repository description on hover

boyan-tonchev:
## What
Add the repository description to the repository selector tooltip.

## Why
Some Workbench clients use generated repository IDs and rely on repository descriptions to identify them. Since the repository selector displays only the repository ID, it is difficult to distinguish between repositories.

## How
Extended the repository selector tooltip to display the repository description alongside the repository ID.

## Additional work
Fixes issues in the repository selector when repository names are too long.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/shared-components/src/components/onto-repository-selector/onto-repository-selector.scss
+++ b/packages/shared-components/src/components/onto-repository-selector/onto-repository-selector.scss
@@ -3,6 +3,16 @@
 }
 
 .onto-repository-selector {
+  .repository-selection {
+    max-width: 16rem;
+    display: flex;
+
+    .repository-id {
+      overflow: hidden;
+      text-overflow: ellipsis;
+    }
+  }
+
   .onto-dropdown {
     &.closed {
       .onto-dropdown-button:not(:hover) {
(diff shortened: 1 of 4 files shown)


packages/shared-components/src/components/onto-repository-selector/onto-repository-selector.scss
```diff
@@ -3,6 +3,16 @@
 }
 
 .onto-repository-selector {
+  .repository-selection {
+    max-width: 250px;
```
yordanalexandrov: Perhaps convert those to REMs

yordanalexandrov (dismissed):

packages/shared-components/src/components/onto-repository-selector/onto-repository-selector.scss
boyan-tonchev: Done

yordanalexandrov (approved):

PR #3025, opened 2026-07-01
GDB-14784 visualize resource with Reactodia

plamen-yordanov:
## What
Visualize resource with Reactodia.

## Why
As part of the ongoing integration of reactodia in the workbench

## How
- Added a new button for "Visualize Reactodia" in `explore.html` which redirects to the reactodia page with a `uri` parameter.
- Added a new `Reactodia` button to the rdf resource search, which also opens reactodia with the given resource in a new tab (like graph visualization)
- Updated `reactodia-page.component.ts` to handle query parameters for starting resources.

## Testing
cypress

## Screenshots
Screenshot from 2026-07-01 12-01-25
Screenshot from 2026-07-01 12-01-47
Screenshot from 2026-07-01 12-02-15

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/legacy-workbench/src/js/angular/explore/controllers.js
```diff
@@ -167,6 +167,12 @@ function ExploreCtrl(
             .search('config', payload.graphConfig?.id);
     };
 
+    $scope.goToReactodia = () => {
+        $location
+            .path('reactodia')
```
svilenvelikov, line 172: Didn't we have these paths as constants somewhere.

PR #3028, opened 2026-07-02
GDB-14792: Introduce manage repository rights in user view

yordanalexandrov:
## What
Added manage repository permissions to the user view, enhancing the user interface to include configuration of management authorities for repositories.

## Why
This change allows users to have an authority to manage  specific repositories

## How
- Added new columns and checkboxes for managing repository permissions in the user roles table.
- Introduced constants and authority checks for manage repository permissions in the backend logic.
- Updated localization files to include new labels and tooltips for manage repository rights.
- Amanded logic to convert to/from UI/GDB models of the granted authorities.

## Also in PR
- Configured `ng-tags-input` to add a watcher for placeholder label change.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/api/src/models/security/authorization/authority-list.ts
```diff
@@ -85,17 +85,20 @@ export class AuthorityList extends ModelList<string> {
...
     const auths = uiModel.getItems();
     auths.forEach(role => {
-      if (role.startsWith(Authority.READ_REPO_PREFIX) || role.startsWith(Authority.WRITE_REPO_PREFIX) || role.startsWith(Authority.GRAPHQL_PREFIX)) {
+      if (role.startsWith(Authority.READ_REPO_PREFIX) || role.startsWith(Authority.WRITE_REPO_PREFIX) || role.startsWith(Authority.MANAGE_REPO_PREFIX) || role.startsWith(Authority.GRAPHQL_PREFIX)) {
         const repoData = AuthoritiesUtil.getRepoFromAuthority(role);
         if (repoData) {
           const {prefix, repo} = repoData;
-          if (prefix === Authority.READ_REPO_PREFIX) {
+          if (prefix === Authority.MANAGE_REPO_PREFIX) {
+            grantedAuthorities[Authority.MANAGE_REPO][repo] = true;
```
plamen-yordanov, line 100: can't we replace all the if/else here with simply `grantedAuthorities[prefix][repo] = true`?

packages/legacy-workbench/src/js/angular/security/templates/user.html
```diff
@@ -256,7 +257,13 @@ <h3>{{'security.repo.rights' | translate}}</h3>
                                             </span>
                                         </th>
 
-                                        <th class="text-xs-center">
+                                        <th class="text-xs-center manage-repository-any">
+                                            <span>
+                                                <input class="manage-repository" type="checkbox" ng-checked="hasManageRepositoryPermission('*')" ng-disabled="true">
```
plamen-yordanov, line 262: always disabled?

packages/legacy-workbench/src/js/angular/security/controllers.js
```diff
@@ -189,7 +189,11 @@ securityModule.controller('UsersCtrl', ['$scope', '$uibModal', 'toastr', '$windo
                                 });
                                 _.each(authorities, function(a) {
                                     // indexOf works in IE 11, startsWith doesn't
-                                    if (a.indexOf(WRITE_REPO_PREFIX) === 0) {
+                                    if (a.indexOf(MANAGE_REPO_PREFIX) === 0) {
+                                        if (repoIds.hasOwnProperty(a.substr(11))) {
```
plamen-yordanov: shouldn't this be substr(12)? Also if we don't support IE anymore, maybe we can use `startsWith` and actual role lengh instead of magic numbers

packages/api/src/models/security/authorization/authority-list.ts
yordanalexandrov: `MANAGE_REPO` and `MANAGE_REPO_PREFIX` are different constants.

```
grantedAuthorities[Authority.MANAGE_REPO][repo]
is not
grantedAuthorities[Authority.MANAGE_REPO_PREFIX][repo]
```

packages/legacy-workbench/src/js/angular/security/templates/user.html
yordanalexandrov: yes. you can not select for a user to manage all repositories. This is a REPO MANAGER role.

You can manually select all repositories, but it is still not the same, since the user will not get access to new repositories.

packages/legacy-workbench/src/js/angular/security/controllers.js
yordanalexandrov: Probably yes

yordanalexandrov: But I suspect this has to be removed as this is for the free access user rights and I doubt it should be able to manage repositories.

packages/legacy-workbench/src/js/angular/security/controllers.js
```diff
@@ -272,6 +272,14 @@ securityModule.controller('DefaultAuthoritiesCtrl', ['$scope', '$http', '$uibMod
         $scope.ok = function() {
             const auth = [];
             $scope.repositoryCheckError = true;
+            for (const index in $scope.grantedAuthorities.MANAGE_REPO) {
```
boyan-tonchev: I think this isn't necessary because it's the OK button handler for the free access dialog, and there is no repository management there.

packages/legacy-workbench/src/js/angular/security/controllers.js
```diff
@@ -412,12 +447,19 @@ securityModule.controller('CommonUserCtrl', ['$rootScope', '$scope', '$http', 't
         };
 
         $scope.writeCheckDisabled = function(repoOrWildCard) {
-            return $scope.userType === UserType.ADMIN
+            return $scope.hasManageRepositoryPermission(repoOrWildCard)
+                || $scope.userType === UserType.ADMIN
                 || $scope.userType === UserType.REPO_MANAGER
```
boyan-tonchev: The `$scope.userType === UserType.ADMIN` and `$scope.userType === UserType.REPO_MANAGER` can be removed because they are part of `hasManageRepositoryPermission`

packages/legacy-workbench/src/js/angular/security/controllers.js
```diff
@@ -430,15 +472,24 @@ securityModule.controller('CommonUserCtrl', ['$rootScope', '$scope', '$http', 't
                 return true;
             }
 
+            if ($scope.hasManageRepositoryPermission(repoOrWildCard)) {
```
boyan-tonchev, line 461: Previous chek if `($scope.userType === UserType.ADMIN || $scope.userType === UserType.REPO_MANAGER)` can be removed, because because they are part of  `hasManageRepositoryPermission`

yordanalexandrov: removed

packages/legacy-workbench/src/js/angular/security/controllers.js
yordanalexandrov: removed

packages/legacy-workbench/src/js/angular/security/controllers.js
yordanalexandrov: removed. I have asked if it is needed for the free access to be configured with manage repo rights and if there is a positive answer I will add another issue.

boyan-tonchev (approved):

plamen-yordanov (approved):

PR #3032, opened 2026-07-03
GDB-14793: Allow Access to Repository Views for Users with MANAGE_REPO Permissions

boyan-tonchev:
## What
Allow access to repository views for users with MANAGE_REPO permissions.

## Why
We want to introduce a new type of user with extended repository permissions. An administrator can grant repository management permissions for specific repositories. A user with these permissions can manage only the repositories for which they have been granted access. However, they are not allowed to create or delete repositories, regardless of their repository-specific management permissions.

## How
- Introduced a new UI authority, ROLE_MANAGE_REPO. Users with this authority can manage only the repositories for which they have been granted management permissions. This authority does not allow creating or deleting repositories.
- ROLE_MANAGE_REPO does not exist in the backend. When the authenticated user is loaded, the UI derives and adds this authority if the user has the ROLE_REPO_MANAGER role or repository management permissions for at least one repository.
- Updated the repository view plugin configuration to use ROLE_MANAGE_REPO, making the repository views accessible to both repository managers and users with repository-specific management permissions.
- Updated the repository catalog view to apply permissions based on the user's authority:
  - Users with ROLE_REPO_MANAGER can view all repositories and perform all repository management actions.
  - Users with ROLE_MANAGE_REPO can view and manage only the repositories for which they have management permissions. They cannot create or delete repositories.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/services/domain/security/mappers/authenticated-user.mapper.ts
+++ b/packages/api/src/services/domain/security/mappers/authenticated-user.mapper.ts
@@ -1,8 +1,8 @@
-import {AuthenticatedUser} from '../../../../models/security/authenticated-user';
-import {AuthenticatedUserResponse} from '../../../../models/security/response-models/authenticated-user-response';
-import {AuthorityList} from '../../../../models/security';
-import {AppSettings} from '../../../../models/users/app-settings';
-import {MapperFn} from '../../../../providers/mapper/mapper-fn';
+import {AuthenticatedUser} from '../../../../models/security';
+import {AuthenticatedUserResponse} from '../../../../models/security';
+import {Authority, AuthorityList} from '../../../../models/security';
+import {AppSettings} from '../../../../models/users';
+import {MapperFn} from '../../../../providers';
 
 /**
  * Mapper class for converting partial AuthenticatedUser objects to complete AuthenticatedUser models.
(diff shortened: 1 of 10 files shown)


packages/api/src/services/domain/security/mappers/authenticated-user.mapper.ts
```diff
@@ -11,9 +11,32 @@ export const mapAuthenticatedUserResponseToModel: MapperFn<AuthenticatedUserResp
   return new AuthenticatedUser({
     username: data.username,
     password: data.password,
-    authorities: new AuthorityList(data.authorities),
+    authorities: new AuthorityList(normalizeAuthorities(data.authorities || [])),
```
yordanalexandrov: Remove `|| []`

yordanalexandrov (approved):

packages/api/src/services/domain/security/mappers/authenticated-user.mapper.ts
boyan-tonchev: Done

PR #3035, opened 2026-07-06
Fix loader not hiding on slow networks

yordanalexandrov:
## What
On slow network the initial booting loader is shown and never disapears

## Why
The SingleSPA events and the component loading order are not guaranteed

## How
- put the clearing of the loading flag inside a timeout

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


yordanalexandrov:
Decided to revert the change that broke it

PR #3049, opened 2026-07-08
GDB-14785 visualize construct query

plamen-yordanov:
## What
Visualize construct queries with Reactodia.

## Why
This feature allows users to visualize the results of SPARQL `CONSTRUCT` queries directly in the Reactodia graph explorer, just like visual graph

## How
- Added a new `exploreReactodiaYasrToolbarElementBuilder` to create a button for navigating to the Reactodia visualization.
- Implemented the `navigateToReactodia` method to handle navigation with query parameters.
- Updated the `config.yasrToolbarPlugins` to include the new toolbar element.
- Created a new `GraphExploreService` to compute graphs based on SPARQL queries.
- Introduced `GraphExploreLink` and related models to manage graph data.
- Updated the `ReactodiaPageComponent` to initialize the graph based on query parameters.
- Added relevant services

## Testing
jest/cypress

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/services/domain/graph-explore/graph-explore-context.service.ts
+++ b/packages/api/src/services/domain/graph-explore/graph-explore-context.service.ts
@@ -1,0 +1,49 @@
+import {ContextService} from '../../context';
+import {DeriveContextServiceContract} from '../../../models/context/update-context-method';
+import {ValueChangeCallback} from '../../../models/context/value-change-callback';
+import {GraphExploreSettings} from '../../../models/graph-explore/graph-explore-settings';
+import {DEFAULT_GRAPH_EXPLORE_SETTINGS} from './graph-explore-settings.config';
+
+type GraphExploreContextFields = {
+  readonly SETTINGS: string;
+};
+
+type GraphExploreContextFieldParams = {
+  readonly SETTINGS: GraphExploreSettings;
+};
+
+/**
+ * Service for managing the graph-explore settings context.
+ */
+export class GraphExploreContextService extends ContextService<GraphExploreContextFields> implements DeriveContextServiceContract<GraphExploreContextFields, GraphExploreContextFieldParams> {
+  readonly SETTINGS = 'graphExploreSettings';
+
+  /**
+   * Updates the default graph-explore settings in the context.
+   *
+   * @param settings - The new default settings object to be set in the context.
+   */
+  updateSettings(settings: GraphExploreSettings): void {
+    this.updateContextProperty(this.SETTINGS, settings);
+  }
+
+  /**
+   * Subscribes to changes in the default graph-explore settings context.
+   *
+   * @param callbackFn - A callback function that will be called when the default settings change.
+   * @returns A function that, when called, will unsubscribe from the default settings changes.
+   */
+  onSettingsChanged(callbackFn: ValueChangeCallback<GraphExploreSettings | undefined>): () => void {
+    return this.subscribe(this.SETTINGS, callbackFn);
+  }
+
+  /**
+   * Retrieves the default graph-explore settings stored in the context. When the context holds no
+   * value yet, falls back to the built-in {@link DEFAULT_GRAPH_EXPLORE_SETTINGS} configuration.
+   *
+   * @return the default graph-explore settings.
+   */
+  getSettings(): GraphExploreSettings {
+    return this.getContextPropertyValue<GraphExploreSettings>(this.SETTINGS) ?? DEFAULT_GRAPH_EXPLORE_SETTINGS;
+  }
+}
--- a/packages/api/src/services/domain/graph-explore/graph-explore-rest.service.ts
+++ b/packages/api/src/services/domain/graph-explore/graph-explore-rest.service.ts
@@ -1,0 +1,23 @@
+import {HttpService} from '../../http/http.service';
+import {GraphExploreRequest} from './request/graph-explore-request';
+import {GraphExploreLinkResponse} from './response/graph-explore-response';
+
+/**
+ * Service for interacting with the graph-explore REST API.
+ */
+export class GraphExploreRestService extends HttpService {
+
+  static readonly EXPLORE_GRAPH_ENDPOINT = 'rest/explore-graph';
+
+  /**
+   * Computes the graph for the given SPARQL query.
+   *
+   * @param request - The graph-explore request payload.
+   * @returns A Promise that resolves to a {@link GraphExploreLinkResponse[]} containing the graph links.
+   */
+  loadGraph(request: GraphExploreRequest): Promise<GraphExploreLinkResponse[]> {
+    return this.post(`${GraphExploreRestService.EXPLORE_GRAPH_ENDPOINT}/graph`, {
+      body: request
+    });
+  }
+}
--- a/packages/api/src/services/utils/routing-utils.ts
+++ b/packages/api/src/services/utils/routing-utils.ts
@@ -1,4 +1,5 @@
 import {WindowService} from '../window';
+import {ObjectUtil} from './object-util';
 
 /**
  * Redirects the current page to a specified URL using the single-spa framework.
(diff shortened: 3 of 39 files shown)


packages/api/src/services/domain/graph-explore/graph-explore-rest.service.ts
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * Computes the graph for the given SPARQL query.
+   *
+   * @param request - The graph-explore request payload.
+   * @returns A Promise that resolves to a {@link GraphExploreLinkResponse[]} containing the graph links.
+   */
+  loadGraph(request: GraphExploreRequest): Promise<GraphExploreLinkResponse[]> {
+    return this.post(`${GraphExploreRestService.EXPLORE_GRAPH_ENDPOINT}/graph`, {
+      headers: {
+        'Content-Type': 'application/json'
```
yordanalexandrov: You can omit this as this is the default content type set for each request

packages/api/src/services/domain/graph-explore/graph-explore-context.service.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  /**
+   * Retrieves the default graph-explore settings from the context, falling back to the
+   * {@link DEFAULT_GRAPH_EXPLORE_SETTINGS} configuration when no value has been set.
+   *
+   * @return the default graph-explore settings.
+   */
+  getDefaultSettings(): GraphExploreSettings {
+    return this.getContextPropertyValue<GraphExploreSettings>(this.DEFAULT_SETTINGS) ?? DEFAULT_GRAPH_EXPLORE_SETTINGS;
```
yordanalexandrov: This is a bit odd regarding naming.

You get the `default` settings, but if there are no `default` settings, you will get the `default` settings.

It does not make sense.

packages/legacy-workbench/src/js/angular/sparql-editor/controllers.js
```diff
@@ -379,6 +379,31 @@ function SparqlEditorCtrl($rootScope,
...
+            inference: yasr.yasqe.getInfer(),
+        };
+
+        const navigate = function() {
+            $location.path('reactodia').search(paramsToParse);
+        };
+        // This button is embedded in the Yasr toolbar, which is outside the Angular context, so we need to
+        // manually trigger the navigation inside the Angular context. See the note in
+        // "navigateToGraphsVisualizations" for more details.
+        if ($scope.$$phase || $rootScope.$$phase) {
```
yordanalexandrov: Won't it work if we use the singleSpa navigate?

I thought it triggers digest.

yordanalexandrov: You probably copied from the other, but it also should work using `navigate` from workbench-api

packages/api/src/services/domain/graph-explore/graph-explore-rest.service.ts
plamen-yordanov: removed

packages/api/src/services/domain/graph-explore/graph-explore-context.service.ts
plamen-yordanov: renamed to `getSettings`

packages/legacy-workbench/src/js/angular/sparql-editor/controllers.js
plamen-yordanov: Replaced with `navigate` and added functionality to add query params, when using it

packages/api/src/services/utils/routing-utils.ts
```diff
@@ -19,9 +20,11 @@ export function navigateTo(url: string): (event?: Event) => void {
  * Navigates to the specified URL using the single-spa framework.
  * Suitable for in-code navigation. If you need to navigate from a view, use <code>navigateTo</code> instead.
  * @param targetUrl - The target URL to which the page should be redirected.
+ * @param queryParams - Optional query parameters to append to the URL. Nullish values are skipped.
  */
-export function navigate(targetUrl: string) {
-  let url = targetUrl;
+export function navigate(targetUrl: string, queryParams?: Record<string, unknown>) {
```
yordanalexandrov, line 25: I am wondering about having a `preserveQueryParams` argument like in angular.

If true we append the query parameters and if false, we replace them.

Maybe in some next iteration of the function.

packages/api/src/services/utils/routing-utils.ts
```diff
@@ -19,9 +20,11 @@ export function navigateTo(url: string): (event?: Event) => void {
...
+  let url = addQueryParams(targetUrl, queryParams);
+
   if (url.startsWith('.')) {
```
yordanalexandrov: Since `addQueryParams` returns the full url we dont need this logic.

We can use `  WindowService.navigateSingleSpa(url);` directly

packages/api/src/services/utils/routing-utils.ts
```diff
@@ -31,6 +34,32 @@ export function navigate(targetUrl: string) {
...
+export function addQueryParams(url: string, queryParams: Record<string, unknown> = {}): string {
+  const entries = Object.entries(queryParams)
+    .filter(([, value]) => !ObjectUtil.isNullOrUndefined(value));
+
+  if (!entries.length) {
+    return url;
+  }
+
+  const parsedUrl = new URL(url, getOrigin());
+  if (!parsedUrl) {
```
yordanalexandrov: Interesting check, when is this supposed to happen?

packages/api/src/services/utils/routing-utils.ts
```diff
@@ -31,6 +34,32 @@ export function navigate(targetUrl: string) {
...
+ * @param url - The base URL to append the query parameters to.
+ * @param queryParams - The query parameters to append.
+ * @returns The URL with the query parameters appended, as an absolute URL.
+ */
+export function addQueryParams(url: string, queryParams: Record<string, unknown> = {}): string {
+  const entries = Object.entries(queryParams)
+    .filter(([, value]) => !ObjectUtil.isNullOrUndefined(value));
+
+  if (!entries.length) {
+    return url;
```
yordanalexandrov: perhaps move the `const parsedUrl = new URL(url, getOrigin());` in the beginning of the function and return `parsedUrl.toString()`

packages/api/src/services/utils/routing-utils.ts
```diff
@@ -31,6 +34,32 @@ export function navigate(targetUrl: string) {
...
+/**
+ * Appends query parameters to a URL. Nullish values are skipped and remaining values are stringified.
+ * Existing query parameters are preserved and merged (a duplicate key is overwritten with the new value),
+ * and any URL fragment (<code>#hash</code>) is kept at the end where it belongs.
+ *
+ * @param url - The base URL to append the query parameters to.
+ * @param queryParams - The query parameters to append.
+ * @returns The URL with the query parameters appended, as an absolute URL.
+ */
+export function addQueryParams(url: string, queryParams: Record<string, unknown> = {}): string {
```
yordanalexandrov: There is not point in having a default value for a readonly object here.

Just check if it is defined and if not, return `parsedUrl.toString()`

packages/api/src/services/utils/routing-utils.ts
```diff
@@ -31,6 +34,32 @@ export function navigate(targetUrl: string) {
...
+  const entries = Object.entries(queryParams)
```
yordanalexandrov: You can skip this whole loop and do a check in the `forEach` when you append them

packages/api/src/services/utils/routing-utils.ts
plamen-yordanov: Who will handle the leading dot? If we are going to always use an absolute URLs we can remove the next `if` statement which checks for absolute and base paths, but I think this one needs to stay. We need to move `addQueryParams` after the dot removal IMO

packages/api/src/services/utils/routing-utils.ts
plamen-yordanov: Initially this was `URL.parse`, which didn't exist in CI, so I replaced it with `new URL`. Looks like the check is no longer needed

packages/api/src/services/utils/routing-utils.ts
plamen-yordanov: Removed the relative parameter check and kept the absolute, since the context + absolute path breaks otherwise. Also corrected the wrong test

packages/api/src/services/utils/routing-utils.ts
plamen-yordanov: Removed this check altogether. If there are query params, we add them and always return `parsedUrl.toString`

packages/api/src/services/utils/routing-utils.ts
plamen-yordanov: made optional

packages/api/src/services/utils/routing-utils.ts
plamen-yordanov: removed and added a check inside the `forEach`

yordanalexandrov (approved):

PR #3050, opened 2026-07-09
GDB-14873: Update GraphQL Endpoint Management for Users with MANAGE_REPO Permissions

boyan-tonchev:
## What
Updated the GraphQL Endpoint Management views to support the new ROLE_MANAGE_REPO role.

## Why
The GraphQL Endpoint Management functionality was previously restricted to users with the repository manager role. With the introduction of the ROLE_MANAGE_REPO role, users with repository-specific management permissions should be able to manage GraphQL endpoints for the repositories they are authorized to manage.

## How
- Replaced the repository manager role restriction for the GraphQL management view with the new repository management permission.
- Extended the core-errors directive to support repository filtering based on the configured authorization role.
- Applied repository filtering across the Create, Manage, and Playground views when no repository is selected.
- Filtered the Source Repository dropdown to expose only repositories for which the user has management permissions in create endpoint view.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/core/directives.js
+++ b/packages/legacy-workbench/src/js/angular/core/directives.js
@@ -1,6 +1,14 @@
 import 'angular/utils/local-storage-adapter';
 import {decodeHTML} from "../../../app";
-import {service, LicenseContextService, ResourceSearchStorageService, Suggestion, GuidesService} from '@ontotext/workbench-api';
+import {
+    Authority,
+    AuthorizationService,
+    service,
+    LicenseContextService,
+    ResourceSearchStorageService,
+    Suggestion,
+    GuidesService,
+} from '@ontotext/workbench-api';
 
 angular
     .module('graphdb.framework.core.directives', [
@@ -115,6 +123,12 @@ function coreErrors($timeout, $location) {
         transclude: true,
         templateUrl: 'js/angular/core/templates/core-errors.html',
         link: function(scope, element, attrs) {
+            const authorizationService = service(AuthorizationService);
+
+            const canManageRepositoryFilter = Authority.ROLE_MANAGE_REPO === attrs.role ? (repo) => authorizationService.canManageRepo(repo) : () => true;
+            const filterByType = attrs.filterByType?.split(',');
+            const filterByTypeFilter = (repo) => !filterByType || filterByType.includes(repo.type);
+
             // watch for changes in the active repository hide host element of this directive
             scope.$watch('getActiveRepository()', function(newValue) {
                 if (newValue && !scope.isRestricted) {
@@ -146,11 +160,12 @@ function coreErrors($timeout, $location) {
                 if (!scope.showRemoteLocations) {
                     remoteLocationsFilter = (repo) => repo.local;
                 }
-                if (scope.isRestricted) {
-                    return scope.getWritableRepositories().filter(remoteLocationsFilter);
-                } else {
-                    return scope.getReadableRepositories().filter(remoteLocationsFilter);
-                }
+
+                const repositories = scope.isRestricted ? scope.getWritableRepositories() : scope.getReadableRepositories();
+                return repositories
+                    .filter(remoteLocationsFilter)
+                    .filter(canManageRepositoryFilter)
+                    .filter(filterByTypeFilter);
             };
 
             scope.showPopoverForRepo = function(event, repository) {
--- a/packages/legacy-workbench/src/js/angular/graphql/controllers/create-graphql-endpoint-view.controller.js
+++ b/packages/legacy-workbench/src/js/angular/graphql/controllers/create-graphql-endpoint-view.controller.js
@@ -12,7 +12,12 @@ import 'angular/graphql/directives/configure-endpoint.directive';
 import 'angular/graphql/directives/generate-endpoint.directive';
 import {GraphqlEventName} from "../services/graphql-context.service";
 import {resolvePlaygroundUrlWithEndpoint} from "../services/endpoint-utils";
-import {RepositoryContextService, service} from '@ontotext/workbench-api';
+import {
+    RepositoryContextService,
+    RepositoryType,
+    service,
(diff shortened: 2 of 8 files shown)


svilenvelikov (approved):

plamen-yordanov (approved):

packages/legacy-workbench/src/js/angular/core/directives.js
```diff
@@ -146,11 +160,11 @@ function coreErrors($timeout, $location) {
                 if (!scope.showRemoteLocations) {
                     remoteLocationsFilter = (repo) => repo.local;
                 }
-                if (scope.isRestricted) {
-                    return scope.getWritableRepositories().filter(remoteLocationsFilter);
-                } else {
-                    return scope.getReadableRepositories().filter(remoteLocationsFilter);
-                }
+
+                return (scope.isRestricted ? scope.getWritableRepositories() : scope.getReadableRepositories())
```
yordanalexandrov: Just my thought, this looks hard to read.

IMO it would be nicer to just assign the repositories to a constant depending on the `scope.isRestricted` and then return then while applying the filters.

packages/legacy-workbench/src/js/angular/graphql/controllers/create-graphql-endpoint-view.controller.js
```diff
@@ -284,16 +308,22 @@ function CreateGraphqlEndpointViewCtrl($scope, $location, $repositories, $transl
...
     const setupSourceRepositories = () => {
         $scope.sourceRepositories = $repositories.getRepositoriesAsSelectMenuOptions(
-            () => $repositories.getLocalReadableGraphdbRepositories(),
+            () => $repositories.getLocalReadableGraphdbRepositories()
+                .filter((repository) => authorizationService.canManageRepo(repository)),
         );
         const activeRepository = repositoryContextService.getSelectedRepository();
+        $scope.hasSelectedRepository = !!activeRepository;
         $scope.isSelectedRepositoryLocal = !!activeRepository?.local;
+        $scope.isGraphDBRepository = RepositoryType.GRAPH_DB === activeRepository.type;
```
yordanalexandrov, line 319: I am just wondering about the `OTHER` repository type that we missed before.

Can you double check if there is also a check in GDB that would not allow OTHER repositories and only GRAPH_DB?

yordanalexandrov, line 319: We should not be so restrictive.

The GQL management and playgorund are not available for ONTOP and FEDEX repositories.
The restriction should be about them, not allow onyl GRAPH_DB.

packages/legacy-workbench/src/js/angular/graphql/templates/graphql-endpoint-management.html
```diff
@@ -6,189 +6,199 @@ <h1 id="title-container">
         <page-info-tooltip></page-info-tooltip>
     </h1>
 
-    <div core-errors></div>
+    <div core-errors role="ROLE_MANAGE_REPO" allowed-repository-types="graphdb"></div>
```
yordanalexandrov: There are `ontop` and `fedex` properties of `scope.isRestricted` which might be used here to filter repositories based on type.

packages/legacy-workbench/src/js/angular/graphql/templates/graphql-playground.html
```diff
@@ -17,7 +17,7 @@ <h1 id="title-container">
         </div>
     </h1>
 
-    <div core-errors></div>
+    <div core-errors role="ROLE_MANAGE_REPO"></div>
```
yordanalexandrov: This is not needed as the playgound is free for everyone.

yordanalexandrov (changes requested):

packages/legacy-workbench/src/js/angular/core/directives.js
boyan-tonchev: Done

packages/legacy-workbench/src/js/angular/graphql/controllers/create-graphql-endpoint-view.controller.js
boyan-tonchev: Endpoints can be created only for graphdb type of repositories, confirmed with the DevEx team.

boyan-tonchev: Endpoints can be created only for graphdb type of repositories, confirmed with the DevEx team.

packages/legacy-workbench/src/js/angular/graphql/templates/graphql-endpoint-management.html
boyan-tonchev: I checked this, but I can't use the current implementation by passing repository types. isRestricted only filters writable and readable repositories, whereas here I need to filter repositories by type.

packages/legacy-workbench/src/js/angular/graphql/templates/graphql-playground.html
boyan-tonchev: Removed

packages/legacy-workbench/src/js/angular/graphql/templates/graphql-endpoint-management.html
yordanalexandrov: can you pass the enum value here?

yordanalexandrov (approved):

packages/legacy-workbench/src/js/angular/graphql/templates/graphql-endpoint-management.html
boyan-tonchev: Done

PR #3054, opened 2026-07-10
GDB-14872: Update TTYG View for Users with MANAGE_REPO Permissions

boyan-tonchev:
## What
Update Agent Management to support the new ROLE_MANAGE_REPO role and introduce per-agent permission evaluation.

## Why
The creation and editing of agents were previously restricted to users with the repository manager role. With the introduction of the ROLE_MANAGE_REPO role, the actions a user can perform depend on the repository associated with each agent rather than on a single global permission. As a result, the canModifyAgent flag in TTYGContextService is no longer sufficient.

## How
- Removed the canModifyAgent flag from TTYGContextService.
- Introduced AgentViewModel to enrich AgentModel with UI-specific permissions.
- Added a mapper that converts AgentModel to AgentViewModel and evaluates the current user's permissions for each agent based on its associated repository.
- Updated the Agent Management view to use the per-agent permissions exposed by AgentViewModel.
- Allowed users to create, edit, and delete only agents associated with repositories they have management permissions for.
- Kept agent cloning available while restricting the target repository selection to repositories the user has management permissions for.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/ttyg/services/agents.mapper.js
+++ b/packages/legacy-workbench/src/js/angular/ttyg/services/agents.mapper.js
@@ -15,6 +15,7 @@ import {
 import {NumericRangeModel, TextFieldModel} from '../../models/form-fields';
 import {md5HashGenerator} from '../../utils/hash-utils';
 import {AGENT_OPERATION} from "./constants";
+import {AgentViewModel} from '../model/agent-view';
 
 /**
  * Converts an agent model to an agent form model.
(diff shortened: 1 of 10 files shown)


packages/legacy-workbench/src/js/angular/ttyg/templates/ttyg.html
```diff
@@ -93,18 +96,18 @@ <h1 id="ttyg-view-title">
                     <i class="ri-question-line"></i>
                 </button>
                 <button class="btn btn-link btn-sm create-agent-btn"
-                        ng-disabled="!canModifyAgent"
+                        ng-disabled="!canManageRepo"
                         ng-click="onOpenNewAgentSettings()"
                         ng-disabled="false"
```
plamen-yordanov: not from this PR, but please remove the duplicate `ng-disabled`

packages/legacy-workbench/src/js/angular/ttyg/services/agents.mapper.js
```diff
@@ -246,3 +247,36 @@ const agentInstructionsMapper = (data) => {
...
+ * @param {AgentModel} agent The agent to convert.
+ * @param {AuthorizationService} authorizationService The authorization service used to determine user permissions.
+ * @returns {AgentViewModel} The converted agent view model.
+ */
+export const mapAgentModelToAgentViewModel = (agent, authorizationService) => {
+    const viewAgent = new AgentViewModel(agent);
+    const canManageRepo = authorizationService.canManageRepo({id: agent.repositoryId, location: ''});
+    const isAdminOrRepoManager = authorizationService.isAdminOrRepoManager();
+
+    viewAgent.canEditAgent = isAdminOrRepoManager || canManageRepo;
```
plamen-yordanov: I believe `canManageRepo` is enough. `isAdminOrRepoManager` is included inside it

packages/legacy-workbench/src/js/angular/ttyg/services/agents.mapper.js
```diff
@@ -246,3 +247,36 @@ const agentInstructionsMapper = (data) => {
...
+ * Converts an {@link AgentModel} into an {@link AgentViewModel} by enriching it
+ * with UI-specific permissions based on the current user's repository management rights.
+ *
+ * @param {AgentModel} agent The agent to convert.
+ * @param {AuthorizationService} authorizationService The authorization service used to determine user permissions.
+ * @returns {AgentViewModel} The converted agent view model.
+ */
+export const mapAgentModelToAgentViewModel = (agent, authorizationService) => {
+    const viewAgent = new AgentViewModel(agent);
+    const canManageRepo = authorizationService.canManageRepo({id: agent.repositoryId, location: ''});
```
plamen-yordanov, line 273: is it safe to assume local repos only?

packages/legacy-workbench/src/js/angular/ttyg/templates/ttyg.html
boyan-tonchev: Done

packages/legacy-workbench/src/js/angular/ttyg/services/agents.mapper.js
boyan-tonchev: You are right. Done

packages/legacy-workbench/src/js/angular/ttyg/services/agents.mapper.js
boyan-tonchev: Yes, an agent can be created only for local GraphDB repositories.

plamen-yordanov (approved):

PR #3062, opened 2026-07-15
Support MANAGE_REPO_xxx role in Workbench

boyan-tonchev:
## What
Add support for repository-specific MANAGE_REPO permissions across the Workbench UI.

## Why
Allow users to manage only the repositories they are authorized to manage without requiring the global repository manager role.

## How
Update authorization checks, repository filtering, page access, and UI actions to use repository-specific management permissions where applicable.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/legacy-workbench/src/js/angular/core/directives.js
```diff
@@ -115,6 +123,12 @@ function coreErrors($timeout, $location) {
         transclude: true,
         templateUrl: 'js/angular/core/templates/core-errors.html',
         link: function(scope, element, attrs) {
+            const authorizationService = service(AuthorizationService);
+
+            const canManageRepositoryFilter = Authority.ROLE_MANAGE_REPO === attrs.filterByRole ? (repo) => authorizationService.canManageRepo(repo) : () => true;
```
yordanalexandrov: This doesn't feel right.

The attribute is `filterByRole` but anything you pass to it will filter MANAGE_XXX role

yordanalexandrov: removed the hardcoded role and changed the filter to use the role that is passed as attribute

yordanalexandrov (approved):

PR #3068, opened 2026-07-20
GDB-14919: Update repository selector tooltips to reflect MANAGE_REPO permissions

boyan-tonchev:
## What
Update the repository selector item tooltips to display the current user's permission for the hovered repository.

## Why
The tooltip displays the current user's access level for the hovered repository. Currently, it only indicates whether the user has read or read/write permissions. With the introduction of the new MANAGE_REPO permission, the tooltip must also indicate when the user has management permission for the hovered repository.

## How
- Expose the `RepositoryPermissionType` enum, which represents all supported repository permission levels.
- Extend the authorization service with a method that returns the permission type granted to the current user for a given repository.
- Use the new method in the repository selector tooltips to resolve the appropriate translation for the granted permission.
- Update the tooltips in the Repositories view and `core-errors` to display repository permissions consistently with the repository selector.
- Replace the term access with permission in the tooltips to use consistent terminology.

## Screenshots
image
image
image
image
image
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/api/src/services/domain/security/authorization.service.ts
```diff
@@ -249,6 +241,40 @@ export class AuthorizationService implements Service {
     return this.canReadGqlRepo(activeRepo) || this.canWriteGqlRepo(activeRepo);
   }
 
+  /**
+   * Returns the highest permission the current user has for the repository.
+   *
+   * @param repository The repository to evaluate.
+   * @returns The highest applicable repository permission.
+   */
+  getRepositoryPermission(repository: Repository): RepositoryPermissionType {
```
svilenvelikov, line 264: There are better ways to implement this resolver. Now, the order is encoded in the if chain which is brittle. Also if I want to change/add roles i need to touch this code which would be scarry. Especially when there is no descfription why this matters.
Better would be to implement a data driven resolver.

packages/legacy-workbench/src/js/angular/templates/repositorySize.html
```diff
@@ -1,57 +1,68 @@
-<div ng-init="getRepositorySize()" style="margin: -4px 0; padding: 10px">
-    <div class="row" ng-if="popoverRepo.title" style="margin-bottom: .5rem">
-        <span>
-            {{popoverRepo.title}}
+<link href="css/repository-info-popover.css?v=[AIV]{version}[/AIV]" rel="stylesheet">
+<div ng-init="getRepositorySize()" class="repository-info-popover" style="margin: -4px 0; padding: 10px">
```
svilenvelikov: i know this is legacy code, but as far as you added a dedicated stylesheet for this dialog, why didn't you extracted these inline styles there?

packages/legacy-workbench/src/js/angular/templates/repositorySize.html
```diff
@@ -1,57 +1,68 @@
...
+        <span class="repository-id">
+            {{'security.repository.title' | translate}}
         </span>
-    </div>
-    <div class="row">
+        <span class="value">
+            {{repository.id}}
+        </span>
+        <div ng-if="popoverRepo.title" style="margin-bottom: .5rem">
+            <span class="repository-descritpin">
```
svilenvelikov: typo in `description`

packages/api/src/services/domain/security/authorization.service.ts
boyan-tonchev: Done

packages/legacy-workbench/src/js/angular/templates/repositorySize.html
boyan-tonchev: Done

packages/legacy-workbench/src/js/angular/templates/repositorySize.html
boyan-tonchev: Fixed

packages/legacy-workbench/src/js/angular/templates/repositorySize.html
```diff
@@ -1,57 +1,75 @@
...
-    <div class="row">
-        <label style="margin-right: 10px">{{'location.label' | translate}}</label>
-        <span class="text-right pull-right">
-			{{popoverRepo.location ? popoverRepo.location : 'Local'}}
-		</span>
+
+    <div class="property mb-1">
+        <span class="property-label">{{'permission.label' | translate}}</span>
+        <span class="property-value text-right">
+            {{('repository.permission.' + getRepositoryPermission(popoverRepo)) | translate}}
```
yordanalexandrov: since the enum is not directly coupled to the view, add `.toLowerCase()` so if someone adds another or changes the casing, it would still work here

packages/api/src/models/repositories/repository-permission-type.ts
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Grants read and write access to the repository.
+   */
+  WRITE = 'write',
+
+  /**
+   * Grants full repository manager permissions, including creating,
+   * deleting, and managing repositories.
+   */
+  MANAGER = 'manager',
```
yordanalexandrov: Why do we need this one here?

If we need it, why don't we also have admin?

packages/api/src/services/domain/security/authorization.service.ts
```diff
@@ -394,4 +411,44 @@ export class AuthorizationService implements Service {
       user.authorities.hasWildcardAuthority(overCurrentRepoGraphql, true)
     );
   }
+
+  /**
+   * Creates and orders the repository permission resolvers from highest to lowest priority.
+   *
+   * @returns {RepositoryPermissionResolver[]} The ordered repository permission resolvers.
+   */
+  private initPermissionResolvers(): RepositoryPermissionResolver[] {
```
yordanalexandrov, line 420: Just worndering, why is this not it's own class with direct methods that can be used instead of the client having to find a resolver and get it's permissions?

packages/api/src/models/repositories/repository-permission-type.ts
boyan-tonchev: I added it to represent all permission types from a repository's perspective. I didn't include the administrator because an administrator is always a manager of every repository.

yordanalexandrov: But why do we need it?

It is not a permission type for a repository

yordanalexandrov: You use it to show `Read/Write/Manage` for repo manager and admin, because they do not have the manage role

packages/api/src/models/repositories/repository-permission-type.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Represents the highest level of permission the current user has for a repository.
```
yordanalexandrov: This is not correct. This is an enum with all possible permissions, not the highest level.

packages/api/src/models/repositories/repository-permission-type.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export enum RepositoryPermissionType {
+  /**
+   * The user has no permissions for the repository.
+   */
+  NONE = 'none',
+
+  /**
+   * Grants read access to the repository.
```
yordanalexandrov: `Grants` is an action and this enum is only used to visuallize the type

packages/api/src/models/repositories/repository-permission-type.ts
boyan-tonchev: You are right, it is not repository permission. I will remove it.

packages/api/src/models/repositories/repository-permission-type.ts
boyan-tonchev: Fixed

packages/api/src/models/repositories/repository-permission-type.ts
boyan-tonchev: Fixed

packages/api/src/models/repositories/repository-permission-type.ts
boyan-tonchev: done

packages/legacy-workbench/src/js/angular/templates/repositorySize.html
boyan-tonchev: Done

yordanalexandrov (approved):

PR #3069, opened 2026-07-20
GDB-14793: Fix repository listing for users with MANAGE permission

boyan-tonchev:
## What
- Users with the MANAGE permission couldn't see repositories for which they had read permission.
- Users with the MANAGE permission couldn't see repositories if the selected repository had GraphQL-only permissions.

## Why
- Repository filtering was added for users with the MANAGE permission to include only repositories for which they have at least read permission. However, this filter excluded repositories for which the user has GraphQL-only permissions.
- The page was restricted to users with the repository read authority. As a result, users with only GraphQL permissions could not access the page.

## How
- Updated the repository filtering to include repositories for which the user has GraphQL read permission.
- Extended the page access restriction to allow users with GraphQL read-only permission.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


yordanalexandrov:
Please add tests for these cases. The `users-and-access` spec should have all you need to do the checks. Maybe skip the tests if there isn't a TR of GraphDB to set for the tests that has the MANAGE permission implementation.

boyan-tonchev:
> Please add tests for these cases. The `users-and-access` spec should have all you need to do the checks. Maybe skip the tests if there isn't a TR of GraphDB to set for the tests that has the MANAGE permission implementation.

Done

yordanalexandrov (approved):

PR #3076, opened 2026-07-22
GDB-14912: Reduce the size of the Solr connector banner title

boyan-tonchev:
## What
- Remove the Solr banner header.
- Update the banner message.
- Reduce the message font size.

## Why
The banner was too verbose and occupied too much space.

## How
- Remove the banner header.
- Update the banner message.
- Reduce the message font size.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/shared-components/src/assets/i18n/en.json
+++ b/packages/shared-components/src/assets/i18n/en.json
@@ -225,8 +225,9 @@
   },
   "deprecation-banner": {
     "solr": {
-      "header": "Solr Connector Deprecation Notice",
-      "content": "The Solr connector has been deprecated and will be removed in version 12 of GraphDB. If you are using it, please contact us via the available support channels to discuss the alternatives."
+      "content_1": "The Solr connector is deprecated and will be removed in GraphDB 12. If you use it,",
+      "content_2": " contact us ",
+      "content_3": "through the available support channels."
     }
   }
 }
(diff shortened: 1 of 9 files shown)


packages/shared-components/src/assets/i18n/en.json
```diff
@@ -225,8 +225,9 @@
   },
   "deprecation-banner": {
     "solr": {
-      "header": "Solr Connector Deprecation Notice",
-      "content": "The Solr connector has been deprecated and will be removed in version 12 of GraphDB. If you are using it, please contact us via the available support channels to discuss the alternatives."
+      "content_1": "The Solr connector is deprecated and will be removed in GraphDB 12. If you use it,",
+      "content_2": " contact us ",
```
yordanalexandrov: Is it not possible to use html tags in the labels for the shared components?

yordanalexandrov: There are several that use `<b>` and `<br>`

yordanalexandrov (changes requested):

packages/shared-components/src/assets/i18n/en.json
boyan-tonchev: It is used by the translated label component, which currently doesn't support it. Should I update it?

yordanalexandrov: Yes, improve it so we don't have this please.

maybe add a prop `allowHTML` or do it for all labels

boyan-tonchev: Done

yordanalexandrov (approved):

PR #3080, opened 2026-07-22
GDB-14855: Prevent repository sync on login page

yordanalexandrov:
## What
Added a condition to stop repository synchronization while on the login page.

## Why
To avoid unnecessary operations and potential errors when the user is not logged into the application.

## How
- Integrated `isLoginPage()` check in `repository-url-sync.service.ts` to bypass sync logic if on the login page.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/workbench/src/app/services/repository-url-sync.service.ts
+++ b/packages/workbench/src/app/services/repository-url-sync.service.ts
@@ -5,7 +5,7 @@ import {
   getCurrentRoute,
   REPOSITORY_ID_PARAM,
   RepositoryContextService,
-  WindowService,
+  WindowService, isLoginPage,
 } from '@ontotext/workbench-api';
 import {ConfirmationProviderService} from './dialog/confirmation-provider.service';
 import {TranslocoService} from '@jsverse/transloco';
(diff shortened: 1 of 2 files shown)


boyan-tonchev (dismissed):

boyan-tonchev (approved):

packages/workbench/src/app/services/repository-url-sync.service.ts
```diff
@@ -47,6 +47,10 @@ export class RepositoryUrlSyncService {
     const isSameRepository = !!selectedRepository && repositoryIdParam === selectedRepository.id;
 
     try {
+      // --- if on login page, do nothing
+      if (isLoginPage()) {
```
plamen-yordanov, line 51: should this still do nothing if we have free access?

yordanalexandrov: The only case I could think of is when the free user goes to the login page and changes the repository.

In this case it is handled in a different way and the repositoryID is appended.

So this even saves a run through the ifs of the method in such case.

plamen-yordanov (approved):

PR #3104, opened 2026-07-28
GDB-14938: Remember the state of the Solr banner in WB

boyan-tonchev:
## What
Changed how the Solr deprecation banner state is persisted.

## Why
The requirements have been revised:
- When security is enabled, the banner state is persisted per user. If User 1 closes the banner, they will not see it again in the current or subsequent sessions. If another user logs in, they will see the banner until they close it.
- When security is disabled, the banner state is global. Once any user closes the banner, it will never be displayed again.
- When security is enabled, free access is enabled, and the user is not logged in, the behavior is the same as when security is disabled.
- When a guide is started, the banner is hidden regardless of whether the user has previously closed it.
- When the guide finishes, the banner is shown or hidden depending on whether the user had previously dismissed it.

## How
- The UserPreferencesService has been updated to store preferences for the shared anonymous user whenever there is no currently logged-in user (that is, when security is disabled or when free access is enabled). This allows all anonymous users to share the same persisted preferences through the anonymous preference key.
- The guide functionality has also been updated to expose whether a guide is currently running:
 - GuideContextService now includes a new HAS_RUNNING_GUIDE field that indicates whether a guide is currently running.
 - GuideService now updates the guide context when a guide starts or finishes, regardless of whether it completes normally or due to an exception.
- The onto-layout component now listens for changes to the guide state and hides the Solr banner while a guide is running. When the guide finishes, the banner is shown again only if it has not been previously dismissed by the current user.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/api/src/services/user-preferences/user-preferences.service.ts
+++ b/packages/api/src/services/user-preferences/user-preferences.service.ts
@@ -1,41 +1,43 @@
 import { service } from '../../providers';
 import { UserPreferencesStorageService } from './user-preferences-storage.service';
-import {AuthenticationService, SecurityContextService} from '../domain/security';
+import { AuthenticationService, SecurityContextService } from '../domain/security';
 import { UserPreferencesContextService } from './user-preferences-context.service';
 import { UserPreferences } from '../../models/user-preference/user-preferences';
 
 /**
  * Service for loading, updating, and persisting user preferences.
  *
- * Preferences are persisted only when the current user is authenticated.
- * Anonymous users keep their preferences only in the current application context,
- * so their choices are reset after a page refresh.
+ * Preferences are stored per current user:
+ * - A currently logged-in user has an individual set of preferences.
+ * - The anonymous user, used when security is disabled or free access is enabled,
+ *   has a shared persisted set of preferences identified by the {@link ANONYMOUS_USER} key.
  */
 export class UserPreferencesService {
+
+  private static readonly ANONYMOUS_USER = 'anonymous';
+
   private readonly userPreferencesStorageService = service(UserPreferencesStorageService);
   private readonly userPreferencesContextService = service(UserPreferencesContextService);
   private readonly securityContextService = service(SecurityContextService);
   private readonly authenticationService = service(AuthenticationService);
 
   /**
-   * Marks the Solr deprecation banner as dismissed.
+   * Marks the Solr deprecation banner as dismissed for the current user.
    *
-   * For authenticated users, the preference is persisted and restored after refresh.
-   * For anonymous users, including non-secured access and free-access mode, the preference is stored
-   * only in the current application context.
+   * The preference is updated in the application context and persisted for the current user.
+   * A currently logged-in user has individual preferences, while the anonymous user shares
+   * a common persisted set of preferences.
    */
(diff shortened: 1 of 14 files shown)


packages/api/src/services/user-preferences/user-preferences.service.ts
```diff
@@ -46,54 +48,74 @@ export class UserPreferencesService {
...
+   * @returns The username of the current user, or {@code undefined} if the username
+   * of the currently logged-in user cannot be determined.
    */
-  private loadCurrentUserPreferences(): UserPreferences | undefined {
-    const username = this.getAuthenticatedUsername();
-
-    // The username is expected to be available for authenticated users, but the service API allows it to be undefined.
-    if (!username) {
-      return new UserPreferences();
+  private getCurrentUsername(): string | undefined {
```
yordanalexandrov, line 76: Is it possible to return undefined?

packages/shared-components/src/components/onto-layout/onto-layout.tsx
```diff
@@ -250,27 +253,31 @@ export class OntoLayout {
...
         this.updateVisibility();
+        this.updateSolrDeprecationBannerVisibility();
       })
     );
 
     this.subscriptions.add(
       this.eventService.subscribe(EventName.LOGOUT, () => {
         this.setNavbarItemVisibility();
         this.updateVisibility();
+        this.updateSolrDeprecationBannerVisibility();
```
yordanalexandrov: Why don't you put this inside `updateVisibility` method?

packages/api/src/services/user-preferences/user-preferences.service.ts
boyan-tonchev: Yes, because of the API `this.securityContextService.getAuthenticatedUser()?`

packages/shared-components/src/components/onto-layout/onto-layout.tsx
boyan-tonchev: Done

yordanalexandrov (approved):

packages/shared-components/src/components/onto-layout/onto-layout.tsx
```diff
@@ -191,7 +194,7 @@ export class OntoLayout {
   private onSolrDeprecationBannerClosedHandler(): () => void {
     return () => {
       this.userPreferencesService.dismissSolrDeprecationBanner();
-      this.updateShowSolrDeprecationBanner();
+      this.updateSolrDeprecationBannerVisibility();
```
yordanalexandrov: Remove this as the line before triggers the user preferences change

yordanalexandrov (dismissed):

packages/shared-components/src/components/onto-layout/onto-layout.tsx
boyan-tonchev: Removed

yordanalexandrov (dismissed):

yordanalexandrov (approved):

PR #3114, opened 2026-07-29
GDB-14973: Add repository to ontop related update endpoints in GDB Workbench

boyan-tonchev:
## What
Updated the file-related Ontop repository endpoints to include the repository identifier in the request body.

## Why
The previous endpoints were not repository-specific, preventing users with `MANAGE_REPO` permissions from accessing them. Including the repository identifier allows authorization to be evaluated per repository.

## How
- Updated `rest/repositories/file/upload`, `rest/repositories/file`, `rest/repositories/file/update`, and `rest/repositories/ontop/jdbc-properties` to pass the repository identifier in the request body.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/repositories/ontop-repo.directive.js
+++ b/packages/legacy-workbench/src/js/angular/repositories/ontop-repo.directive.js
@@ -6,6 +6,7 @@ import {OntopRepositoryError} from "../models/ontop/ontop-repository-error";
 import {OntopDriverData} from "../models/ontop/ontop-driver-data";
 import {JdbcDriverType} from "../models/ontop/jdbc-driver-type";
 import {LoggerProvider} from "../core/services/logger-provider";
+import {RepositoryReference} from '../models/repository/repository';
 
 const logger = LoggerProvider.logger;
 // A link to Ontop's website with all Ontop configuration keys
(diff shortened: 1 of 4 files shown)


packages/legacy-workbench/src/js/angular/repositories/ontop-repo.directive.js
```diff
@@ -371,9 +381,19 @@ function ontopRepoDirective($uibModal, RepositoriesRestService, toastr, Upload,
             });
         };
 
+        const getRepositoryReference = (repositoryInfo) => {
+            if ($scope.editRepoPage) {
```
svilenvelikov, line 393: Here you assume the RepoositoryInfo type from the api module, but it's unclear if you don't know about it.
Can you define a type here in the legacy and use it for the variable above and the return type here. Otherwise the API might diverge overtime.

yordanalexandrov (dismissed):

packages/legacy-workbench/src/js/angular/repositories/ontop-repo.directive.js
boyan-tonchev: Done

yordanalexandrov (approved):

PR #3121, opened 2026-07-30
GDB-14831 add local store persistence to Reactodia

plamen-yordanov:
## What
Add local store persistence to Reactodia.

## Why
This change improves user experience by ensuring that the diagram state is preserved across page refreshes.

## How
Added a subscription to navigation end event in `reactodia-page.component.ts` to clear the diagram state, when a user has navigated away from the view manually. Refreshes preserve the previous diagram

## Testing
cypress

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/workbench/src/app/pages/reactodia/reactodia-page.component.ts
+++ b/packages/workbench/src/app/pages/reactodia/reactodia-page.component.ts
@@ -1,13 +1,17 @@
 import {Component, inject, OnDestroy, OnInit, signal} from '@angular/core';
 import {ActivatedRoute} from '@angular/router';
 import {
+  EventName,
+  EventService,
   GraphExploreLink,
   GraphExploreService,
   LanguageContextService,
   RepositoryContextService,
   service,
-  SubscriptionList
+  SubscriptionList,
+  WindowService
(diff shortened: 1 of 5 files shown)


packages/workbench/src/app/pages/reactodia/reactodia-page.component.ts
```diff
@@ -70,6 +76,16 @@ export class ReactodiaPageComponent implements OnInit, OnDestroy {
     });
   }
 
+  /**
+   * Tells the `graphwise-reactodia` web component to drop its persisted diagram state whenever a
+   * navigation ends, so a stale diagram is not restored on the next visit to the page.
+   */
+  private subscribeToNavigationEnd() {
```
yordanalexandrov, line 83: This is subscription to navigationEnd based on the method name and to navigationStart based on the logic inside it.

Make it the same.

plamen-yordanov: done

yordanalexandrov (approved):

PR #3149, opened 2026-08-06
GDB-14387: Update resource view for triple terms

yordanalexandrov:
## What
Enhance the resource view to support triple terms, including role management and display adjustments.

## Why
This update allows for better handling and visualization of triple terms in the resource view, improving user experience and functionality.

## How
- Added `TRIPLE_TERM` role to the role types.
- Updated role tab visibility logic to conditionally render based on resource type.
- Implemented methods to shorten IRIs of triple terms and build lookup links for SPARQL queries.
- Adjusted tests to validate new behavior for triple terms in the resource view.

## Testing
added
Must be tested with RDF 1.2 compatible release of GDB

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests


e2e-tests/e2e-legacy/resource/resource.spec.js
```diff
@@ -330,54 +359,149 @@ describe('Resource view', () => {
...
+
+        it('should not offer the "triple term" role tab when the resource is not a triple term', () => {
+            // When I load a resource which is not a triple term.
+            ResourceSteps.visit(`uri=${SUBJECT_RESOURCE_ENCODED}&role=subject`);
+
+            // Then I expect the "triple term" tab to not be rendered at all, because that role
+            // applies to triple terms only,
+            ResourceSteps.verifyRoleTabMissing(TRIPLE_TERM_ROLE_LABEL);
+            // and only the roles of a plain resource to be offered.
+            ResourceSteps.getRoleTabs().should('have.length', ResourceSteps.getAllRoles().length);
```
boyan-tonchev: Shouldnt  we update the getAllRoles method to return the new triple term?

packages/legacy-workbench/src/js/angular/explore/controllers.js
```diff
@@ -342,7 +371,22 @@ function ExploreCtrl(
...
+     * resource in favour of the default one.
+     *
+     * @return {string} the role of the loaded resource
+     */
+    const resolveRole = () => {
+        if ($scope.resourceInfo.triple) {
+            return RoleType.TRIPLE_TERM;
+        }
+        const role = $location.search().role;
+        return role && role !== RoleType.TRIPLE_TERM ? role : RoleType.SUBJECT;
```
boyan-tonchev: We can extend the check to verify that the role is one of the expected roles. If it is not, return the subject.

packages/legacy-workbench/src/pages/explore.html
```diff
@@ -125,6 +125,7 @@ <h1 fit-text fit-text-min="30">
             </div>
             <onto-graph-explore-split-button
                 ng-custom-element
+                ngce-prop-disabled="isLoading && resourceInfo.triple"
```
boyan-tonchev: Shouldn't this be `ngce-prop-disabled="isLoading || !!resourceInfo.triple"`?

boyan-tonchev (approved):

packages/legacy-workbench/src/pages/explore.html
yordanalexandrov: Yes, it is a mistake

PR #3150, opened 2026-08-07
GDB-15032: Introduce a mechanism that loads the correct labels in the new workbench

boyan-tonchev:
## What
Add version-based cache busting for language bundles

## Why
Ensure users receive updated translations after a new application build instead of reusing cached language JSON files.

## How
- Extend the language configuration model with a new `languageBundleVersion` property. This value is used as the version parameter when loading language bundles.
- The default configuration sets the value to `[AIV]{version}[/AIV]`, which is replaced with the actual Workbench version during the build process. This mechanism is already used for other resources to prevent browsers from serving stale cached files.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/api/src/services/language/language-rest.service.ts
+++ b/packages/api/src/services/language/language-rest.service.ts
@@ -12,10 +12,14 @@ export class LanguageRestService extends HttpService {
    * Retrieves the translation bundle for a specific language.
    *
    * @param languageCode - The code of the language for which to fetch translations.
+   * @param languageBundleVersion - A language bundle version identifier used for cache busting
    * @returns A Promise that resolves to a TranslationBundle containing the translations for the specified language.
    */
-  getLanguage(languageCode: string): Promise<TranslationBundle> {
-    return this.get<TranslationBundle>(`${this.I18N_ENDPOINT}/${languageCode}.json`);
+  getLanguage(languageCode: string, languageBundleVersion?: string): Promise<TranslationBundle> {
+    return this.get<TranslationBundle>(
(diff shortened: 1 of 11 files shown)


yordanalexandrov (dismissed):

packages/api/src/services/language/language-rest.service.ts
```diff
@@ -12,10 +12,14 @@ export class LanguageRestService extends HttpService {
...
    * @param languageCode - The code of the language for which to fetch translations.
+   * @param languageBundleVersion - A language bundle version identifier used for cache busting
    * @returns A Promise that resolves to a TranslationBundle containing the translations for the specified language.
    */
-  getLanguage(languageCode: string): Promise<TranslationBundle> {
-    return this.get<TranslationBundle>(`${this.I18N_ENDPOINT}/${languageCode}.json`);
+  getLanguage(languageCode: string, languageBundleVersion?: string): Promise<TranslationBundle> {
+    return this.get<TranslationBundle>(
+      `${this.I18N_ENDPOINT}/${languageCode}.json`,
+      languageBundleVersion ? {params: {version: languageBundleVersion}} : undefined
```
yordanalexandrov, line 21: other endpoints use `v` parameter

boyan-tonchev: Yes, the parameter name is not important. I named it version to make it more readable.

yordanalexandrov (approved):

PR #3162, opened 2026-08-11
GDB-14998: Add a scrollbar to the Cluster Settings sidebar for identifier tags

boyan-tonchev:
## What
Added scrolling to the identifier tags table.

## Why
When there were many tags, users could not see all of them.

## How
- Moved the multi-region-specific CSS for displaying tags into its own file.
- Reorganized the template and CSS to display a scrollbar when there are many tags.
- Made the table header and the first row, which contains the input for adding new tags, sticky so they remain visible while scrolling.
- Updated the error messages to make tag-name errors visible.
- Improved the error styling because the previous style made the messages difficult to notice.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/css/cluster-configuration/cluster-configuration.css
+++ b/packages/legacy-workbench/src/css/cluster-configuration/cluster-configuration.css
@@ -1,0 +1,23 @@
+.cluster-configuration {
+     /*Add top margin to match the other side panels.*/
+    margin-top: 48px;
+    height: calc(100vh - 48px);
+    color: var(--gw-panel-color);
+    background-color: var(--gw-panel-background);
+    border: 1px solid var(--gw-panel-border-color);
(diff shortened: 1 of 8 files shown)


packages/legacy-workbench/src/css/cluster-configuration/cluster-configuration.css
```diff
@@ -0,0 +1,23 @@
+.cluster-configuration {
+     /*Add top margin to match the other side panels.*/
+    margin-top: 48px;
```
yordanalexandrov, line 3: Isn't there a header height property?

yordanalexandrov (approved):

packages/legacy-workbench/src/css/cluster-configuration/cluster-configuration.css
boyan-tonchev: I am not sure whether this is needed at all. I just made it consistent with the others.

PR #3174, opened 2026-08-14
GDB-14999: Introduce a way to pass a null values for temperature and top P

boyan-tonchev:
## What
Add support for removing the configured temperature and top P values in the TTYG Create/Edit Agent dialog.

## Why
Currently, once a value is set for temperature or top P, it cannot be removed.

## How
Add controls for enabling and disabling the temperature and top P settings. When disabled, their values are omitted when the agent configuration is saved.

## Screenshots
| Temperature | top P |
|-------------|-----------------------------|
|image|image|
|image|image|

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/e2e-tests/e2e-legacy/ttyg/create-agent.spec.js
+++ b/e2e-tests/e2e-legacy/ttyg/create-agent.spec.js
@@ -172,6 +172,60 @@ describe('TTYG create new agent', () => {
         TTYGViewSteps.getAgent(0).should('contain', 'Test Agent').and('contain', 'starwars');
     });
 
+    it('should create an agent without temperature or topP parameters', () => {
+        // GIVEN: I have opened the Create Agent dialog and filled in the minimum required fields.
+        openCreateAgentDialog(repositoryId)
+        TtygAgentSettingsModalSteps.getDialog().should('be.visible');
+
+        fillAgentName('Test Agent');
+        TtygAgentSettingsModalSteps.enableSparqlExtractionMethod();
+        TtygAgentSettingsModalSteps.selectSparqlMethodSparqlQuery();
+
+        // WHEN: I disable the temperature setting.
+        TtygAgentSettingsModalSteps.toggleTemperature();
+
+        // AND: I disable the topP setting.
+        TtygAgentSettingsModalSteps.toggleTop();
+
+        // AND: I save the agent.
+        TTYGStubs.stubAgentCreate(1000);
+        TTYGStubs.stubAgentListGet(
+            '/ttyg/agent/get-agent-list-new-agent.json',
+            1000,
+        );
+
+        TtygAgentSettingsModalSteps.saveAgent();
+        TtygAgentSettingsModalSteps.getCreatingAgentLoader().should('be.visible');
+
+        // THEN: I expect the create-agent request not to contain the temperature and topP parameters.
+        cy.wait('@create-agent').then((interception) => {
+            // eslint-disable-next-line no-undef
+            assert.deepEqual(interception.request.body, {
+                id: 'id',
+                name: 'Test Agent',
+                repositoryId: 'starwars',
+                model: 'gpt-4o',
+                contextSize: 128000,
+                seed: 0,
+                assistantsInstructions: {
+                    systemInstruction: '',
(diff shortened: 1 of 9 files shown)


e2e-tests/e2e-legacy/ttyg/create-agent.spec.js
```diff
@@ -172,6 +172,60 @@ describe('TTYG create new agent', () => {
...
+    it('should create an agent without temperature or topP parameters', () => {
+        // GIVEN: I have opened the Create Agent dialog and filled in the minimum required fields.
+        openCreateAgentDialog(repositoryId)
+        TtygAgentSettingsModalSteps.getDialog().should('be.visible');
+
+        fillAgentName('Test Agent');
+        TtygAgentSettingsModalSteps.enableSparqlExtractionMethod();
+        TtygAgentSettingsModalSteps.selectSparqlMethodSparqlQuery();
+
+        // WHEN: I disable the temperature setting.
```
plamen-yordanov, line 184: When I create a new agent, both settings are disabled initially. Here you disable them manually. What am I missing?

packages/legacy-workbench/src/js/angular/models/ttyg/agent-form.js
```diff
@@ -57,11 +57,26 @@ export class AgentFormModel {
          * @private
          */
         this._temperature = data && data.temperature || new NumericRangeModel({value: 0.7, minValue: 0, maxValue: 2, step: 0.1});
+
+        /**
+         *
+         * @type {boolean}
+         * @private
+         */
+        this._temperatureEnabled = !!this._temperature.value;
```
plamen-yordanov: Shouldn't we be able to set 0 as a value?

packages/legacy-workbench/src/js/angular/ttyg/services/agents.mapper.js
```diff
@@ -35,8 +35,19 @@ export const agentFormModelMapper = (agentModel, defaultAgentModel, operation )
...
     agentFormModel.contextSize = agentModel.contextSize || defaultAgentModel.contextSize;
     agentFormModel.contextSizeCopy = defaultAgentModel.contextSize;
-    agentFormModel.temperature.value = agentModel.temperature !== undefined ? agentModel.temperature : defaultAgentModel.temperature;
-    agentFormModel.topP.value = agentModel.topP !== undefined ? agentModel.topP : defaultAgentModel.topP;
+
+    // Set temperature related properties.
+    const temperatureFormModel = new NumericRangeModel({...DEFAULT_TEMPERATURE_RANGE});
+    temperatureFormModel.value = agentModel.temperature !== undefined ? agentModel.temperature : defaultAgentModel.temperature;
+    agentFormModel.temperature = temperatureFormModel;
+    agentFormModel.temperatureEnabled = agentFormModel.temperature.value !== undefined && agentFormModel.temperature.value !== null;
```
plamen-yordanov: In `agent-form` we set these with `!!value`, so one of these is not correct. Also you can use `ObjectUtils#isNullOrUndefined` if this is the correct approach

plamen-yordanov (changes requested):

e2e-tests/e2e-legacy/ttyg/create-agent.spec.js
boyan-tonchev: The TTYG tests are stubbed, and the fixture that provides the default agent settings includes temperature and topP. See https://github.com/Ontotext-AD/graphdb-workbench/blob/master/e2e-tests/fixtures/ttyg/agent/get-agent-defaults.json#L8

packages/legacy-workbench/src/js/angular/models/ttyg/agent-form.js
boyan-tonchev: Fixed

packages/legacy-workbench/src/js/angular/ttyg/services/agents.mapper.js
boyan-tonchev: Fixed

plamen-yordanov (approved):

PR #3185, opened 2026-08-25
GDB-14969 update user permissions table header layout

plamen-yordanov:
## What
Update the user permissions table layout

## Why
It was missed during implementation

## How
Modified the `user.html` file to adjust the table structure and added a new column for permissions.

## Testing
n/a

## Screenshots
Screenshot from 2026-08-25 16-40-48
Screenshot from 2026-08-25 16-40-55

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/e2e-tests/e2e-legacy/setup/users-and-access/user-and-access.spec.js
+++ b/e2e-tests/e2e-legacy/setup/users-and-access/user-and-access.spec.js
@@ -128,7 +128,6 @@ describe('User and Access', () => {
                 UserAndAccessSteps.getCustomRoleFieldError().should('not.be.visible');
 
                 // When I create the user with a valid custom role
-                UserAndAccessSteps.toggleWriteAccessAny();
                 SecurityStubs.spyOnUserCreate();
                 UserAndAccessSteps.confirmUserCreate();
                 // Then the user should be created with that custom role
--- a/packages/legacy-workbench/src/css/user.css
+++ b/packages/legacy-workbench/src/css/user.css
@@ -3,7 +3,8 @@
 }
 
 .wb-user .repository-column {
-    width: 50%;
+    width: 40%;
+    vertical-align: top;
 }
 
 .wb-user .read-column {
@@ -22,6 +23,8 @@
     width: 15%;
 }
 
+.wb-user .permissions-column,
+.wb-user .read-column,
 .wb-user .graphql-column,
 .wb-user .graphql-rights,
 .wb-user .graphql-any {
@@ -35,8 +38,12 @@
 
 .wb-user .bordered-table {
     border: 1px solid var(--gw-datatable-border-color);
-    border-collapse: collapse;
+    border-collapse: separate;
     width: 100%;
+}
+
+.wb-user .permissions-subheader th {
+    border-bottom: 1px solid var(--gw-datatable-header-cell-border-color);
 }
 
 .wb-user .graphql-icon {
--- a/e2e-tests/steps/setup/user-and-access-steps.js
+++ b/e2e-tests/steps/setup/user-and-access-steps.js
@@ -1,5 +1,7 @@
 import {EnvironmentStubs} from "../../stubs/environment-stubs";
 
+const CENTER_SCROLL = {scrollBehavior: 'center'};
+
 export class UserAndAccessSteps {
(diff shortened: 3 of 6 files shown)


packages/legacy-workbench/src/css/user.css
```diff
@@ -4,6 +4,11 @@
 
 .wb-user .repository-column {
     width: 50%;
+    vertical-align: top;
+}
+
+.wb-user .permissions-column {
+    width: 45%;
```
yordanalexandrov: This is a bit strange.
The `repository-column` is 50%.
Then we have the `graphql-column` column at 15%.

And this is 45%?
this is a total of 110%

IMO we should set either only the `repository-column` and let the others spread, or set the permissions columns and let the `repository-column` spread.

packages/legacy-workbench/src/css/user.css
```diff
@@ -35,10 +42,14 @@
 
 .wb-user .bordered-table {
     border: 1px solid var(--gw-datatable-border-color);
-    border-collapse: collapse;
+    border-collapse: separate;
```
yordanalexandrov, line 41: Why do we want the borders to be set to `separate`?

packages/legacy-workbench/src/css/user.css
```diff
@@ -35,10 +42,14 @@
...
     width: 100%;
 }
 
+.wb-user .scrollable-table .bordered-table thead tr:nth-of-type(2) th {
```
yordanalexandrov: this is strange selector

Why not add a class to the table row, it is not a generic table

packages/legacy-workbench/src/css/user.css
plamen-yordanov: Because the manage section was squeezed and the icon appeared on the second row. This just widened it. I'll delete it and reduce the repository column

packages/legacy-workbench/src/css/user.css
plamen-yordanov: whenever there is a scroll present and it is used, some borders disappear. This is present in the current implementation. `separate` fixes that

packages/legacy-workbench/src/css/user.css
plamen-yordanov: reduced the repository column to 40% and removed this

packages/legacy-workbench/src/css/user.css
plamen-yordanov: this is how it is styled globally in `core.less`. I didn't want to risk changing the global style for `scrollable-table` so I added an override in the `user.css`. Added a class

yordanalexandrov (dismissed):

yordanalexandrov (approved):

e2e-tests/e2e-legacy/setup/users-and-access/user-and-access.spec.js
```diff
@@ -128,7 +128,6 @@ describe('User and Access', () => {
                 UserAndAccessSteps.getCustomRoleFieldError().should('not.be.visible');
 
                 // When I create the user with a valid custom role
-                UserAndAccessSteps.toggleWriteAccessAny();
```
yordanalexandrov, line 131: Why this change?

yordanalexandrov: And why does it pass!?

yordanalexandrov: Ok, I saw it.

Now I am wondering why did it work before

yordanalexandrov (dismissed):

e2e-tests/e2e-legacy/setup/users-and-access/user-and-access.spec.js
plamen-yordanov: because the `realClick`s are flaky

plamen-yordanov: only one of the two worked

yordanalexandrov (approved):

PR #3187, opened 2026-08-26
GDB-14999 preserve agent temperature and topP values when toggling

yordanalexandrov:
## What
Ensure agent temperature and topP fields preserve their current values when toggled and initialize based on whether an agent is new or existing.

## Why
When editing an agent, the temperature and topP fields appeared reset to defaults in the form and saving overwrote the persisted values with default settings.

## How
- Updated packages/legacy-workbench/src/js/angular/ttyg/controllers/agent-settings-modal.controller.js to cache and restore temperature and topP values when toggles are switched and to recalculate the high temperature warning on model updates.
- Updated packages/legacy-workbench/src/js/angular/ttyg/services/agents.mapper.js to initialize temperature and topP from defaults only when creating a new agent and from the existing agent model when editing.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
+++ b/packages/legacy-workbench/src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
@@ -103,6 +103,8 @@ function AgentSettingsModalController(
     // =========================
 
     const CHAT_GPT_RETRIEVAL_CONNECTOR_NAME = 'ChatGPT Retrieval';
+    let lastTemperatureValue = DEFAULT_TEMPERATURE_RANGE.value;
+    let lastTopPValue = DEFAULT_TOP_P_RANGE.value;
 
     // =========================
     // Public variables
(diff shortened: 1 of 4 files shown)


packages/legacy-workbench/src/js/angular/ttyg/controllers/agent-settings-modal.controller.js
```diff
@@ -103,6 +103,8 @@ function AgentSettingsModalController(
     // =========================
 
     const CHAT_GPT_RETRIEVAL_CONNECTOR_NAME = 'ChatGPT Retrieval';
+    let agentTemperatureValue = DEFAULT_TEMPERATURE_RANGE.value;
```
plamen-yordanov: Rename these to `lastTemperatureValue` and `lastTopPValue` or something along those lines, so their purpose is more clear

yordanalexandrov: done

plamen-yordanov (approved):

PR #3188, opened 2026-08-27
GDB-15106 handle cluster without leader in multi-region tab

yordanalexandrov:
## What
Handle clusters without an elected leader in the multi-region configuration tab and add an informational alert when no topology is available.

## Why
The multi-region tab threw "Cannot read properties of undefined (reading 'topologyStatus')" when the cluster had no elected leader due to slow loading.

## How
- Updated multi-region.directive.js to safely set `$scope.topology` using optional chaining when no leader is present and added a debug log for the cluster configuration response.
- Adjusted multi-region.html to show an info alert when `topology` is missing and guard primary/secondary sections with explicit `topology` checks.
- Refined layout in multi-region.css to move the height constraint from the container to `.multi-region-content`.

## Testing
Added

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/clustermanagement/templates/cluster-configuration/multi-region.html
+++ b/packages/legacy-workbench/src/js/angular/clustermanagement/templates/cluster-configuration/multi-region.html
@@ -3,7 +3,11 @@
 <div onto-loader ng-if="loader" class="multi-region-loader" message="getLoaderMessage" size="100"></div>
 
 <div class="multi-region">
-    <div class="multi-region-content" ng-if="topology.state === TopologyState.PRIMARY_NODE">
+    <div ng-if="!topology" class="no-leader-alert">
+        <div class="alert alert-info">{{'cluster_management.cluster_configuration_multi_region.leader_not_elected' | translate}}</div>
+    </div>
+
+    <div ng-if="topology.state === TopologyState.PRIMARY_NODE" class="multi-region-content">
         <h4 class="title">{{'cluster_management.cluster_configuration_multi_region.primary_cluster' | translate}}
             <span ng-if="statusLoader" class="ri-reset-left-line icon-1-25x loader cluster-status-loader"></span>
         </h4>
--- a/e2e-tests/e2e-legacy/cluster/cluster-configuration/cluster-configuration-multi-region.spec.js
+++ b/e2e-tests/e2e-legacy/cluster/cluster-configuration/cluster-configuration-multi-region.spec.js
@@ -64,6 +64,26 @@ describe('Cluster configuration', () => {
         });
     });
 
+    it('should show an alert when there is no elected leader in the cluster', () => {
+        // Given there is no elected leader in the cluster
+        ClusterStubs.stubClusterGroupStatusWithoutLeader();
+
+        // When I have opened the cluster management page
+        ClusterPageSteps.visit();
+        // And I click on edit properties and open Multi-region tab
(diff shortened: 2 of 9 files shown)


packages/legacy-workbench/src/js/angular/clustermanagement/templates/cluster-configuration/multi-region.html
```diff
@@ -108,7 +112,7 @@ <h4 class="title">{{'cluster_management.cluster_configuration_multi_region.prima
         </form>
     </div>
 
-    <div ng-if="topology.state === TopologyState.SECONDARY_NODE">
+    <div ng-if="true && topology.state === TopologyState.SECONDARY_NODE">
```
plamen-yordanov: remove this

packages/legacy-workbench/src/js/angular/clustermanagement/templates/cluster-configuration/multi-region.html
```diff
@@ -3,7 +3,11 @@
 <div onto-loader ng-if="loader" class="multi-region-loader" message="getLoaderMessage" size="100"></div>
 
 <div class="multi-region">
-    <div class="multi-region-content" ng-if="topology.state === TopologyState.PRIMARY_NODE">
+    <div ng-if="!topology" class="no-leader-alert">
+        <div class="alert alert-info">{{'cluster_management.cluster_configuration_multi_region.leader_not_elected' | translate}}</div>
+    </div>
+
+    <div ng-if="true && topology.state === TopologyState.PRIMARY_NODE" class="multi-region-content">
```
plamen-yordanov: remove ng-if="true

packages/legacy-workbench/src/js/angular/clustermanagement/templates/cluster-configuration/multi-region.html
```diff
@@ -3,7 +3,11 @@
 <div onto-loader ng-if="loader" class="multi-region-loader" message="getLoaderMessage" size="100"></div>
 
 <div class="multi-region">
-    <div class="multi-region-content" ng-if="topology.state === TopologyState.PRIMARY_NODE">
+    <div ng-if="!topology" class="no-leader-alert">
```
plamen-yordanov, line 6: I'm wondering if it would be a good idea to add an additional check for loaded topology. It's undefined at start and may load slow, or not at all. The message would be visible then for the wrong reasons

packages/legacy-workbench/src/js/angular/clustermanagement/templates/cluster-configuration/multi-region.html
yordanalexandrov: done

packages/legacy-workbench/src/js/angular/clustermanagement/templates/cluster-configuration/multi-region.html
yordanalexandrov: done

packages/legacy-workbench/src/js/angular/clustermanagement/templates/cluster-configuration/multi-region.html
yordanalexandrov: won't do

too much effort, too little reward

plamen-yordanov (approved):

PR #3200, opened 2026-09-02
GDB-15131 fix flaky repositories import test selectors

yordanalexandrov:
## What
Make import server files E2E tests use index-based selectors instead of text-based matching and add index-specific classes to import resource rows.

## Why
The repositories import tests were flaky because selectors relied on text matching, which sometimes targeted the wrong resource row.

## How
- e2e-tests/steps/import/import-steps.js: Replaced generic getResource with getResourceTitleRow, added index-based resource accessors, and updated selection, status checks, delete, import, and reset helpers to use index-based queries.
- packages/legacy-workbench/src/js/angular/import/templates/import-resource-tree.html: Added resource-{{$index}} class to both title and info rows to allow stable index-based selection in tests.
- e2e-tests/e2e-legacy/import/import-server-files.spec.js: Updated all expectations to use getResourceTitleRow and rely on index-based selection instead of text matching.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/import/templates/import-resource-tree.html
+++ b/packages/legacy-workbench/src/js/angular/import/templates/import-resource-tree.html
@@ -146,7 +146,7 @@
         </thead>
         <tbody>
         <tr ng-repeat-start="resource in displayResources track by resource.importResource.hash"
-            class="row title-row"
+            class="row title-row" data-test="resource-{{$index}}"
             ng-class="{ 'even-row-start': $even, 'odd-row-start': $odd}">
             <td class="cell cell-select">
                 <input type="checkbox" ng-model="resource.selected" ng-click="selectionChanged(resource)"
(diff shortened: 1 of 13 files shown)


packages/legacy-workbench/src/js/angular/import/templates/import-resource-tree.html
```diff
@@ -146,7 +146,7 @@
         </thead>
         <tbody>
         <tr ng-repeat-start="resource in displayResources track by resource.importResource.hash"
-            class="row title-row"
+            class="row title-row resource-{{$index}}"
```
plamen-yordanov: why class and not `data-test`?

yordanalexandrov: right, I will change it

yordanalexandrov: done

plamen-yordanov (approved):

PR #3202, opened 2026-09-02
GDB-15134: Censor connector query secrets

yordanalexandrov:
## What
Censored sensitive fields in the query visualization if marked as `sensitive`.

## Why
To prevent unintended exposure of sensitive information, such as passwords, during query visualizations, especially in scenarios like screen sharing.

## How
- Added a check for the `__sensitive` flag in fields within the query visualization logic.
- Replaced sensitive field values with a placeholder (`'********'`) when flagged as secret.

## Screenshots
image

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/externalsync/controllers.js
+++ b/packages/legacy-workbench/src/js/angular/externalsync/controllers.js
@@ -8,6 +8,7 @@ const modules = [
 ];
 
 const logger = LoggerProvider.logger;
+const SENSITIVE_FIELD_MASK = 'MASKED_SENSITIVE_FIELD';
 
 angular
     .module('graphdb.framework.externalsync.controllers', modules)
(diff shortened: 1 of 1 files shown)


packages/legacy-workbench/src/js/angular/externalsync/controllers.js
```diff
@@ -143,6 +143,8 @@ function createConnectorQuery(name, prefix, fields, options, reportError) {
                 fcopy[options[i].__name] = fromArrayMap(fcopy[options[i].__name], $translate);
             } else if (options[i].__type === 'JsonString') {
                 fcopy[options[i].__name] = angular.fromJson(fcopy[options[i].__name]);
+            } else if (options[i].__sensitive === true && fcopy[options[i].__name]) {
+                fcopy[options[i].__name] = '********';
```
plamen-yordanov: wasn't this supposed to be an invalid placeholder `${password}` and not stars? `*******` is a valid password and a user may create a connector with that password

yordanalexandrov: Yes
I though it will set it as is but it converts it correctly.

I will see how to go around it.

yordanalexandrov: added a const `SENSITIVE_FIELD_MASK` and replaced the string in the JSON with it removing the quotes, making it invalid

plamen-yordanov (approved):

PR #3219, opened 2026-09-04
GDB-14908 fix multi-region panel primary tags and scrolling

yordanalexandrov:
## What
Adjust multi-region cluster configuration to correctly handle primary tags and panel scrolling. Sorting of primary tags is done in backend.

## Why
Primary tags were not preserved in the cluster configuration and the multi-region panel scrollbar disappeared, making content inaccessible and loaders unreliable.

## How
- Updated TopologyStatus.fromJSON and the ClusterConfiguration constructor, accessors, and toJSON in packages/legacy-workbench/src/js/angular/models/clustermanagement/cluster.js to pass through and serialize primaryTags.
- Adjusted layout and scrolling styles in packages/legacy-workbench/src/css/cluster-configuration/multi-region.css to restore proper scrollbar behavior.
- Updated the multi-region directive in packages/legacy-workbench/src/js/angular/clustermanagement/directives/cluster-configuration/multi-region.directive.js to populate $scope.primaryTags from the backend response and simplified the loader logic to set the loader flag directly.

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [ ] Tests

### DIFF
--- a/packages/legacy-workbench/src/js/angular/models/clustermanagement/cluster.js
+++ b/packages/legacy-workbench/src/js/angular/models/clustermanagement/cluster.js
@@ -195,11 +195,11 @@ export class Node {
 export class TopologyStatus {
     /**
      * @param {string} state - The current state of the topology.
-     * @param {{string, number}[]} [primaryTags=[]] - A map of primary tags and their associated values (optional).
+     * @param {Object<string, number>} [primaryTags={}] - Map of primary cluster tags to their values.
      * @param {number} [primaryIndex] - The index of the primary cluster.
      * @param {string} [primaryLeader] - The leader's rpc address of the primary cluster.
      */
-    constructor(state, primaryTags = new Map(), primaryIndex, primaryLeader) {
+    constructor(state, primaryTags = {}, primaryIndex, primaryLeader) {
         this._state = state;
         this._primaryTags = primaryTags;
         this._primaryIndex = primaryIndex;
@@ -210,6 +210,7 @@ export class TopologyStatus {
         return this._state;
     }
 
+    /** @return {Object<string, number>} Map of primary cluster tags to their values. */
     get primaryTags() {
         return this._primaryTags;
     }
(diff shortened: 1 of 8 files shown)


packages/legacy-workbench/src/js/angular/models/clustermanagement/cluster.js
```diff
@@ -830,10 +839,11 @@ export class ClusterConfiguration {
             verificationTimeout: this._verificationTimeout,
             transactionLogMaximumSizeGB: this._transactionLogMaximumSizeGB,
             batchUpdateInterval: this._batchUpdateInterval,
-            nodes: this._nodes
+            nodes: this._nodes,
         };
         if (this._secondaryTag !== undefined) json.secondaryTag = this._secondaryTag;
         if (this._primaryNodes !== undefined) json.primaryNodes = this._primaryNodes;
+        if (this._primaryTags !== undefined) json.primaryTags = this._primaryTags;
```
plamen-yordanov: is this check needed? `this._primaryTags` is has a default value of `[]`. Also, could you expand the `if` statements with curly braces?

packages/legacy-workbench/src/js/angular/models/clustermanagement/cluster.js
```diff
@@ -229,8 +229,7 @@ export class TopologyStatus {
      * @return {TopologyStatus} The mapped TopologyStatus instance.
      */
     static fromJSON({state, primaryTags = {}, primaryIndex, primaryLeader}) {
-        const primaryTagsMap = Object.entries(primaryTags);
-        return new TopologyStatus(state, primaryTagsMap, primaryIndex, primaryLeader);
+        return new TopologyStatus(state, primaryTags, primaryIndex, primaryLeader);
```
plamen-yordanov, line 233: `primaryTags` default to `new Map` in the TopologyStatus constructor, but `fromJSON` defaults to a empty object.

yordanalexandrov: removed the default here.

packages/legacy-workbench/src/js/angular/models/clustermanagement/cluster.js
yordanalexandrov: removed the default value and expanded the statements

packages/legacy-workbench/src/js/angular/models/clustermanagement/cluster.js
plamen-yordanov: shouldn't we remove the map instead? We are using `primaryTags[tag]`

yordanalexandrov: well .. JS types

I made it an object
@param {Object<string, number>} [primaryTags={}] - Map of primary cluster tags to their values.

plamen-yordanov (dismissed):

plamen-yordanov (approved):

PR #3220, opened 2026-09-04
GDB-13849 refactor navigate to login functionality

plamen-yordanov:
## What
Refactor navigate to login functionality.

## Why
There are multiple instances of `navigate('login')` across the codebase which lacked uniform handling of query and path params.

## How
- Replaced `navigate('login')` calls with the newly introduced `navigateToLoginPage` method.
- Implemented `navigateToLoginPage` in `routing-utils.ts` to handle query and path params, including encoding the return URL.
- Removed the navigation the authentication.service.ts#logout as it duplicated navigation to login
- Removed duplicate encoding and decoding

## Testing
jest

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests

### DIFF
--- a/packages/root-config/src/bootstrap/security/security-bootstrap.ts
+++ b/packages/root-config/src/bootstrap/security/security-bootstrap.ts
@@ -10,9 +10,8 @@ import {
   SecurityService,
   service,
   WindowService,
-  UrlPathParams,
   SecurityConfig,
-  getLocationPathWithQueryParams,
+  navigateToLoginPage,
 } from '@ontotext/workbench-api';
 import {LoggerProvider} from '../../services/logger-provider';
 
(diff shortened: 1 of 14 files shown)


packages/root-config/src/bootstrap/security/security-bootstrap.ts
boyan-tonchev, line 88: Shouldn't we use `navigateToLoginPage()` here?

plamen-yordanov: we should. Applied

boyan-tonchev (approved):

PR #3226, opened 2026-09-09
GDB-13270 Allow multiple executions of explain_query and explain_results

plamen-yordanov:
## What
Allow multiple executions of explain query and explain results

## Why
The `explain_query` and `explain_results` cases had a guard that returned when the query already carried its marker comment. The second and onward queries weren't sent

## How
Extracted `applyQueryLLMComment` which adds the comment to the query. If the comment is the same only the comment addition is skipped and not the entire query sending.

## Testing
n/a

## Screenshots
cypress

## Checklist
- [x] Branch name
- [x] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/legacy-workbench/package.json
```diff
@@ -76,7 +76,7 @@
         "ng-tags-input": "^3.2.0",
         "oclazyload": "^1.1.0",
         "ontotext-graphql-playground-component": "^0.0.9",
-        "ontotext-yasgui-web-component": "^1.7.2",
+        "ontotext-yasgui-web-component": "^1.7.3",
```
yordanalexandrov, line 79: I have released `1.7.4` from the 1.7 branch. `1.7.4` is from main branch.

plamen-yordanov: did you mean  `1.8.0` is from main?

yordanalexandrov: Yes. I though this is for 3.5

plamen-yordanov: Released and replaced with 1.8.0

yordanalexandrov (approved):

PR #3241, opened 2026-09-18
GDB-14749: Expose isRestricted check in the new workbench.

boyan-tonchev:
## What
Introduces a shared restriction mechanism for the new workbench, used to determine whether the current view should show its restriction banner (missing license, no write access, or unsupported repository type) and to adjust which repositories the user is offered when switching repositories.

## Why
`PageRestrictionsComponent` already had the banner UI and the logic to turn a restricted state into user-facing messages, but nothing actually drove it: its `isRestricted` signal was a dead stub that was never updated, so the banner could never reflect the actual state.

This implements the missing piece in the same way the legacy workbench already does it through its `core-errors` directive.

## How
A route now declares the conditions it cares about in its `WorkbenchRouteData`. `AppComponent` reads that declaration on every navigation and passes it to a new `RestrictionService`, which is the only place that knows how to turn the declaration and the current repository, license, and security state into an actual restricted/not-restricted decision.

The service reacts not only to navigation, but also to repository, license, and security changes, so the decision stays correct even when the view itself does not change.

The resulting decision is stored in a new `RestrictionContextService`, a shared piece of state that both the Angular and, eventually, AngularJS sides can read from, in the same way repository and license state already work.

`PageRestrictionsComponent` now simply listens to that context instead of calculating the state itself and passes the value to the existing `RestrictionResolverService` to render the banner. This means the UI itself does not change; only the source of the restriction state has been implemented.

## Checklist
- [x] Branch name
- [ ] Target branch
- [x] Commit messages
- [x] Squash commits
- [x] MR name
- [x] MR Description
- [x] Tests


packages/api/src/models/restrictions/view-restriction.ts
```diff
@@ -0,0 +1,23 @@
+import {Model} from '../common';
+
+/**
+ * The restriction conditions that can apply to a view.
+ */
+export type ViewRestrictionCondition = 'write' | 'ontop' | 'fedx' | 'license';
```
yordanalexandrov: Give more meaningful names. Maybe make it an enum, so the keys are more descriptive.

packages/api/src/models/restrictions/view-restriction.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Describes the conditions that restrict a given view.
+ */
+export class ViewRestriction extends Model<ViewRestriction> {
+  readonly restrictions: ViewRestrictionCondition[];
+
+  /**
+   * Creates a new ViewRestriction instance.
```
yordanalexandrov: The constructor doesn't really need a comment unless it does something special which it should not do.
But I'm just being pick.

packages/api/src/services/restriction/restriction-context.service.ts
```diff
@@ -0,0 +1,79 @@
...
+  updateIsViewRestricted(isViewRestricted: boolean): void {
+    this.updateContextProperty(this.IS_VIEW_RESTRICTED, isViewRestricted);
+  }
+
+  /**
+   * Retrieves the isViewRestricted value from the current context.
+   *
+   * @returns The current isViewRestricted value, or `false` if not set.
+   */
+  isViewRestrictedSnapshot(): boolean {
```
yordanalexandrov: Remove the `Snapshot` suffix. I don't think we add that in other context services.

packages/api/src/services/restriction/restriction.service.ts
```diff
@@ -0,0 +1,115 @@
...
+import {ViewRestriction, ViewRestrictionCondition} from '../../models/restrictions';
+import {LifecycleHooks} from '../../providers/service/lifecycle-hooks';
+import {service} from '../../providers';
+import {RestrictionContextService} from './restriction-context.service';
+import {RepositoryContextService} from '../domain/repository';
+import {LicenseContextService} from '../domain/license';
+import {AuthorizationService, SecurityContextService} from '../domain/security';
+
+/**
+ * Generic service for calculating view restrictions.
```
yordanalexandrov: This does not look llike a generic service. It looks like a View Restrictions service.

Maybe rename it as `ViewRestrictionService` and remove all `View` from method names

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.html
```diff
@@ -17,7 +17,7 @@
   @if (repositoryList() && !selectedRepository()) {
     <app-repository-picker-list
       class="repository-list"
-      [requireWriteAccess]="isRestricted()"
+      [requireWriteAccess]="isViewRestricted()"
```
yordanalexandrov: Why `isViewRestricted` is passed to `requeireWriteAccess`?

packages/workbench/src/app/app.component.ts
```diff
@@ -99,4 +93,53 @@ export class AppComponent implements OnInit, OnDestroy {
...
+          this.updateGuideServices();
+        }
+      })
+    );
+  }
+
+  /**
+   * Recalculates the view restriction on every completed Angular navigation.
+   */
+  private subscribeToViewRestrictionUpdates(): void {
```
yordanalexandrov: This name is not correct.

packages/api/src/models/restrictions/view-restriction.ts
```diff
@@ -0,0 +1,23 @@
...
+   *
+   * @param data - Partial data to initialize the ViewRestriction object.
+   */
+  constructor(data?: Partial<ViewRestriction>) {
```
yordanalexandrov: Why do se need to pass an object for only one field? use the field as an argument and assign it directly

packages/api/src/services/restriction/restriction.service.ts
```diff
@@ -0,0 +1,115 @@
...
+  private readonly authorizationService = service(AuthorizationService);
+
+  private readonly conditionCheckers: Record<ViewRestrictionCondition, () => boolean> = {
+    license: () => this.isViewLicenseRestricted(),
+    write: () => this.isViewWriteRestricted(),
+    ontop: () => this.isViewOntopRestricted(),
+    fedx: () => this.isViewFedxRestricted(),
+  };
+
+  onCreated(): void {
```
yordanalexandrov: Subscribe to user change.
When I change the user the restrictions are not evaluated.

Try with free access

packages/api/src/services/restriction/restriction-context.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+type RestrictionContextFieldParams = {
+  readonly IS_VIEW_RESTRICTED: boolean;
+  readonly VIEW_RESTRICTION: ViewRestriction;
+};
+
+/**
+ * Service for managing restriction context in the application.
+ */
+export class RestrictionContextService extends ContextService<RestrictionContextFields> implements DeriveContextServiceContract<RestrictionContextFields, RestrictionContextFieldParams>, LifecycleHooks {
```
yordanalexandrov: Not sure why it implemented `LifecycleHooks`

packages/api/src/services/restriction/restriction.service.ts
boyan-tonchev: We decided that the service should be generic, regardless of the fact that it is currently used only for view restrictions. This is why the service is called RestrictionService, while the view-related methods have view in their names.

packages/workbench/src/app/components/page-restrictions/page-restrictions.component.html
boyan-tonchev: Because when the view is restricted, we filter the repositories to which the user has write access. Maybe it would be better to rename the picker list field to `isViewRestricted`. Is this what you had in mind?

packages/api/src/models/restrictions/view-restriction.ts
boyan-tonchev: I made it similar to the other models.

packages/api/src/services/restriction/restriction.service.ts
boyan-tonchev: You are right. I miss this.

